Issue #26, opened 2026-04-10
parse_toml truncates quoted values containing '#'


ylohnitram:
## Problem

The `parse_toml` helper introduced in #9 / PR #25 strips inline comments via a naive `line.split("#", 1)[0]`. This means a `#` character inside a quoted value is incorrectly treated as the start of a comment and the rest of the value is truncated.

Verified against the new parser:

```toml
[gitlab]
token = "glpat-abc#def"
```

```
$ parse_toml token
glpat-abc
```

Expected: `glpat-abc#def` (the entire quoted value).

## Impact

**LOW today.** None of the current `colony.example.toml` templates contain a `#` inside any quoted value, and GitLab personal access tokens (`glpat-...`) in practice do not contain `#`. URLs could theoretically contain `#` fragments but that would not apply to a GitLab base URL. The project path uses `/`, not `#`.

**Why it still matters.** This is a silent data-corruption bug. A user who ever puts a `#` in any value (custom token format, URL with fragment, project name with `#`, future config keys) will see their value silently truncated with no error, and debugging will be painful because the surrounding code trusts `parse_toml` to return the literal TOML value.

The pre-#9 parser had the same class of bug for a different reason (it stripped all whitespace), so this is not a regression introduced by #9, just a latent issue that was not in scope for #9's fix.

## Scope

Two files today (after #9 merges):

- `dev-apprenticeship/code-review/scripts/start-colony.sh`
- `dev-apprenticeship/triage/scripts/start-colony.sh`

If #13 (section awareness) also adds a section parser that calls `split("#", 1)`, the same fix must apply there.

If the shared-helper follow-up from this issue's sibling (file a separate issue for `parse_toml` deduplication) is taken first, there is only one file to fix.

## Proposed fix

Replace the naive comment stripper with a proper state machine that respects quoted strings. Rough sketch:

```python
def strip_inline_comment(line: str) -> str:
    out = []
    quote = None
    i = 0
    while i < len(line):
        ch = line[i]
        if quote:
            out.append(ch)
            if ch == "\\" and i + 1 < len(line):
                out.append(line[i + 1])
                i += 2
                continue
            if ch == quote:
                quote = None
        else:
            if ch == "#":
                break
            if ch in ('"', "'"):
                quote = ch
            out.append(ch)
        i += 1
    return "".join(out)
```

Then call `line = strip_inline_comment(raw)` instead of `raw.split("#", 1)[0]`.

## Test plan

[...]

Issue #13, opened 2026-04-10
parse_toml ignores TOML sections, head -1 wins silently


ylohnitram:
## Problem

`parse_toml` in `start-colony.sh` uses `head -1` to pick the first matching line:

```bash
parse_toml() {
    grep -E "^[[:space:]]*$1[[:space:]]*=" "$CONFIG" 2>/dev/null \
      | head -1 \
      | sed '...'
}
```

If a config file has the same key name in two sections — for example `name = "foo"` under `[gitlab]` and again under `[llm]` — `parse_toml name` silently returns the first match regardless of which section the caller meant. No error, no warning.

## Impact

Low today because the example config only uses each key once. Becomes a real footgun as soon as the config grows sections that share key names (e.g. `url` under both `[gitlab]` and some future `[registry]`).

The script treats TOML as flat key-value pairs and ignores sections entirely. This is wrong in principle but fixable either by teaching `parse_toml` to respect `[section]` headers or by using a proper TOML parser.

## Scope

- `dev-apprenticeship/triage/scripts/start-colony.sh`
- `dev-apprenticeship/code-review/scripts/start-colony.sh`
- `dev-apprenticeship/planning/scripts/start-colony.sh`

## Related

Reported by QA on #8 as LOW-6. Related to MEDIUM-2 (both are `parse_toml` issues) but the two have independent fixes — MEDIUM-2 is about how values are cleaned, LOW-6 is about how keys are scoped to sections.

ylohnitram:
## Decision

Extend `parse_toml` to accept an optional section name and restrict matching to lines that occur after the matching `[section]` header and before the next `[next_section]` header. Callers in `start-colony.sh` become `parse_toml gitlab url`, `parse_toml llm backend`, etc. Depends on #9 having landed first. This plan assumes #9's python3-based extractor is already in place.

## Scope

Same three files as #9:

- `dev-apprenticeship/planning/scripts/start-colony.sh`
- `dev-apprenticeship/code-review/scripts/start-colony.sh`
- `dev-apprenticeship/triage/scripts/start-colony.sh`

Modify the `parse_toml()` function body and the six call sites in each file (lines ~27-29 and ~46).

## Concrete fix

Replace the python3 heredoc from #9 with a section-aware version:

```bash
parse_toml() {
    # Usage: parse_toml <section> <key>
    # Returns the value of <key> under [<section>], preserving internal whitespace.
[...]

Issue #9, opened 2026-04-10
start-colony.sh parse_toml strips whitespace from values


ylohnitram:
## Problem

`dev-apprenticeship/*/scripts/start-colony.sh` parses TOML config values with a helper that pipes the final match through `tr -d ' '`:

```bash
parse_toml() {
    grep -E "^[[:space:]]*$1[[:space:]]*=" "$CONFIG" 2>/dev/null \
      | head -1 \
      | sed 's/.*= *"\{0,1\}\([^"]*\)"\{0,1\}/\1/' \
      | tr -d ' '
}
```

The trailing `tr -d ' '` strips **all** whitespace from the value, not just leading or trailing. Any TOML value that legitimately contains a space is corrupted.

## Reproduction

```bash
cat > /tmp/test.toml <<TOML
[llm]
command = "claude --model sonnet"
TOML

parse_toml() {
    grep -E "^[[:space:]]*$1[[:space:]]*=" /tmp/test.toml 2>/dev/null \
      | head -1 | sed 's/.*= *"\{0,1\}\([^"]*\)"\{0,1\}/\1/' | tr -d ' '
}
parse_toml command
```

Expected: `claude --model sonnet`
Actual: `claude--modelsonnet`

## Impact

- **Latent today.** The only values currently parsed by the script are `url`, `token`, and `project`, none of which should contain spaces.
- **Active failure mode:** as soon as `llm.command = "claude --model sonnet"` or any other TOML string with spaces is introduced, the daemon will launch with the wrong command and either fail silently or invoke the wrong binary.
- Cross-colony: the same `parse_toml` lives verbatim in `triage`, `code-review`, and `planning` start scripts.

## Scope

- `dev-apprenticeship/triage/scripts/start-colony.sh`
- `dev-apprenticeship/code-review/scripts/start-colony.sh`
- `dev-apprenticeship/planning/scripts/start-colony.sh`

## Related

Reported by QA on #8 as MEDIUM-2. Inherited from the code-review colony pattern.

ylohnitram:
## Decision

Replace the `tr -d ' '` whitespace-stripping sed pipeline in `parse_toml` with a python3-based TOML extractor that preserves internal whitespace in values and only trims surrounding whitespace and quotes. Apply the same fix to all three colonies (triage, code-review, planning).

## Scope

Three files, same bug:
- `dev-apprenticeship/planning/scripts/start-colony.sh` lines 23-25
[...]

PR #28, opened 2026-04-10
refactor: extract parse_toml to shared tools/parse-toml.sh helper

### LINKED ISSUE #27: extract parse_toml to shared helper to prevent drift
## Problem

After #9 / PR #25 lands, the 19-line `parse_toml` python3 helper is duplicated verbatim across two `start-colony.sh` files:

Planning, implementation, and release `start-colony.sh` scripts do not yet parse TOML, but when they grow that need (likely soon, same config format) the duplication will spread to 5 copies.

Two planned follow-up issues will touch the same function body again:

- **#13** adds TOML section awareness (state machine tracking `[section]` headers).
- Separate issue (sibling of this one) fixes the naive `#` comment stripper to respect quoted strings.

Each of those fixes will have to edit every copy, and the first time one copy drifts we have a silent bug that only one colony experiences.

## Impact

**LOW today** because there are only two copies and both are byte-identical. **MEDIUM over time** as the function grows and more colonies need it. Classic drift risk.

## Proposed fix

Extract the helper to a single file sourced by every `start-colony.sh`.

Option A, shared helper at `tools/parse-toml.sh`:

```bash
# tools/parse-toml.sh
# Source this file to get a `parse_toml KEY` function that reads $CONFIG.
parse_toml() {
    python3 - "$CONFIG" "$1" <<'PY'
...body...
PY
}
```

Each `start-colony.sh` then sources it:

```bash
SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
COLONY_DIR="$(dirname "$SCRIPT_DIR")"
REPO_ROOT="$(cd "$COLONY_DIR/../.." && pwd)"
# shellcheck source=../../../tools/parse-toml.sh
. "$REPO_ROOT/tools/parse-toml.sh"
```

Option B, per-federation helper at `dev-apprenticeship/common/parse-toml.sh`. Cleaner if other federations want their own parsers, uglier because it introduces a new `common/` dir.

**Recommendation:** Option A. `tools/` already holds cross-federation shell scripts and `tools/colony-lint.sh` is the natural sibling.

## Sequencing

This issue should land **before** #13 and the `#`-in-value fix, so those two can each edit one file instead of two. If this arrives after #13, it just means #13's 2-file diff rebases cleanly onto the extraction.

Not a hard blocker for either.

## Scope

- Create `tools/parse-toml.sh` with the extracted function and a shellcheck-friendly header.
- Modify `dev-apprenticeship/code-review/scripts/start-colony.sh` to source it.
- Modify `dev-apprenticeship/triage/scripts/start-colony.sh` to source it.

## Test plan

1. `./tools/colony-lint.sh` must pass including shellcheck. The `source` line must use a shellcheck directive so the linter can resolve the target.
2. Regression reproduction from #9: the same fixture TOML must yield the same values for `backend`, `command`, `url`, `token`, `project`.
3. Confirm both start-colony.sh scripts still fail with the existing "Config not found" error for missing configs (guard regression check).
4. `bash -n` and shellcheck on all touched files.

## Related

Reported by `agentis-qa-agent` on PR #25 as LOW finding #3. Separate issue per the granularity rule.

ylohnitram:
## Summary

The python3 `parse_toml` body from #9 was duplicated verbatim across `code-review` and `triage` `start-colony.sh`. Two planned follow-ups (#13 section awareness, #26 inline comment fix) will need to touch the same function again, and the first drift between the copies is a silent bug per colony.

Extracts the helper into `tools/parse-toml.sh` and sources it from both `start-colony.sh` scripts. The helper expects `$CONFIG` to be set before `parse_toml KEY` is called, same calling convention as before.

No behavior change. Net -10 lines (+38 / -48).

Closes #27

## Scope

- New file: `tools/parse-toml.sh` (30 lines, non-executable, shellcheck-friendly header)
- `dev-apprenticeship/code-review/scripts/start-colony.sh`: replace the 20-line function with a 4-line source block
- `dev-apprenticeship/triage/scripts/start-colony.sh`: same

## Test plan

- [x] `./tools/colony-lint.sh` passes (25 passed, 0 failed, 5 skipped)
- [x] `bash -n` passes on all three modified/new files
- [x] Round-trip regression against the #9 fixture TOML via the actual shared helper:
  - `url` -> `https://gitlab.example.com`
  - `token` -> `glpat-xyz`
  - `project` -> `my org/my project` (whitespace preserved)
  - `backend` -> `cli`
  - `command` -> `claude --model sonnet` (whitespace preserved)
  - `nosuchkey` -> empty string
- [x] Config-not-found guard still fires: `start-colony.sh /nonexistent/path` exits 1 with the existing error message

### DIFF
--- a/dev-apprenticeship/code-review/scripts/start-colony.sh
+++ b/dev-apprenticeship/code-review/scripts/start-colony.sh
@@ -18,30 +18,10 @@ if [ ! -f "$CONFIG" ]; then
     exit 1
 fi
 
-# Parse GitLab config from TOML and export for gitlab-api.sh
-# Accepts both flush ("key = ...") and indented ("  key = ...") TOML keys.
-# Preserves internal whitespace in values (only trims surrounding whitespace
-# and matching quotes). Stops at the first match and ignores inline comments.
-parse_toml() {
-    python3 - "$CONFIG" "$1" <<'PY'
-import sys
-path, key = sys.argv[1], sys.argv[2]
-with open(path, "r", encoding="utf-8") as f:
-    for raw in f:
-        line = raw.split("#", 1)[0].rstrip("\n")
-        stripped = line.lstrip()
-        if not stripped.startswith(key):
-            continue
-        rest = stripped[len(key):].lstrip()
-        if not rest.startswith("="):
-            continue
-        value = rest[1:].strip()
-        if len(value) >= 2 and value[0] == value[-1] and value[0] in ('"', "'"):
-            value = value[1:-1]
-        print(value)
-        break
-PY
-}
+# Parse GitLab config from TOML via the shared helper.
+REPO_ROOT="$(cd "$SCRIPT_DIR/../../.." && pwd)"
+# shellcheck source=../../../tools/parse-toml.sh
+. "$REPO_ROOT/tools/parse-toml.sh"
 
 GITLAB_URL=$(parse_toml "url")
 GITLAB_TOKEN=$(parse_toml "token")
--- a/dev-apprenticeship/code-review/scripts/start-colony.sh
+++ b/dev-apprenticeship/code-review/scripts/start-colony.sh
@@ -8,7 +8,9 @@
 
 set -e
 
-SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+# Resolve symlinks on $0 itself so the script works when invoked via a symlink.
+SCRIPT_PATH="$(python3 -c 'import os, sys; print(os.path.realpath(sys.argv[1]))' "$0")"
+SCRIPT_DIR="$(dirname "$SCRIPT_PATH")"
 COLONY_DIR="$(dirname "$SCRIPT_DIR")"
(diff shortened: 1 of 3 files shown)


ylohnitram:
## QA re-check after 3726c64

Two blockers from the previous QA pass are fixed inline:

- **CRITICAL** (CI failure, SC1091): added `# shellcheck disable=SC1091` next to each `source` line in both `start-colony.sh` files. CI colony-lint now passes.
- **HIGH** (symlink regression, #29): `SCRIPT_DIR` is now derived from `python3 os.path.realpath("$0")` instead of `cd "$(dirname "$0")" && pwd`. Verified locally by symlinking the triage script into `/tmp`, running it, and confirming it resolves to the real `dev-apprenticeship/triage/config/colony.toml` path. Closes #29.

`tools/parse-toml.sh` itself is unchanged since the prior pass. Redundant shebang (LOW) stays as follow-up #30.

### Status
- CI: colony-lint green (https://github.com/Replikanti/agentis-colonies/actions/runs/24250056758)
- Closes: #27, #29
- Follow-up: #30

PR #31, opened 2026-04-10
fix: make parse_toml section-aware (#13)

### LINKED ISSUE #13: parse_toml ignores TOML sections, head -1 wins silently

ylohnitram:
## Summary

- `tools/parse-toml.sh` now takes a section argument: `parse_toml SECTION KEY`. Matching is restricted to lines under `[SECTION]`, so same-name keys in sibling sections no longer collide silently.
- Both `start-colony.sh` scripts in triage and code-review updated to pass explicit sections: `parse_toml gitlab url`, `parse_toml llm backend`, etc.

Before the fix, a config with `url` under both `[gitlab]` and `[llm]` would always return the `[gitlab]` value regardless of which one the caller wanted, with no error or warning.

## Test plan

- [x] Targeted reproduction: fixture with colliding `url` under `[gitlab]` and `[llm]` now returns the correct value for each section. Before the fix both returned `https://gitlab.example.com`.
- [x] Real example config: `parse_toml gitlab url`, `parse_toml gitlab token`, `parse_toml gitlab project`, `parse_toml llm backend` all extract correct values from `triage/config/colony.example.toml`.
- [x] Array-of-tables regression: `parse_toml colony name` correctly returns `"triage"` and does not bleed into `[[agents]]` entries.
- [x] Quoted value containing `=`: round-trips untouched.
- [x] Missing key and wrong section both return empty string.
- [x] `bash -n` clean on both modified scripts.
- [x] `./tools/colony-lint.sh` passes (25 passed, 0 failed).

Closes #13

### DIFF
--- a/dev-apprenticeship/code-review/scripts/start-colony.sh
+++ b/dev-apprenticeship/code-review/scripts/start-colony.sh
@@ -26,9 +26,9 @@ REPO_ROOT="$(cd "$SCRIPT_DIR/../../.." && pwd)"
 # shellcheck disable=SC1091  # colony-lint runs shellcheck without -x
 . "$REPO_ROOT/tools/parse-toml.sh"
 
-GITLAB_URL=$(parse_toml "url")
-GITLAB_TOKEN=$(parse_toml "token")
-GITLAB_PROJECT_RAW=$(parse_toml "project")
+GITLAB_URL=$(parse_toml gitlab url)
+GITLAB_TOKEN=$(parse_toml gitlab token)
+GITLAB_PROJECT_RAW=$(parse_toml gitlab project)
 
 if [ -z "$GITLAB_URL" ] || [ -z "$GITLAB_TOKEN" ] || [ -z "$GITLAB_PROJECT_RAW" ]; then
     echo "Error: GitLab config incomplete in $CONFIG"
@@ -45,7 +45,7 @@ export GITLAB_PROJECT
 export COLONY_DIR
 
 # Parse LLM backend
-LLM_BACKEND=$(parse_toml "backend")
+LLM_BACKEND=$(parse_toml llm backend)
 
 AGENTS=(
     style_reviewer
--- a/dev-apprenticeship/triage/scripts/start-colony.sh
+++ b/dev-apprenticeship/triage/scripts/start-colony.sh
@@ -26,9 +26,9 @@ REPO_ROOT="$(cd "$SCRIPT_DIR/../../.." && pwd)"
 # shellcheck disable=SC1091  # colony-lint runs shellcheck without -x
 . "$REPO_ROOT/tools/parse-toml.sh"
 
-GITLAB_URL=$(parse_toml "url")
-GITLAB_TOKEN=$(parse_toml "token")
-GITLAB_PROJECT_RAW=$(parse_toml "project")
+GITLAB_URL=$(parse_toml gitlab url)
+GITLAB_TOKEN=$(parse_toml gitlab token)
+GITLAB_PROJECT_RAW=$(parse_toml gitlab project)
 
 if [ -z "$GITLAB_URL" ] || [ -z "$GITLAB_TOKEN" ] || [ -z "$GITLAB_PROJECT_RAW" ]; then
     echo "Error: GitLab config incomplete in $CONFIG"
@@ -45,7 +45,7 @@ export GITLAB_PROJECT
 export COLONY_DIR
 
 # Parse LLM backend
-LLM_BACKEND=$(parse_toml "backend")
+LLM_BACKEND=$(parse_toml llm backend)
 
 AGENTS=(
     issue_creator
--- a/tools/parse-toml.sh
+++ b/tools/parse-toml.sh
@@ -1,21 +1,30 @@
 #!/bin/bash
 # tools/parse-toml.sh: shared TOML key lookup for colony start scripts.
 #
-# Usage: source this file, then call `parse_toml KEY`. Expects $CONFIG to
-# point at the TOML file to read. Returns the first matching value on
-# stdout, preserving internal whitespace and stripping matching quotes.
+# Usage: source this file, then call `parse_toml SECTION KEY`. Expects
+# $CONFIG to point at the TOML file to read. Returns the first matching
+# value under `[SECTION]` on stdout, preserving internal whitespace and
+# stripping matching quotes. Keys outside the requested section are
+# ignored so same-name keys in sibling sections do not collide.
 #
 # Not executable: source only.
 # shellcheck shell=bash
 
 parse_toml() {
-    python3 - "$CONFIG" "$1" <<'PY'
+    python3 - "$CONFIG" "$1" "$2" <<'PY'
 import sys
-path, key = sys.argv[1], sys.argv[2]
+path, section, key = sys.argv[1], sys.argv[2], sys.argv[3]
+target_header = "[" + section + "]"
+in_section = False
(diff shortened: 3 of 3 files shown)


ylohnitram:
## QA Report — PR #31 (fix: make parse_toml section-aware)

### Summary
- **Primary fix verified**: the sibling-key collision from #13 is resolved. A fixture with `url` under both `[gitlab]` and `[llm]` now returns distinct values for each, and the real `colony.example.toml` parses all four values (`gitlab.url`, `gitlab.token`, `gitlab.project`, `llm.backend`) correctly.
- **Colony-lint CI**: green (22 passed / 0 failed / 1 skipped on the PR's CI run, all four `*: shellcheck OK` lines present, `tools: shellcheck OK` present).
- **Local checks**: `bash -n` clean on all three modified files, `./tools/colony-lint.sh` locally reports 25 passed / 0 failed / 5 skipped (shellcheck skipped locally, CI covers it).
- **Scope of call-site update**: complete. `grep parse_toml` across the repo shows only the two in-tree callers (triage and code-review `start-colony.sh`), both migrated. The planning/implementation/release colonies do not source `parse-toml.sh` at all.
- **Symlink smoke test**: still works. Running `start-colony.sh` via a symlink from `/tmp` correctly resolves `$SCRIPT_DIR`, sources `parse-toml.sh`, and spawns the daemons (PR #28 fix not regressed).
- **Header comment**: already updated to `parse_toml SECTION KEY` on line 4.
- **No leftover TODO shim** around `parse_toml` anywhere.

**Verdict: LGTM, approving.** The primary bug is fixed and no blockers found. Two LOW follow-ups filed (see below).

### Findings

| # | Severity | Category | Finding | Repro | Filed |
|---|----------|----------|---------|-------|-------|
| 1 | LOW | parser | Section header with whitespace inside brackets (`[ gitlab ]`) is not matched, though `tomllib` accepts it | See detail 1 | [x] #32 |
| 2 | LOW | ux    | Calling `parse_toml` with only one argument silently returns empty instead of erroring | See detail 2 | [x] #33 |
| 3 | INFO | parser | Pre-existing: `#` inside a quoted value still truncates (not introduced by this PR) | See detail 3 | [x] tracked by #26 |
| 4 | INFO | docs  | Helper header comment already reflects the new signature | N/A | [x] no action |

### Severity guide
- **CRITICAL/HIGH**: none
- **MEDIUM**: none
- **LOW**: minor UX / spec gaps that do not block merging
- **INFO**: observations, pre-existing behavior, things I checked and they are fine

### Checklist of things checked

- [x] Sibling-key collision (`url` in both `[gitlab]` and `[llm]`) — fixed
- [x] Nested table `[llm.openai]` — `parse_toml llm url` does not bleed into `llm.openai`
- [x] Nested table `[llm.openai]` — `parse_toml llm.openai url` returns the nested value
- [x] Array-of-tables `[[agents]]` — does not mis-match `parse_toml agents name`, does not bleed out of the preceding `[gitlab]`, does not bleed into the following `[llm]`
- [x] Missing section returns empty
- [x] Missing key in present section returns empty
- [x] Trailing inline comment on section header (`[gitlab] # comment`) — works
- [x] Prefix-collision keys (`urly` vs `url` in same section) — `=` guard protects it
- [x] Quoted values round-trip untouched (single and double)
- [x] Real example `colony.example.toml` parses for all four fields
- [x] Symlink smoke test (PR #28 fix not regressed)
- [x] `bash -n` clean on all three modified files
- [x] `./tools/colony-lint.sh` passes locally and in CI
- [x] `shellcheck` passes in CI (CI log confirms, shellcheck 0.9.0)
- [x] SC1091 disable directives retained in both `start-colony.sh` files
- [x] No other call sites across the repo (`grep parse_toml`)
- [x] Planning / implementation / release colonies do not source `parse-toml.sh`, so they are unaffected
- [x] Helper header comment documents the new `parse_toml SECTION KEY` signature
- [x] No leftover TODO shim around `parse_toml`
- [x] Section header with whitespace inside brackets (`[ gitlab ]`) — gap documented, filed #32
- [x] Silent empty return on 1-arg call — gap documented, filed #33

### Details

#### Finding 1: section header with whitespace inside brackets is not matched

**Filed as follow-up: #32**

**Reproduction:**
```toml
[ gitlab2 ]
url = "https://spaced.example.com"
```

```bash
export CONFIG=fixture.toml
. tools/parse-toml.sh
parse_toml gitlab2 url
# expected: https://spaced.example.com
# actual:   (empty)
```

The parser uses `stripped == target_header` for exact string comparison, so `[ gitlab2 ]` does not match `[gitlab2]`. The TOML spec allows optional whitespace around bare keys in headers, and Python's `tomllib` parses `[ gitlab2 ]` as section `gitlab2` without issue. None of the bundled example configs use this formatting, so the in-tree flow is unaffected. A user hand-editing their `colony.toml` with the TOML-idiomatic spacing would hit it. Not a regression (the helper never handled this), not a blocker for PR #31.

#### Finding 2: `parse_toml` silently returns empty when called with only one argument

**Filed as follow-up: #33**

**Reproduction:**
```bash
export CONFIG=triage/config/colony.example.toml
. tools/parse-toml.sh
parse_toml gitlab    # one arg, missing KEY
echo "exit=$?"       # exit=0, empty stdout
```

With the new 2-arg signature, bash passes `"$2"` as empty string to Python. `key = ""`, `startswith("")` matches everything, but `rest.startswith("=")` then rejects all lines, so the function prints nothing and exits 0. A stray copy-paste of the old `parse_toml url` signature would surface as an empty variable rather than a loud shell error. Strict `set -u` scripts are already protected (bash aborts with `$2: unbound variable`), but the two current callers use `set -e` without `set -u`. Not a regression (the old 1-arg signature is gone) but worth hardening the helper with a simple arg-count check.

#### Finding 3: INFO — `#` inside quoted value still truncates (pre-existing)

**Tracked in: #26**

The parser does `raw.split("#", 1)[0]` before any quote-awareness, so a value like `a = "has # inside"` is truncated to `"has`. This is already tracked in issue #26 and is **not introduced by PR #31** (same behavior in the pre-PR code). Noted here only so reviewers know I checked.

---

### Follow-ups filed

- #32 — parse_toml misses sections with whitespace inside brackets (LOW)
- #33 — parse_toml silently returns empty when called with missing key argument (LOW)

Both filed per the auto-follow-up rule (no blockers, no permission needed). Implementation plans are posted as comments on each issue.

PR #34, opened 2026-04-10
fix: rename exec shell to exec sh across all agents (#455)

ylohnitram:
## Summary

- Rename `exec shell` to `exec sh` in every agent under `dev-apprenticeship/triage/agents/` and `dev-apprenticeship/code-review/agents/`. 37 call sites across 9 files.
- `sh` is the language name the runtime actually honors; `shell` was a silent no-op because the parser accepted any identifier but the runtime whitelist is `["python", "js", "sh"]`.
- The `gitlab-api.sh` scripts under `scripts/` are bash, not agent source. They are deliberately not touched (per the plan in `Replikanti/agentis-core#455`).

## Why

Every agent wrapped the exec call in `try { exec shell cmd; } catch e { ""; };`. The runtime error `unknown language 'shell'` was swallowed by the catch, the caller's `if len(raw) < 3 return;` short-circuited, and the tick did nothing. This is the bug tracked in `Replikanti/agentis-core#455` (QA report on `Replikanti/agentis-colonies#8`, HIGH-1).

## Relationship to upstream fix

`Replikanti/agentis-core#456` rejects unknown `exec` languages at parse time. The downstream rename in this PR does not depend on that upstream fix landing, because `exec sh` is already valid syntax under the current `agentis v1.1.0`. The two PRs are independent and can merge in either order. However, ordering matters for users who already run deployed colonies:

- Merge this PR first: colonies keep working on `v1.1.0` and gain working agents immediately.
- Merge `agentis-core#456` and tag `v1.1.1`: every `exec shell` call starts failing at parse time. Any deployment that has not yet taken this rename will break on upgrade.

Merging this PR before cutting `v1.1.1` avoids the break window.

## Test plan

- [x] `git grep -n 'exec shell' dev-apprenticeship/*.ag` returns zero hits after the rename.
- [x] `git grep -c 'exec sh' dev-apprenticeship/**/*.ag` shows 37 (matches the pre-rename count of `exec shell`).
- [x] `tools/colony-lint.sh` passes on the rename: 25 passed / 0 failed / 5 skipped. All 9 `.ag` files parse cleanly under `agentis v1.1.0`.
- [x] No accidental touches inside string literals, comments, or bash scripts (grep confirmed each of the 37 occurrences was a live `exec shell <expr>` call, not a doc string).
- [x] `gitlab-api.sh` files under `scripts/` untouched (the plan in `agentis-core#455` explicitly says to leave them alone; they are bash, not agent source).

## Files touched

| File | Lines |
[...]

PR #37, opened 2026-04-11
feat(planning): implement planning colony agents

### LINKED ISSUE #36: Implement planning colony (4 agents)
## Problem

The `dev-apprenticeship/planning` colony is scaffolded (README, config template, empty `start-colony.sh`) but has zero `.ag` agent files. It also has no `scripts/gitlab-api.sh` wrapper.

Per the meta-tracker #5, planning is the third of five colonies. Triage and code-review are already implemented. Planning is the natural next step because it is the simplest of the three remaining: it only reads from GitLab and posts comments (no write to labels, no approvals, no assignments).

### Current state

\`\`\`
$ find dev-apprenticeship/planning -type f
dev-apprenticeship/planning/config/colony.example.toml
dev-apprenticeship/planning/scripts/start-colony.sh   # OLD, pre-hardening
dev-apprenticeship/planning/agents/.gitkeep            # empty
dev-apprenticeship/planning/README.md
\`\`\`

Missing:
- All 4 agent files (\`scope_estimator.ag\`, \`risk_assessor.ag\`, \`task_decomposer.ag\`, \`plan_reviewer.ag\`)
- \`scripts/gitlab-api.sh\` wrapper
- Hardening of \`start-colony.sh\` to match triage/code-review (symlink-safe path, \`parse_toml\` via shared helper, \`GITLAB_*\` env export, \`--enable-exec\` flag, LLM backend selection)

## Out of scope

- Running the colony end-to-end against a real GitLab instance (follow-up after merge)
- Federation bus wiring between planning and triage (future phase)
- Fixing the stale \`exec shell\` comments in other colonies' \`gitlab-api.sh\` (tracked by #35)

## Related

- Meta tracker: #5
- Reference implementations: triage (\`dev-apprenticeship/triage/agents/*.ag\`), code-review (\`dev-apprenticeship/code-review/agents/*.ag\`)

ylohnitram:
## Summary

- Adds the four planning colony agents (scope_estimator, risk_assessor, task_decomposer, plan_reviewer) with learned thresholds for observe/suggest/act
- Adds a lean `scripts/gitlab-api.sh` wrapper (read-only endpoints plus `add-note`) since planning is observe-and-suggest — it never writes labels, approves, or merges
- Hardens `scripts/start-colony.sh` to match the triage template: symlink-safe path resolution, `parse_toml` via shared helper, `GITLAB_*` env export, LLM backend detection, `--enable-exec` flag

## Agent behavior

Each agent runs as its own daemon and follows the same tick shape:

1. **Observe**: pull recent GitLab state (merged MRs, issues, incident patterns), summarize via `prompt() -> Type`, record via `learn(...)` with `"observed"` outcome.
2. **Act**: pick the newest issue labeled `needs-planning` and propose its contribution.
3. **Confidence gate**: read per-agent override from memo; at `>= 0.6` emit the suggestion bus event, otherwise stay in observe mode.
4. **Memoize `last_check`**: use `--since` on the next tick.

### scope_estimator
- Types: `ScopeEstimate { issue_id, phases, points, reasoning, confidence }`, `CalibrationSummary { observed_pairs, ratio_hint, tendencies }`
- Learns from shipped MRs: compares implied scope to actual diff size, remembers over/under ratios
- Emits `planning:scope_estimate`

### risk_assessor
- Types: `RiskAssessment { issue_id, risks, severity, confidence }`, `IncidentPatterns { observed_count, recurring_themes }`
- Learns from recent incident/bug/blocker issues: extracts recurring themes (flaky services, integration gotchas)
- Emits `planning:risks`

### task_decomposer
- Types: `TaskBreakdown { issue_id, subtasks, count, confidence }`, `DecompositionPatterns { epics_seen, avg_subtasks, style }`
- Learns granularity from past epics: average subtask count and decomposition style
- Emits `planning:breakdown` with a 3-8 item numbered markdown list

### plan_reviewer
- Listens on all three peer buses, stashes each partial output in memo under `plan_reviewer:<issue_id>:<slot>` so outputs arriving across multiple ticks can still be combined
- When all three slots are populated for the newest unplanned issue, assembles a single markdown plan with **Scope**, **Risks**, **Subtasks** sections
- Self-reviews against past criteria via `recommend("review_plan", ["completeness"])`
- At effective confidence `>= 0.85`: posts as issue comment via `gitlab-api.sh add-note`
- At `>= 0.6` but `< 0.85`: emits `planning:draft_plan` for human review
- Always calls `learn("review_plan", ...)` on every observation

## gitlab-api.sh endpoints

```
issues --needs-planning [--since ISO8601]
issue <iid>
issue-notes <iid>
add-note <iid> --body <text>
merge-requests [--state merged] [--since ISO8601]
mr <iid>
```

No `label`, `approve`, `assign`, or `merge` endpoints. Planning only reads and comments. The header comment says this explicitly so the next colony implementer does not drift the surface.

## Test plan

- [x] `./tools/colony-lint.sh` passes (structure, config, markdown, .ag syntax for all 4 agents)
- [x] `agentis commit` parses all 4 agent files without errors
- [x] `bash -n` clean on both scripts
- [x] `start-colony.sh` spawns 4 daemons mirroring `AGENTS=(scope_estimator risk_assessor task_decomposer plan_reviewer)`
- [x] `plan_reviewer.ag` uses `listen()` with 500ms timeout on all three peer buses and keys partial outputs by issue_id so multi-tick aggregation works
- [x] `gitlab-api.sh` has no write endpoints beyond `add-note`

Closes #36

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/dev-apprenticeship/planning/agents/plan_reviewer.ag
+++ b/dev-apprenticeship/planning/agents/plan_reviewer.ag
@@ -1,0 +1,175 @@
+// plan_reviewer.ag -- Planning colony: assembles the final plan.
+//
+// Listens for scope estimates, risk assessments, and task breakdowns
+// from the three peer agents. Once all three have reported on the
+// same issue, it assembles a combined plan (markdown), checks it
+// against learned review criteria, and either posts it as a comment
+// (high confidence) or emits it for human review (medium confidence).
+//
+// Partial peer outputs are stashed in memo under
+// plan_reviewer:<issue_id>:<slot> so inputs arriving across multiple
+// ticks can still be combined.
+//
+// Run as daemon: agentis daemon agents/plan_reviewer.ag --colony planning
+// Requires: exec_foreign capability, GITLAB_* env vars
+
+cb 600;
+
+type ScopeEstimate {
+    issue_id: int,
+    phases: int,
+    points: int,
+    reasoning: string,
+    confidence: float
+}
+
+type RiskAssessment {
+    issue_id: int,
+    risks: string,
+    severity: string,
+    confidence: float
+}
+
+type TaskBreakdown {
+    issue_id: int,
+    subtasks: string,
+    count: int,
+    confidence: float
+}
+
+type DraftPlan {
+    issue_id: int,
+    plan: string,
+    review_score: float,
+    confidence: float
+}
+
+fn get_confidence() -> float {
+    let raw = recall_latest("plan_reviewer:confidence");
+    if len(raw) > 0 {
+        return prompt("Return this number as a decimal. Nothing else.", raw) -> float;
+    };
+    return 0.0;
+}
+
+fn stash(slot: string, issue_id: int, body: string) -> void {
+    if len(body) > 2 {
+        memo_write("plan_reviewer:" + to_string(issue_id) + ":" + slot, body);
+    };
+}
+
+fn tick(reason: string) -> void {
+    // 1. Collect peer outputs emitted this tick (partial over time).
+    let scope = listen("planning:scope_estimate", 500);
+    let risks = listen("planning:risks", 500);
+    let breakdown = listen("planning:breakdown", 500);
+
+    let scope_str = to_string(scope);
+    let risks_str = to_string(risks);
+    let breakdown_str = to_string(breakdown);
+
+    // 2. Stash each partial under the issue id it belongs to.
+    // We need the issue id to route peer outputs; ask the LLM once per
+    // non-empty bucket rather than trying to parse the record ourselves.
+    if len(scope_str) > 2 {
+        let scope_iid = prompt("Return just the issue_id integer from this JSON record. Nothing else.", scope_str) -> int;
+        stash("scope", scope_iid, scope_str);
+    };
+    if len(risks_str) > 2 {
+        let risks_iid = prompt("Return just the issue_id integer from this JSON record. Nothing else.", risks_str) -> int;
+        stash("risks", risks_iid, risks_str);
+    };
+    if len(breakdown_str) > 2 {
+        let bd_iid = prompt("Return just the issue_id integer from this JSON record. Nothing else.", breakdown_str) -> int;
+        stash("breakdown", bd_iid, breakdown_str);
+    };
+
+    // 3. Pick the newest unplanned issue and see if we have all three
+    //    slots filled for it. If not, keep waiting.
+    let issues_raw = try {
+        exec sh "./scripts/gitlab-api.sh issues --needs-planning";
+    } catch e {
+        print("[plan_reviewer] GitLab unplanned-issue poll failed:", e);
+        "";
+    };
+
+    if len(issues_raw) < 3 {
+        return;
+    };
+
+    let target_iid = prompt(
+        "Return just the issue_id integer of the newest issue in this JSON array. Nothing else.",
+        issues_raw
+    ) -> int;
+
+    let s = recall_latest("plan_reviewer:" + to_string(target_iid) + ":scope");
+    let r = recall_latest("plan_reviewer:" + to_string(target_iid) + ":risks");
+    let b = recall_latest("plan_reviewer:" + to_string(target_iid) + ":breakdown");
+
+    if len(s) < 2 {
+        print("[plan_reviewer] Waiting on scope for issue", target_iid);
+        return;
+    };
+    if len(r) < 2 {
+        print("[plan_reviewer] Waiting on risks for issue", target_iid);
+        return;
+    };
+    if len(b) < 2 {
+        print("[plan_reviewer] Waiting on breakdown for issue", target_iid);
+        return;
+    };
+
+    // 4. Assemble the combined plan and score it against learned criteria.
+    let rec = recommend("review_plan", ["completeness"]);
+    let context = "Issue id: " + to_string(target_iid) +
+        "\nScope estimate JSON: " + s +
+        "\nRisk assessment JSON: " + r +
+        "\nTask breakdown JSON: " + b +
+        "\nPast review criteria: " + to_string(rec);
+
+    let draft = prompt(
+        "You are the plan reviewer for the planning colony. You have three peer outputs for one issue: scope, risks, and task breakdown. Assemble a single markdown plan suitable for posting as an issue comment. Sections: **Scope**, **Risks**, **Subtasks**. Then self-review against past criteria and return a completeness score. Return JSON: issue_id (int, the target issue), plan (string, the full markdown plan), review_score (float 0-1, your self-review against past criteria), confidence (float 0-1, aggregate confidence combining all three peer confidences with the review score).",
+        context
+    ) -> DraftPlan;
+
+    let override_conf = get_confidence();
+    let effective = if override_conf > 0.0 { override_conf; } else { draft.confidence; };
+
+    print("[plan_reviewer] Issue", draft.issue_id, "-> review_score:", draft.review_score, "effective:", effective);
+
+    if effective >= 0.85 {
+        let post_cmd = "./scripts/gitlab-api.sh add-note " + to_string(draft.issue_id) + " --body " + shell_escape(draft.plan);
+        try { exec sh post_cmd; } catch e {
+            print("[plan_reviewer] Failed to post note:", e);
+        };
+        learn(
+            "review_plan",
+            "issue " + to_string(draft.issue_id),
+            "posted: score=" + to_string(draft.review_score),
+            "acted",
+            "planning"
+        );
+        print("[plan_reviewer] Posted plan to issue", draft.issue_id);
+    } else {
+        if effective >= 0.6 {
+            emit("planning:draft_plan", draft);
+            learn(
+                "review_plan",
+                "issue " + to_string(draft.issue_id),
+                "emitted: score=" + to_string(draft.review_score),
+                "emitted",
+                "planning"
+            );
+            print("[plan_reviewer] Emitted draft plan for issue", draft.issue_id);
+        } else {
+            learn(
+                "review_plan",
+                "issue " + to_string(draft.issue_id),
+                "low confidence: score=" + to_string(draft.review_score),
+                "observed",
+                "planning"
+            );
+            print("[plan_reviewer] Observing (confidence:", effective, ")");
+        };
+    };
+}
--- a/dev-apprenticeship/planning/agents/plan_reviewer.ag
+++ b/dev-apprenticeship/planning/agents/plan_reviewer.ag
@@ -15,28 +15,6 @@
 
 cb 600;
 
-type ScopeEstimate {
-    issue_id: int,
-    phases: int,
-    points: int,
-    reasoning: string,
-    confidence: float
-}
-
-type RiskAssessment {
-    issue_id: int,
-    risks: string,
-    severity: string,
-    confidence: float
-}
-
-type TaskBreakdown {
-    issue_id: int,
-    subtasks: string,
-    count: int,
-    confidence: float
-}
-
 type DraftPlan {
     issue_id: int,
     plan: string,
@@ -50,6 +28,15 @@ fn get_confidence() -> float {
         return prompt("Return this number as a decimal. Nothing else.", raw) -> float;
     };
     return 0.0;
+}
+
+// listen() returns Value::Void on timeout, which to_string() renders as the
+// literal string "void". Treat that (and the empty string) as "no message".
+fn received(s: string) -> bool {
+    if s == "void" {
+        return false;
+    };
+    return len(s) > 2;
 }
 
 fn stash(slot: string, issue_id: int, body: string) -> void {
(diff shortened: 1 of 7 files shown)


ylohnitram:
# Combined QA + Code Review — PR #37 (planning colony)

## Verdict: NEEDS CHANGES

One HIGH bug (markdown flattened on post) and one HIGH runtime waste bug (plan_reviewer burns CB every empty tick) together defeat the stated purpose of the colony. Neither is cosmetic — both produce wrong or wasteful behavior immediately on first run. The rest of the PR (parity with triage template, agent syntax, colony-lint, toml alignment) is clean.

## PR checklist re-verification

- [x] \`./tools/colony-lint.sh\` passes — re-ran, 29 passed / 0 failed / 5 shellcheck skips
- [x] \`agentis commit\` parses all 4 agent files without errors — re-ran each against a temp .agentis/ repo, all committed clean
- [x] \`bash -n\` clean on both scripts — re-ran, both OK
- [x] \`start-colony.sh\` spawns 4 daemons mirroring the AGENTS array — confirmed, AGENTS=(scope_estimator risk_assessor task_decomposer plan_reviewer) matches the 4 .ag files
- [x] \`plan_reviewer.ag\` uses \`listen()\` with 500ms timeout on all three peer buses and keys partial outputs by issue_id — confirmed (but see QA-1 for a bug in how empty listens are handled)
- [x] \`gitlab-api.sh\` has no write endpoints beyond \`add-note\` — confirmed, no label/approve/assign/merge

All checklist items remain accurately checked. The new bugs below were not covered by the existing checklist.

## QA findings

| # | Sev | Area | Finding | Status |
|---|-----|------|---------|--------|
| QA-1 | HIGH | plan_reviewer.ag | Spurious prompt+stash every tick when listen() times out, burning ~45% of the cb budget and corrupting the memo store | - [ ] |
| QA-2 | HIGH | gitlab-api.sh (add-note) | \`tr '\n' ' '\` flattens every newline in the posted body, destroying the markdown formatting of the assembled plan | - [ ] |
| QA-3 | MEDIUM | gitlab-api.sh (add-note) | sed-based JSON escaping does not escape control characters (0x00-0x1F), producing invalid JSON if the LLM plan contains one | - [ ] |
| QA-4 | LOW | colony.example.toml | cb_budget values (800/600/1000/800) do not match the \`cb N;\` declarations in the agent files (500/500/500/600); triage and code-review keep these aligned | - [ ] |
| QA-5 | LOW | plan_reviewer.ag | get_confidence() override semantics differ from the three peers and from code-review/approval_decider.ag — peers fall through to 0.0, plan_reviewer falls through to draft.confidence | - [ ] |
| QA-6 | LOW | README.md (planning) | No mention that the three peer agents default to \`confidence = 0.0\` and therefore emit nothing on a fresh install until the operator manually seeds \`<agent>:confidence\` via memo | - [ ] |

## Code review findings

| # | Sev | Area | Finding | Status |
|---|-----|------|---------|--------|
| CR-1 | LOW | plan_reviewer.ag | Types ScopeEstimate, RiskAssessment, TaskBreakdown are declared but never used (peer outputs arrive as strings over the peer TCP path and are passed verbatim to prompt()) | - [ ] |
| CR-2 | LOW | plan_reviewer.ag | Prompts describe peer outputs as \"JSON record\" but cross-process listen() returns an Agentis Display-formatted string like \`ScopeEstimate { issue_id: 42, ... }\`, not JSON. The LLM usually recovers, but the wording is misleading and brittle | - [ ] |
| CR-3 | NIT | all 4 agents | Comment on line 1 uses the phrase \"exec sh\" correctly, but per-colony convention (triage) still says \"exec shell\" in its headers — minor style inconsistency, not introduced here | - [ ] |
| CR-4 | NIT | add-note JSON escape | Prefer the python3 json.dumps approach already used by code-review/post-note for consistency and correctness | - [ ] |

## Details

### QA-1 — plan_reviewer wastes CB and corrupts memo on empty ticks (HIGH)

\`plan_reviewer.ag\` line 63-85:
\`\`\`
let scope = listen(\"planning:scope_estimate\", 500);
let risks = listen(\"planning:risks\", 500);
let breakdown = listen(\"planning:breakdown\", 500);

let scope_str = to_string(scope);
let risks_str = to_string(risks);
let breakdown_str = to_string(breakdown);

if len(scope_str) > 2 {
    let scope_iid = prompt(\"Return just the issue_id integer from this JSON record. Nothing else.\", scope_str) -> int;
    stash(\"scope\", scope_iid, scope_str);
};
\`\`\`

When \`listen()\` times out (the common case — queue empty), the evaluator returns \`Value::Void\`. \`to_string(Void)\` is hard-coded in \`agentis-core/src/evaluator/types.rs:978\` as the literal string \`\"void\"\`:

\`\`\`
Value::Void => write!(f, \"void\"),
\`\`\`

\`len(\"void\") == 4\`, so the \`> 2\` guard is always true. For each empty channel, plan_reviewer calls \`prompt(..., \"void\") -> int\` (50 CB, and a real LLM call if backend is cli/http), then \`memo_write(\"plan_reviewer:<bogus_iid>:<slot>\", \"void\")\` which leaves junk files like \`.agentis/memo/plan_reviewer:0:scope.jsonl\` in the memo store forever.

Impact, per empty tick:
- 3 spurious prompt calls = 150 CB burned plus 3 real LLM calls
- 3 bogus memo files written under whatever int the LLM hallucinates from \"void\"
- tick budget cb=600; 150 wasted is ~25%, and every tick is empty until somebody manually seeds \`scope_estimator:confidence\` \`risk_assessor:confidence\` \`task_decomposer:confidence\` to >= 0.6 (see QA-6). So on a fresh install, the bug fires every tick forever.

Also: the \`stash()\` helper adds a second guard \`if len(body) > 2\`, which similarly passes for \"void\", so the body really is written to memo.

Fix options:
- Guard explicitly on non-void: \`if scope_str != \"void\" && len(scope_str) > 2 { ... }\`
- Better: bump the length threshold to something meaningful (e.g., \`> 10\`) and combine with a void check
- Best: add a language builtin like \`is_void(x)\` rather than string-matching

### QA-2 — add-note destroys markdown newlines (HIGH)

\`dev-apprenticeship/planning/scripts/gitlab-api.sh:98\`:
\`\`\`
ESCAPED_BODY=\$(printf '%s' \"\$BODY\" | sed 's/\\\\/\\\\\\\\/g; s/\"/\\\\\"/g' | tr '\n' ' ')
gl_post \"\$API/issues/\$ID/notes\" \"{\\\"body\\\":\\\"\$ESCAPED_BODY\\\"}\"
\`\`\`

The \`tr '\n' ' '\` call replaces every newline in the plan with a space before sending. Reproduced locally with a faked curl:

\`\`\`
\$ bash gitlab-api.sh add-note 42 --body \$'## Scope\n\nThis is a multi-line plan.\n\n- bullet 1\n- bullet 2\n\n## Risks\n\n- dependency risk'
BODY: {\"body\":\"## Scope  This is a multi-line plan.  - bullet 1 - bullet 2  ## Risks  - dependency risk\"}
\`\`\`

\`plan_reviewer.ag\` assembles a full markdown plan with \`**Scope**\`, \`**Risks**\`, \`**Subtasks**\` sections and bullet lists. After add-note runs, the GitLab comment is one wall of text, unreadable, with no sections and no bullets. The whole point of plan_reviewer is to post a well-formatted plan — this regression silently defeats that on every autonomous post.

Code-review/post-note got this right — it uses \`python3 -c 'import sys,json; print(json.dumps({\"body\": sys.stdin.read()}))'\` which preserves newlines as \`\\n\` JSON escapes. Planning should do the same.

### QA-3 — add-note JSON escaping is incomplete (MEDIUM)

The sed pipeline only handles \`\\\\\` and \`\"\`. JSON requires escaping every control char 0x00-0x1F (\\b \\f \\n \\r \\t and generic \\uXXXX). Example with a raw 0x01:

\`\`\`
\$ bash gitlab-api.sh add-note 42 --body \$'Tab\there\nand control\\x01char'
BODY: {\"body\":\"Tab\there and controlchar\"}
\`\`\`

The 0x01 byte is passed through literally, producing invalid JSON per RFC 8259. GitLab's JSON parser will reject it. The tab is also passed raw (most parsers accept raw tabs in strings but strictly it should be \\t). Unlikely the LLM will emit a raw 0x01, but if it does the entire comment post will 400 and the caller will swallow the \`try { ... } catch\` without telling anyone. Use python3 json.dumps as in code-review.

### QA-4 — cb_budget in colony.example.toml does not match \`cb N;\` in the .ag files (LOW)

| agent | .ag file | example.toml |
|-------|----------|--------------|
| scope_estimator | cb 500 | cb_budget = 800 |
| risk_assessor | cb 500 | cb_budget = 600 |
| task_decomposer | cb 500 | cb_budget = 1000 |
| plan_reviewer | cb 600 | cb_budget = 800 |

Triage and code-review keep these aligned (e.g. triage/labeler.ag \`cb 400\` vs colony.example.toml \`cb_budget = 400\`). Since start-colony.sh spawns daemons from the .ag file directly and never reads \`[[agents]].cb_budget\` from the toml, the runtime honors the \`cb N;\` declaration. The toml entry is cosmetic. Still a consistency slip — pick one number per agent and use it in both places.

### QA-5 — plan_reviewer confidence semantics diverge from peers (LOW)

Three peer agents:
\`\`\`
fn get_confidence() -> float {
    let raw = recall_latest(\"scope_estimator:confidence\");
    if len(raw) > 0 { return prompt(...) -> float; };
    return 0.0;
}
// later:
if confidence >= 0.6 { emit(...); } else { observe; }
\`\`\`

plan_reviewer:
\`\`\`
let override_conf = get_confidence();
let effective = if override_conf > 0.0 { override_conf; } else { draft.confidence; };
if effective >= 0.85 { act; } else if effective >= 0.6 { emit; } else { observe; }
\`\`\`

So plan_reviewer uses the LLM-returned draft.confidence as the default and memo as an override, while the peers use 0.0 as the default and memo as the only source. Net effect on a fresh install: plan_reviewer is ready to act on day 1 (if it ever gets inputs), but the peers it depends on will never emit anything — so plan_reviewer's more permissive semantics are moot, and the colony does nothing.

This may be intentional (plan_reviewer's draft.confidence is an aggregate of three peer confidences, so it's \"more trustworthy\"), but the inconsistency with every other aggregator in the federation (compare code-review/approval_decider.ag which uses the 0.0 default) is worth a decision and a comment explaining the choice.

### QA-6 — undocumented fresh-install observe-only behavior (LOW)

None of the planning docs mention that on first run all four agents will produce no GitLab output until the operator runs something like \`agentis memo write scope_estimator:confidence 0.7\` (or equivalent) for each agent. The README's \"Autonomy Gradient\" table implies confidence grows automatically from experience, but no code in the planning colony writes to the \`<agent>:confidence\` memo keys — the thresholds are purely manual-override, never learned. Either document the seeding requirement, or have agents actually update their own confidence memo from their \`learn()\` history.

This is a pre-existing convention across the whole federation (triage/labeler.ag, code-review/approval_decider.ag all do the same), so downgrading from MEDIUM to LOW, but the planning README is a fresh place to document it.

### CR-1 — dead type declarations in plan_reviewer.ag (LOW)

Lines 18-38 declare \`ScopeEstimate\`, \`RiskAssessment\`, \`TaskBreakdown\` but plan_reviewer never instantiates any of them — peer outputs arrive from cross-process listen() as \`Value::String\` (the emit-side Display format) and are passed verbatim into \`prompt()\`. Remove the three unused type declarations, or actually \`prompt(scope_str) -> ScopeEstimate\` to get structured access.

### CR-2 — prompts describe cross-process peer output as \"JSON\" (LOW)

Cross-process \`listen()\` retrieves the emit-side \`format!(\"{value}\")\` string. For a Value::Struct, that's \`ScopeEstimate { issue_id: 42, phases: 3, ... }\`, not JSON. Calling it a \"JSON record\" in the prompt works in practice because the LLM is flexible, but it's technically wrong and relies on LLM flexibility. Either use JSON-encoded emit (if agentis supports that) or rephrase the prompt to \"Agentis struct record\".

### CR-3 — \"exec sh\" vs \"exec shell\" comment drift (NIT)

Comment on line 3 of each planning agent says \"Called by .ag agents via exec sh\" (new). Triage agents still say \"exec shell\" in their header comments. Pure doc drift, no behavior impact. Not introduced here — triage was updated in #34 but the headers inside the triage .ag files appear to still say \"exec shell\". Separate cleanup.

### CR-4 — prefer python3 json.dumps in add-note (NIT)

Already detailed in QA-2 and QA-3. Code-review/post-note is the template to copy.

## Pre-existing notes (not blockers for this PR)

- \`start-colony.sh\` passes \`--enable-exec\` and \`--backend <X>\` to \`agentis daemon\`, but \`agentis daemon\` parses neither flag (only \`--deny-exec\` exists). They are silently ignored as noise. Daemon mode grants all capabilities by default via \`evaluator.grant_all()\` (\`agentis-core/src/cli/daemon.rs:418\`), so exec_foreign works without \`--enable-exec\`. LLM backend is actually picked up from the config file (\`llm.backend = cli\`). Pre-existing convention shared by triage and code-review.
- \`dev-apprenticeship/README.md\` table marks \`implementation\` and \`release\` as \"Active\" but both have empty \`agents/\` directories. Not touched by this PR.

## What to fix before merging

1. Fix QA-1 (guard plan_reviewer stash against \`Void\`/\"void\")
2. Fix QA-2 (switch add-note to python3 json.dumps so newlines survive)
3. Fix QA-3 (same fix also covers the control-char case)

Everything else (QA-4/5/6, CR-1..4) can land in a follow-up. QA-1 and QA-2 block merge because they break the stated purpose of the colony on the very first tick.

ylohnitram:
## QA fixes pushed (d9b4974)

### HIGH blockers — fixed
- **QA-1** `plan_reviewer` void-guard: added `received(s)` helper that rejects the literal string `"void"` (the Display form of `Value::Void` from a timed-out `listen()`) and empty strings. Empty ticks no longer burn CB on spurious prompts or corrupt memo with hallucinated issue ids.
- **QA-2** `add-note` newline preservation: replaced the `sed + tr '\n' ' '` escape pipe with `python3 -c 'json.dumps(...)'`, mirroring code-review's `post-note`. Smoke-tested locally with a multi-section markdown body: newlines survive as `\n`, quotes and backslashes are escaped, JSON parses. Markdown formatting in the GitLab comment is now preserved.

### Cleanup — also fixed
- **QA-4** `colony.example.toml` cb_budget aligned with actual `cb N;` in each .ag file (500/500/500/600 instead of the drifted 800/600/1000/800).
- **QA-5** `plan_reviewer` confidence semantics: dropped the `override_conf > 0 ? override_conf : draft.confidence` fallback and matched peers' pattern (memo-only, defaults to 0.0).
- **CR-1** Removed unused `ScopeEstimate` / `RiskAssessment` / `TaskBreakdown` type declarations from plan_reviewer.ag — only `DraftPlan` is actually destructured.
- **CR-2** Fixed misleading "JSON record" wording in peer-id extraction prompts (peer outputs cross the bus as struct Display format, not JSON).
- **QA-3** Auto-resolved by switching to python3 json.dumps (control chars 0x00-0x1F are escaped properly now).

### Deferred (not blockers)
- **QA-6** Confidence-memo seeding convention is a federation-wide gap (all colonies require operators to write `<agent>:confidence` memo keys before anything is emitted). Worth documenting once in `dev-apprenticeship/README.md` rather than in this PR.
- **CR-3 / CR-4** Wording nits.
- Pre-existing `--enable-exec` / `--backend` flag passing convention in every colony's `start-colony.sh` (silently ignored by the daemon, daemon mode grants all caps by default) — not touched because it is cross-colony convention, not new with this PR.

### Re-verification
```
./tools/colony-lint.sh
# Results: 29 passed, 0 failed, 5 skipped (shellcheck not installed)

bash -n dev-apprenticeship/planning/scripts/gitlab-api.sh  # clean
agentis commit  # all 4 agents parse clean
```

PR description checklist remains accurate — none of the original boxes needed flipping.

PR #42, opened 2026-04-11
fix(tools): bundle parse_toml hardening fixes (#26, #30, #32, #33)

### LINKED ISSUE #26: parse_toml truncates quoted values containing '#'

### LINKED ISSUE #30: tools/parse-toml.sh has redundant shebang on a source-only file
## Context

Found during QA of PR #28 (`refactor/parse-toml-shared-helper`).

## Observation

`tools/parse-toml.sh` is chmod 644 (non-executable, source-only) and has this header:

```bash
#!/bin/bash
# tools/parse-toml.sh: shared TOML key lookup for colony start scripts.
#
# Usage: source this file, then call \`parse_toml KEY\`. Expects \$CONFIG to
# point at the TOML file to read. Returns the first matching value on
# stdout, preserving internal whitespace and stripping matching quotes.
#
# Not executable: source only.
# shellcheck shell=bash
```

The `#!/bin/bash` shebang on line 1 is redundant with `# shellcheck shell=bash` on line 9 for a source-only file. ShellCheck's own convention for sourced-only libraries is to drop the shebang and rely on the `# shellcheck shell=bash` directive alone — see <https://www.shellcheck.net/wiki/SC2148>.

The file header comment explicitly states "Not executable: source only." so the shebang is misleading: it implies the file can be executed, which it cannot (and should not).

## Suggested fix

Remove the `#!/bin/bash` line and keep only `# shellcheck shell=bash` as the shell hint.

## Severity

LOW — cosmetic, does not affect behavior. ShellCheck passes without warnings either way.

### LINKED ISSUE #32: parse_toml misses sections with whitespace inside brackets
## Problem

`tools/parse-toml.sh` uses exact string match on section headers (`stripped == target_header`). This means a header like `[ gitlab ]` (with whitespace inside the brackets) is not recognized as section `gitlab`, even though the TOML spec allows optional whitespace around bare keys in headers.

### Reproduction

```toml
[ gitlab ]
url = "https://spaced.example.com"
```

```bash
export CONFIG=fixture.toml
. tools/parse-toml.sh
parse_toml gitlab url
# expected: https://spaced.example.com
# actual:   (empty)
```

The real Python `tomllib` accepts `[ gitlab ]` as section `gitlab` without issue, so users editing their `colony.toml` by hand could get a silent empty value for `GITLAB_URL` and see `Error: GitLab config incomplete` even though the key is present.

### Scope

Discovered during QA of PR #31 (fix for #13). Not a regression — same behavior exists after the section-aware fix because the exact match was kept. All in-tree example configs use `[gitlab]` with no interior whitespace, so none of the bundled colonies hit this today.

### Severity

LOW. Cosmetic formatting of user configs is unlikely, and the failure mode is loud ("GitLab config incomplete") rather than silently wrong. Filed as a follow-up to PR #31 so the section-aware parser stays spec-compliant for hand-edited TOML.

### LINKED ISSUE #33: parse_toml silently returns empty when called with missing key argument
## Problem

After PR #31 made `parse_toml` section-aware, the function now takes two positional arguments: `parse_toml SECTION KEY`. If a caller accidentally passes only one argument (e.g. lingering old-style `parse_toml url` from pre-#31 code), the function does not fail loudly.

### Reproduction

```bash
export CONFIG=triage/config/colony.example.toml
. tools/parse-toml.sh
parse_toml gitlab    # one arg, no KEY
echo "exit=$?"
```

Result: exit code 0, empty stdout. No warning, no indication that the caller used the wrong signature.

Under the covers, the Python heredoc receives `key = ""`, `startswith("")` is always true, and the `rest.startswith("=")` guard then rejects every line, so the function prints nothing and exits cleanly.

Under `set -u`, bash itself aborts with `$2: unbound variable` before Python is invoked, so strict-mode scripts are already protected. The two current in-tree callers (`start-colony.sh` in triage and code-review) use `set -e` but not `set -u`, so a typo would silently produce empty config values and the callers' subsequent "GitLab config incomplete" check would catch it — but the error message would point at the config file, not the script bug.

### Scope

Discovered during QA of PR #31. Not a regression (the old single-arg signature is simply gone), but worth hardening the helper against silent misuse during the transition.

### Severity

LOW. All in-tree callers have been migrated in PR #31 and CI passes. Filed so future copy-paste of the old signature into a new colony fails loudly rather than silently.

ylohnitram:
## Summary

Four small fixes to `tools/parse-toml.sh`, bundled because they all touch the same 40-line helper. Every current caller (`triage`, `code-review`, `planning` `start-colony.sh`) still parses a fresh `colony.toml` correctly against the new helper.

| # | Severity | Fix |
|---|---|---|
| #26 | LOW | `#` inside quoted value was silently truncated. Replace naive `split("#", 1)` with state-machine `strip_inline_comment()` that respects `"..."` and `'...'` |
| #30 | NIT | Drop redundant `#!/bin/bash` shebang — file is chmod 644, source only (SC2148) |
| #32 | LOW | Section matching was exact, so `[ gitlab ]` with interior whitespace was skipped. Parse bracket contents and strip whitespace |
| #33 | LOW | `parse_toml` with missing KEY silently returned empty. Add `[ $# -ne 2 ]` guard that exits 2 with stderr message |

## Tests

Adds `tools/test-parse-toml.sh` with 25 assertions covering the four fixes plus regressions:

- `#26`: hash in double-quoted + single-quoted value, inline comment stripped, inline comment after quoted hash
- `#32`: spaced `[ gitlab ]`, double-spaced `[  llm  ]`, tab-padded section header
- `#33`: zero args, missing KEY, extra args — all exit 2 with stderr
- Regression: unknown section, unknown key, empty quoted value, sibling section same-name key, plain key/value
- Combined multi-section fixture exercising all four fixes at once

Adds a test-runner hook at the end of `tools/colony-lint.sh` that discovers and runs every `tools/test-*.sh`, so future regressions fail the lint automatically. This is the first `tools/test-*.sh` in the repo; the hook is general.

## Test plan

- [x] `tools/test-parse-toml.sh` — 25 pass / 0 fail
- [x] `./tools/colony-lint.sh` — 30 pass / 0 fail / 5 skip (shellcheck skipped locally only)
- [x] CI green (colony-lint including shellcheck on CI runners)
- [x] Manual smoke: sourced the new helper and called `parse_toml gitlab url|token|project` + `parse_toml llm backend` from all three sourcing colonies, all return expected values
- [x] `bash -n` clean on all three changed files

## Out of scope

- Moving to a real TOML parser (`python3 -c 'import tomllib'`). Considered and rejected — introduces a Python 3.11+ requirement for the helper, whereas the four targeted fixes keep the blast radius inside 40 lines.

Closes #26
Closes #30
Closes #32
Closes #33

### DIFF
--- a/tools/colony-lint.sh
+++ b/tools/colony-lint.sh
@@ -262,6 +262,21 @@ if command -v shellcheck &>/dev/null; then
     fi
 fi
 
+# --- Tools unit tests ---
+test_scripts=()
+while IFS= read -r -d '' f; do
+    test_scripts+=("$f")
+done < <(find "$REPO_ROOT/tools" -name "test-*.sh" -print0 2>/dev/null)
+
+for t in "${test_scripts[@]}"; do
+    if bash "$t" &>/dev/null; then
+        pass "tools: $(basename "$t") unit tests"
+    else
+        fail "tools: $(basename "$t") unit tests"
+        bash "$t" 2>&1 | tail -20
+    fi
+done
+
 # --- Summary ---
 echo ""
 echo "Results: $PASS passed, $FAIL failed, $SKIP skipped"
--- a/tools/parse-toml.sh
+++ b/tools/parse-toml.sh
@@ -1,4 +1,3 @@
-#!/bin/bash
 # tools/parse-toml.sh: shared TOML key lookup for colony start scripts.
 #
 # Usage: source this file, then call `parse_toml SECTION KEY`. Expects
@@ -11,17 +10,51 @@
 # shellcheck shell=bash
 
 parse_toml() {
+    if [ "$#" -ne 2 ]; then
+        echo "parse_toml: usage: parse_toml SECTION KEY (got $# args)" >&2
+        return 2
+    fi
     python3 - "$CONFIG" "$1" "$2" <<'PY'
 import sys
 path, section, key = sys.argv[1], sys.argv[2], sys.argv[3]
-target_header = "[" + section + "]"
+
+
+def strip_inline_comment(line):
+    """Strip a trailing `# comment` from a TOML line, respecting quoted
+    strings so a `#` inside `"..."` or `'...'` is preserved."""
+    out = []
+    quote = None
+    i = 0
+    n = len(line)
+    while i < n:
+        ch = line[i]
+        if quote:
+            out.append(ch)
+            if ch == "\\" and i + 1 < n:
+                out.append(line[i + 1])
+                i += 2
+                continue
+            if ch == quote:
+                quote = None
+        else:
+            if ch == "#":
+                break
+            if ch in ('"', "'"):
+                quote = ch
+            out.append(ch)
+        i += 1
+    return "".join(out)
+
+
 in_section = False
 with open(path, "r", encoding="utf-8") as f:
     for raw in f:
-        line = raw.split("#", 1)[0].rstrip("\n")
+        line = strip_inline_comment(raw).rstrip("\n")
         stripped = line.strip()
         if stripped.startswith("[") and stripped.endswith("]"):
-            in_section = (stripped == target_header)
+            # Allow `[ gitlab ]` with interior whitespace per TOML spec.
+            sect_name = stripped[1:-1].strip()
+            in_section = (sect_name == section)
             continue
         if not in_section:
             continue
--- a/tools/test-parse-toml.sh
+++ b/tools/test-parse-toml.sh
@@ -11,6 +11,7 @@ set -eu
 
 SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
 # shellcheck source=parse-toml.sh
+# shellcheck disable=SC1091  # colony-lint runs shellcheck without -x
 . "$SCRIPT_DIR/parse-toml.sh"
 
 PASS=0
@@ -90,6 +91,17 @@ else
     fail "#33: missing KEY" "rc=2 + stderr msg" "rc=$rc out=$out"
 fi
 
+# Zero-arg call must also fail loudly.
+set +e
+out="$(parse_toml 2>&1 >/dev/null)"
+rc=$?
+set -e
+if [ "$rc" -eq 2 ] && [ -n "$out" ]; then
+    pass "#33: zero args exits 2 with stderr message"
+else
+    fail "#33: zero args" "rc=2 + stderr msg" "rc=$rc out=$out"
+fi
+
 # Also verify the too-many-args path.
 set +e
 out="$(parse_toml gitlab url extra 2>&1 >/dev/null)"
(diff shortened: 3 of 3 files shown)


ylohnitram:
## QA + Code Review — CLEAN

### Summary

- **Test harness:** 25/25 pass (`bash tools/test-parse-toml.sh`)
- **Colony lint:** 30 pass / 0 fail / 5 skip
- **CI:** SUCCESS (run 24282802634), mergeable
- **Wrapper compatibility:** all three sourcing colonies (triage, code-review, planning) still resolve `url`/`token`/`project`/`backend` correctly

### Issue-by-issue verification

| Issue | Fix location | Verified by |
|---|---|---|
| #26 (`#` in quoted value) | `strip_inline_comment()` state machine at `tools/parse-toml.sh:22-46` | Tests 2, 3, 8 (double + single quotes); live `glpat-abc#def` fixture |
| #30 (redundant shebang) | No shebang at `tools/parse-toml.sh:1`, only `# shellcheck shell=bash` hint | `head -1 tools/parse-toml.sh` |
| #32 (spaced section header) | `sect_name = stripped[1:-1].strip()` at `tools/parse-toml.sh:56` | Tests 4, 9 (spaces + tab padding) |
| #33 (missing KEY arg) | `[ "$#" -ne 2 ]` guard at `tools/parse-toml.sh:13-16` returning 2 with stderr | Test 5 covers zero, one, and three args |

### Parser robustness — extra edge cases probed live

| Edge case | Result |
|---|---|
| Key prefix collision (`url` vs `urlish`) | Correct both orderings |
| Trailing comment on section header `[gitlab] # comment` | Correct — `strip_inline_comment` runs before header detection |
| No-space comment after quoted value `"plain"#comment` | Correct |
| Sibling section same-name key (`gitlab.url` vs `llm.url`) | Correct — Test 11 |

### Test-runner hook

The loop at `tools/colony-lint.sh:271-278` on a failing `test-*.sh`:
1. Falls through to `fail`, incrementing `FAIL`
2. Re-runs the test with `tail -20` so failure output is visible in CI logs
3. Script exits non-zero at `tools/colony-lint.sh:283` via `[ "$FAIL" -eq 0 ]`

Loud-failure behavior confirmed.

### Findings

**None.** Nothing to block merge, nothing worth a follow-up issue.

### Recommendation

**CLEAN — ready to merge.**

Closes #26, #30, #32, #33.

PR #43, opened 2026-04-11
fix(triage,code-review): URL-encode gitlab-api query parameters

ylohnitram:
## Summary

- Apply the `gl_get_q` + `curl -G --data-urlencode` pattern from PR #40 (planning) to triage and code-review wrappers
- Fix the same query-string injection bug in three places: triage `issues`, code-review `merge-requests`, code-review `issues`
- Bonus: triage `update-issue --assignee` user lookup had the same bug — usernames with `+`, `&`, or spaces silently fragmented the query

## What changed

| File | Endpoints rewritten |
|---|---|
| `dev-apprenticeship/triage/scripts/gitlab-api.sh` | `issues`, `update-issue --assignee` lookup |
| `dev-apprenticeship/code-review/scripts/gitlab-api.sh` | `merge-requests`, `issues` |

Both files now expose `gl_get_q <url> [--data-urlencode k=v ...]` alongside the existing `gl_get`. Path-only GETs (numeric IID lookups, hardcoded `per_page=100`, etc.) keep using `gl_get` — there is no injection vector when no user input flows into the query.

## What did not change

- `mr`, `mr-changes`, `mr-notes`, `mr-approvals`, `members` in code-review — all hardcoded query strings or numeric IID paths
- `issue`, `issue-events`, `issue-notes`, `members`, `labels` in triage — same reason
- JSON body encoding in `create-issue` / `update-issue` — separate bug class, would be a follow-up if it matters

## Test plan

- [x] Mock-curl smoke test on every rewritten endpoint, verifying the `--data-urlencode` pairs flow through with tricky values like `2024-01-01T00:00:00+02:00`, `alice+bob & co`
- [x] Verified the assignee lookup runs `gl_get_q` first, then the PUT, both with correct args
- [x] `./tools/colony-lint.sh` — 30 pass / 0 fail / 5 skip (same as main)
- [x] `bash -n` clean on both files

Closes #41

Issue #14, opened 2026-04-10
planning gitlab-api.sh does not URL-encode query parameters


ylohnitram:
## Problem

`dev-apprenticeship/planning/scripts/gitlab-api.sh` builds GitLab API query strings without URL-encoding user-supplied values. The `--since` and `--labels` flags interpolate raw strings straight into the query:

```bash
# conceptual
URL="$GITLAB_URL/api/v4/projects/$GITLAB_PROJECT/issues?updated_after=$SINCE&labels=$LABELS"
```

GitLab's API tolerates ISO8601 timestamps and simple label strings, so the current usage works by accident. Anything containing `&`, `#`, `?`, spaces, or non-ASCII characters will break the query.

## Impact

- Today: fine. Agents only pass ISO8601 timestamps and comma-separated ASCII labels.
- Tomorrow: as soon as an agent passes a label with a space (`"in progress"`) or a human types a non-trivial value, the query silently fragments.

## Scope

`dev-apprenticeship/planning/scripts/gitlab-api.sh`

Code-review colony may have similar scripts with the same pattern. Check and fix in one pass, or file separate issues per colony if the fix is different.

## Related

Reported by QA on #8 as LOW-7.

ylohnitram:
## Decision

Introduce a small `urlenc` helper that pipes a value through `python3 -c 'import sys,urllib.parse; print(urllib.parse.quote(sys.argv[1], safe=""))'` and wrap `$SINCE` and `$LABELS` in it when building the `issues` request URL. Apply the same fix to the `--since` parameter in `code-review/scripts/gitlab-api.sh`, where the same bug exists.

## Scope

Primary file (issue body):
- `dev-apprenticeship/planning/scripts/gitlab-api.sh` — the `issues)` case block. Search for `URL="$API/issues` and the two `updated_after=$SINCE` / `labels=$LABELS` interpolations (currently lines ~77-83).

Cross-colony hit (verified by grep):
- `dev-apprenticeship/code-review/scripts/gitlab-api.sh` — two `issues-like` blocks at lines ~52-66 and ~115-130. Both interpolate `$SINCE` raw into `updated_after=`. The `--since` path is vulnerable to the same `+`, space, and non-ASCII breakage.
- `dev-apprenticeship/triage/scripts/gitlab-api.sh` — `issues)` case at lines ~58-74. Same `$SINCE` interpolation bug. Fix it too.

[...]

PR #48, opened 2026-04-11
fix(triage): escape JSON bodies in create-issue/update-issue via python3 json.dumps

### LINKED ISSUE #44: triage/gitlab-api.sh: JSON body construction in create-issue/update-issue breaks on quotes and backslashes
## Problem

Same bug class as #14/#41 (URL encoding) but on the JSON body axis instead of the query string axis.

In `dev-apprenticeship/triage/scripts/gitlab-api.sh`, the `create-issue` and `update-issue` endpoints build their JSON bodies with raw string concatenation:

```bash
# create-issue (around line 110)
BODY="{\"title\":\"$TITLE\""
if [ -n "$DESC" ]; then
    BODY="$BODY,\"description\":\"$DESC\""
fi
...

# update-issue (around line 140)
BODY="{"
SEP=""
if [ -n "$ADD_LABELS" ]; then
    BODY="$BODY${SEP}\"add_labels\":\"$ADD_LABELS\""
    ...
```

Any value containing a `"`, `\`, newline, or control character produces malformed JSON. Titles like `Fix "unterminated string" error` are routine LLM output from `issue_creator.ag`.

## Severity

MEDIUM. Currently latent because the agent layer (`.ag` exec sh) breaks on the same value first when it tries to pass `draft.title` through shell word-splitting. Will become HIGH the moment anyone fixes the shell escaping in the `.ag` callers (which would then let broken JSON reach the server).

## Fix

Use the same `python3 -c 'json.dumps(...)'` pattern already proven in `code-review/scripts/gitlab-api.sh` (`post-note`, around line 120) and `planning/scripts/gitlab-api.sh` (`add-note`). Build a dict of only the fields that are non-empty, then dump it.

Suggested shape for `create-issue`:

```bash
JSON_BODY=$(python3 - <<PY
import os, json
body = {"title": os.environ["TITLE"]}
if os.environ.get("DESC"):
    body["description"] = os.environ["DESC"]
if os.environ.get("LABELS"):
    body["labels"] = os.environ["LABELS"]
if os.environ.get("PRIORITY"):
    body["priority"] = os.environ["PRIORITY"]
print(json.dumps(body))
PY
)
gl_post "$API/issues" "$JSON_BODY"
```

Same treatment for `update-issue` (add_labels, remove_labels, priority, assignee_ids). Export the vars before the heredoc so python can read them without going through shell quoting.

## Affected files

- `dev-apprenticeship/triage/scripts/gitlab-api.sh` — `create-issue` and `update-issue`

## Not in scope

- `code-review/scripts/gitlab-api.sh` — `post-note` already uses `python3 json.dumps`
- `planning/scripts/gitlab-api.sh` — `add-note` already uses `python3 json.dumps`

## Related

- #14, #41 (query string URL encoding — same bug class, different axis)
- Agent-side shell escaping for `.ag` exec sh calls — separate upstream fix, would promote this to HIGH

ylohnitram:
## Summary

`dev-apprenticeship/triage/scripts/gitlab-api.sh` built `create-issue` and `update-issue` JSON bodies by concatenating strings, so any title or description containing a `"`, `\`, newline, or control character produced malformed JSON. LLM output from `issue_creator.ag` routinely contains those characters, so titles like `Fix "unterminated string" error` were dropped on the floor.

Same bug class as #14/#41 (URL encoding) but on the JSON body axis.

Replaces the raw concat with the `python3 json.dumps` pattern already proven in `code-review/post-note` and `planning/add-note`. Values pass through env vars to keep argv clean, the heredoc is single-quoted so bash does nothing to the payload. Only non-empty fields are included. `update-issue` keeps the existing user-ID lookup flow (PR #43 URL encoding, unchanged) and wraps the final body including `assignee_ids` in json.dumps.

Closes #44

## Test plan

- [x] `tools/colony-lint.sh` -> 30 passed, 0 failed
- [x] Mock-curl smoke test covering 5 cases (simple, quotes+newlines+backslash, labels+priority, assignee, title-only) — all 5 emitted bodies parse as valid JSON via `python3 json.loads`
- [x] Test case 2 specifically reproduces the issue-report example `Fix "unterminated string" error` with embedded newlines and backslash
- [x] `assignee_ids` payload contains the int user ID, not a quoted string
- [x] Regression: `labels`, `members`, `issues`, `issue-events`, `issue-notes` endpoints untouched
- [ ] QA report posted as comment

## Sample smoke test output

```
VALID: {"title": "Simple title", "description": "A body", "labels": "bug,urgent"}
VALID: {"title": "Fix \"unterminated string\" error", "description": "line 1\nline 2\nwith \"quoted\" bits and \\backslash"}
VALID: {"add_labels": "needs-triage", "remove_labels": "stale", "priority": "high"}
VALID: {"add_labels": "bug", "assignee_ids": [42]}
VALID: {"title": "Minimal"}
ALL 5 JSON BODIES VALID
```

### DIFF
--- a/dev-apprenticeship/triage/scripts/gitlab-api.sh
+++ b/dev-apprenticeship/triage/scripts/gitlab-api.sh
@@ -124,19 +124,23 @@ case "$CMD" in
             echo '{"error": "--title is required"}' >&2
             exit 1
         fi
-        # Build JSON body
-        BODY="{\"title\":\"$TITLE\""
-        if [ -n "$DESC" ]; then
-            BODY="$BODY,\"description\":\"$DESC\""
-        fi
-        if [ -n "$LABELS" ]; then
-            BODY="$BODY,\"labels\":\"$LABELS\""
-        fi
-        if [ -n "$PRIORITY" ]; then
-            BODY="$BODY,\"priority\":\"$PRIORITY\""
-        fi
-        BODY="$BODY}"
-        gl_post "$API/issues" "$BODY"
+        # Build JSON body via python3 json.dumps so titles and descriptions
+        # containing quotes, backslashes, newlines, or control chars are
+        # escaped correctly. Values are passed via env vars to keep argv
+        # clean and avoid re-quoting hell.
+        JSON_BODY=$(TITLE="$TITLE" DESC="$DESC" LABELS="$LABELS" PRIORITY="$PRIORITY" python3 - <<'PY'
+import os, json
+body = {"title": os.environ["TITLE"]}
+if os.environ.get("DESC"):
+    body["description"] = os.environ["DESC"]
+if os.environ.get("LABELS"):
+    body["labels"] = os.environ["LABELS"]
+if os.environ.get("PRIORITY"):
+    body["priority"] = os.environ["PRIORITY"]
+print(json.dumps(body))
+PY
+)
+        gl_post "$API/issues" "$JSON_BODY"
         ;;
 
     update-issue)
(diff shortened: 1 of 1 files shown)


ylohnitram:
## Combined QA + Review Report

### Summary

| Check | Result |
|---|---|
| PR scope matches claim | PASS |
| colony-lint | PASS (30 passed, 0 failed, 5 skipped) |
| shellcheck (stdin via docker) | PASS (no findings on triage/code-review/planning scripts) |
| create-issue smoke tests | PASS (8 cases incl. quotes, backslash, newline, ctrl char, unicode, shell metachars, labels with quotes, missing title error path) |
| update-issue smoke tests | PASS (all 3 user-requested edge cases + priority-only + regression unknown flag) |
| assignee_ids type | PASS (int, not string) |
| Regression on untouched endpoints | PASS (issues, issue, issue-events, issue-notes, members, labels all behave unchanged) |
| No more JSON concat bugs in triage/code-review/planning | PASS |
| set -e propagates python failures | PASS (verified) |
| CI | PASS (colony-lint green on the branch) |

All three edge cases the author requested I cover:

```
=== update-issue only --assignee ===        -> {"assignee_ids": [42]}    VALID
=== update-issue --assignee empty lookup === -> {}                        VALID
=== update-issue no args ===                 -> {}                        VALID
```

### Findings

|---|---|---|---|---|
| 1 | INFO | scope | `Closes #44` is accurate as scoped: #44 explicitly limited itself to the script JSON-concat layer and noted the `.ag` shell-concat layer was a separate latent bug. PR #48 closes #44 correctly. End-to-end round-trip of `Fix "unterminated string" error` still does not work because the `.ag` caller layer shell-splits the title first. Filed as follow-up #49. | - [x] |
| 2 | LOW | ux | `update-issue <id>` with no modifying flags produces `{}` which becomes a no-op PUT with ambiguous response. Not a current-caller bug. Filed as follow-up #50. | - [x] |
| 3 | INFO | consistency | Env-var pattern here vs stdin pattern in `post-note`/`add-note` is a justified divergence: the precedents carry a single `body` field (stdin is elegant), the new code carries up to 5 fields (env vars are the natural multi-field extension). Both are equivalently safe. No change requested. | - [x] |
| 4 | INFO | defensive | `int(os.environ["ASSIGNEE_ID"])` could theoretically raise `ValueError` if `$USER_ID` were non-numeric, but the upstream `grep -o '"id":[0-9]*' \| cut -d: -f2` guarantees digits-only or empty, and the `if os.environ.get("ASSIGNEE_ID"):` guard short-circuits empty. Verified no code path can reach `int()` with bad input. No change requested. | - [x] |

No BLOCKER, HIGH, or MEDIUM findings in the PR itself. The PR is strictly a correctness improvement at the script layer and is consistent with existing precedents.

### Review

**Code quality**: The new code matches the shape of `post-note`/`add-note` in spirit and extends the pattern cleanly to the multi-field case. The comment block at line 127-130 explains why env vars are used instead of stdin, which is exactly the right amount of documentation. The `int(os.environ["ASSIGNEE_ID"])` conversion is the right move (prevents `["42"]` string-id bug that would have otherwise bit the first person to use `--assignee`).

**Safety**: Verified that:
- The single-quoted `<<'PY'` heredoc prevents bash from touching the payload.
- Env var pass-through keeps argv clean.
- `set -e` catches python-subshell failures in the `JSON_BODY=$(...)` capture.
- Python stderr still surfaces to the script's stderr when a failure occurs.
- All 8 smoke-test payloads round-trip through `json.loads` cleanly.

**Reproduction of issue #44 example**:

```
desc with quotes+backslash+newline+ctrl:
  {"title": "t", "description": "Has \"quotes\", \\backslash, newline\nand \u0001 control char"}
  VALID
```

The fix correctly handles the exact bug class #44 describes at the script layer.

**End-to-end note**: While the script layer is now watertight, the `.ag` caller still shell-concatenates LLM strings into `exec sh` command strings, so any `"` in `draft.title` gets mangled before reaching the script. This is exactly the condition #44 marked as a separate latent bug ("will become HIGH the moment anyone fixes the shell escaping in the .ag callers"). I have filed #49 as the follow-up to track the caller-layer fix. This is not a blocker for merging #48 — the script layer fix is strictly better than before and is a prerequisite for the end-to-end fix.

**Untouched endpoints confirmed unchanged**: `issues`, `issue`, `issue-events`, `issue-notes`, `members`, `labels`, and the error branch all regression-tested PASS.

**No lurking JSON concat elsewhere**: Grepped `dev-apprenticeship/**` for `BODY=.*\\"` and `-d "{` patterns — no matches. `code-review/approve` uses static literal `{}`, `post-note` and `add-note` already use `json.dumps`. There are no other endpoints that would benefit from this fix.

### Verdict

Ready to merge. The PR does exactly what it advertises at the layer it claims to fix, with clean tests and clean regression. Two follow-ups filed (#49 for the caller layer, #50 for the empty-body UX) to capture the items this PR deliberately does not address.

PR #56, opened 2026-04-11
fix(triage): use shell_escape for LLM strings in exec sh commands

### LINKED ISSUE #49: .ag callers shell-concatenate LLM strings into exec sh commands (follow-up to #44)
## Problem

Issue #44 was scoped to the JSON body concat in `dev-apprenticeship/triage/scripts/gitlab-api.sh`. That layer is now fixed in PR #48.

#44 explicitly noted the bug was latent because the `.ag` caller layer breaks on the same value first via shell word-splitting. Per #44 own severity rubric:

> Currently latent because the agent layer (.ag exec sh) breaks on the same value first when it tries to pass draft.title through shell word-splitting. Will become HIGH the moment anyone fixes the shell escaping in the .ag callers (which would then let broken JSON reach the server).

With PR #48 merged, the script layer is fixed but the caller layer is not. End-to-end round-trip of the issue-report example `Fix \"unterminated string\" error` still does not work.

## Reproduction

`dev-apprenticeship/triage/agents/issue_creator.ag` line 84 builds the command string by concatenating the LLM-returned title into a quoted template:

```
let create_cmd = \"./scripts/gitlab-api.sh create-issue --title \\\"\" + draft.title + \"\\\" --description \\\"\" + draft.description + \"\\\" --labels \\\"\" + draft.suggested_labels + \"\\\"\";
exec sh create_cmd;
```

If `draft.title = Fix \"unterminated string\" error`, the assembled string becomes:

```
./scripts/gitlab-api.sh create-issue --title \"Fix \"unterminated string\" error\" --description \"...\" --labels \"...\"
```

Shell tokenization collapses `\"Fix \"` + `unterminated` + `\" \"` into tokens `Fix unterminated`, `string error`, then the while-loop argparse silently drops the orphan tokens via its `*)` catchall.

Verified end-to-end:

```
$ bash gitlab-api.sh create-issue --title \"Fix \"unterminated string\" error\" --description \"body\" --labels \"bug\"
-> captured body: {\"title\": \"Fix unterminated\", \"description\": \"body\", \"labels\": \"bug\"}
```

The title is silently truncated to `Fix unterminated`. JSON is well-formed but wrong.

## Affected call sites

All `.ag` files that build exec sh commands by concatenating LLM output:

- `dev-apprenticeship/triage/agents/issue_creator.ag:84` (title, description, labels)
- `dev-apprenticeship/triage/agents/router.ag:86` (assignee username)
- `dev-apprenticeship/triage/agents/prioritizer.ag:83` (labels)
- `dev-apprenticeship/triage/agents/labeler.ag:84` (labels)
- All `post_cmd` construction sites in `code-review/agents/*.ag` and `planning/agents/*.ag` (bodies with markdown)

## Severity

HIGH. Per #44's own rubric this is now HIGH (no longer latent) because PR #48 removed the JSON-layer safety net that was absorbing the breakage.

## Suggested fix

Teach the `.ag` agents to pass values safely. Options:

1. Write values to a temp file or env var and reference those from the script (pattern match to how GITLAB_TOKEN is passed).
2. Add an `exec sh argv` form that takes argv as a list of strings instead of a single command string, so no shell word-splitting happens.
3. Have `gitlab-api.sh` read title/description from env vars or stdin so the caller never needs to shell-quote them.

Option 3 is the smallest change and mirrors the env-var pattern PR #48 uses internally.

## Out of scope

This is a separate follow-up to the script-layer fix in PR #48. Do not block #48 on this.

ylohnitram:
## Summary

- Wrap LLM-provided strings with `shell_escape()` in the 4 triage `.ag` agents that build `exec sh` command strings (issue_creator title/description/labels, router assignee, prioritizer priority, labeler labels).
- Closes the end-to-end round-trip bug where titles containing double quotes (e.g. `Fix "unterminated string" error`) were silently truncated by shell word-splitting before ever reaching `gitlab-api.sh`, and closes the latent shell-injection vector where a malicious title could execute arbitrary commands via backticks or `$(...)`.
- Uses the existing `shell_escape` builtin added in agentis-core v1.1.0 (Replikanti/agentis#454), matching the same pattern code-review and planning already use for post-note bodies. Triage was the last colony missing it.

Closes #49

## Why this divergence from the options listed in the issue

The original issue body suggested env vars, stdin, or a new `exec sh argv` form. All three require plumbing that `.ag` does not have today. `shell_escape` is a one-builtin-call-per-site fix that uses primitive infrastructure specifically built for this class of bug. Plan was posted on #49 for approval before implementation.

## Test plan

- [x] `agentis commit` parses all 4 edited `.ag` files cleanly
- [x] Python mirror of `shell_escape` + `bash -c` + mock curl: 10 round-trip cases pass (plain, double quotes, single quotes, backslashes, newlines, backtick+cmdsub injection, tabs/unicode, quote mix, multiline description, comma-separated labels)
- [x] Original #44/#49 repro `Fix "unterminated string" error` round-trips intact instead of truncating to `Fix unterminated`
- [x] `tools/colony-lint.sh dev-apprenticeship` passes
- [ ] QA agent review (will run after push)

## Out of scope

- Timestamp and numeric ID concatenations (`--since $ts`, `update-issue $iid`) are not shell-escaped because the values are either own-generated ISO 8601 strings or numeric IDs from parsed JSON, neither of which contain LLM taint or shell metacharacters.
- Code-review and planning already use `shell_escape` for their LLM-provided bodies, so no change needed there.

### DIFF
--- a/dev-apprenticeship/triage/agents/issue_creator.ag
+++ b/dev-apprenticeship/triage/agents/issue_creator.ag
@@ -81,7 +81,7 @@ fn tick(reason: string) -> void {
         ) -> DraftIssue;
 
         if confidence >= 0.85 {
-            let create_cmd = "./scripts/gitlab-api.sh create-issue --title \"" + draft.title + "\" --description \"" + draft.description + "\" --labels \"" + draft.suggested_labels + "\"";
+            let create_cmd = "./scripts/gitlab-api.sh create-issue --title " + shell_escape(draft.title) + " --description " + shell_escape(draft.description) + " --labels " + shell_escape(draft.suggested_labels);
             let result = try {
                 exec sh create_cmd;
             } catch e {
--- a/dev-apprenticeship/triage/agents/labeler.ag
+++ b/dev-apprenticeship/triage/agents/labeler.ag
@@ -81,7 +81,7 @@ fn tick(reason: string) -> void {
         ) -> LabelAction;
 
         if confidence >= 0.85 {
-            let update_cmd = "./scripts/gitlab-api.sh update-issue " + to_string(action.issue_id) + " --add-labels \"" + action.labels + "\"";
+            let update_cmd = "./scripts/gitlab-api.sh update-issue " + to_string(action.issue_id) + " --add-labels " + shell_escape(action.labels);
             let result = try { exec sh update_cmd; } catch e {
                 print("[labeler] Failed to apply labels:", e);
                 "";
--- a/dev-apprenticeship/triage/agents/prioritizer.ag
+++ b/dev-apprenticeship/triage/agents/prioritizer.ag
@@ -80,7 +80,7 @@ fn tick(reason: string) -> void {
         ) -> PriorityAction;
 
         if confidence >= 0.85 {
-            let update_cmd = "./scripts/gitlab-api.sh update-issue " + to_string(action.issue_id) + " --add-labels \"" + action.priority + "\"";
+            let update_cmd = "./scripts/gitlab-api.sh update-issue " + to_string(action.issue_id) + " --add-labels " + shell_escape(action.priority);
             let result = try { exec sh update_cmd; } catch e {
                 print("[prioritizer] Failed to set priority:", e);
                 "";
--- a/dev-apprenticeship/triage/agents/router.ag
+++ b/dev-apprenticeship/triage/agents/router.ag
@@ -83,7 +83,7 @@ fn tick(reason: string) -> void {
         ) -> RouteAction;
 
         if confidence >= 0.85 {
-            let update_cmd = "./scripts/gitlab-api.sh update-issue " + to_string(action.issue_id) + " --assignee \"" + action.assignee + "\"";
+            let update_cmd = "./scripts/gitlab-api.sh update-issue " + to_string(action.issue_id) + " --assignee " + shell_escape(action.assignee);
             let result = try { exec sh update_cmd; } catch e {
                 print("[router] Failed to assign:", e);
                 "";


ylohnitram:
## QA + Code Review Report

**Verdict: LGTM** — safe to merge.

### Summary

- Fix is correct, complete, and minimal.
- `shell_escape` POSIX single-quote wrapping is bulletproof against every injection vector I could construct.
- Pattern matches `code-review/` and `planning/` exactly. No drift.
- Smoke test `/tmp/smoke-49/run.py`: **10/10 pass**, including the original `Fix \"unterminated string\" error` repro and an active `bad\`id\`\$(rm -rf /)` command-substitution attempt.
- No call sites missed in `triage/`. Audited every `exec sh` in all 4 files.
- Out-of-scope items (`--since \$ts`, `update-issue \$iid`) are justified.

### Findings

| 1 | INFO | correctness | `shell_escape` correctly handles backticks, `\$(...)`, single quotes, backslashes, newlines, tabs, and Unicode. POSIX single-quoted strings have no escape processing, and every embedded `'` is pre-broken into `'\\''`, so the outer quoting invariant is never violated. | - [x] Verified |
| 2 | INFO | completeness | All 4 `exec sh <var>` sites in triage that concatenate an LLM return value are wrapped. Other `exec sh` in triage are either literal strings (`labels`, `members`, `date -u ...`) or `issues_cmd()` which concatenates a memo-store timestamp written by `date -u +%Y-%m-%dT%H:%M:%SZ` (never LLM-tainted). | - [x] Verified |
| 3 | INFO | consistency | Pattern matches `code-review/agents/*.ag` and `planning/agents/plan_reviewer.ag`: `\" --<flag> \" + shell_escape(value)`. Identical idiom. | - [x] Verified |
| 4 | INFO | smoke test | All 10 round-trip cases pass: plain, double quotes (original #44 repro), single quote, backslash, newline, backtick+cmdsub injection, tabs/unicode, quote-mix, multiline description, comma-separated labels. | - [x] Verified |
| 5 | INFO | scope | `--since \$ts` skipped correctly — `ts` is own-generated ISO 8601 from `date -u`, never LLM-tainted. `update-issue \$iid` skipped correctly — `issue_id` is typed `int` in the `RouteAction`/`PriorityAction`/`LabelAction` structs, so prompt-return coercion forces integer parsing before `to_string()` runs. | - [x] Verified |
| 6 | LOW | hardening | No lint rule prevents future regressions if a new `.ag` file forgets to wrap with `shell_escape`. Filed as **#57** for follow-up, not a merge blocker. | - [ ] #57 |

### Injection attempt I could not land

Tried to escape the outer single quotes with:

- `'; rm -rf /; '` — pre-broken by `'\\''` replacement into `'\\''; rm -rf /; '\\''`, which is 3 shell tokens (close, literal `; rm -rf /;`, re-open) all still inside the argv string the outer bash receives.
- Embedded literal `'\''` sequence in the input — the `shell_escape` pass escapes each `'` independently, so the user-provided `\\`, `'`, `'`, `'` get converted to `\\`, `'\\''`, `'\\''`, `'\\''` — safe.
- Backslash-escaped close quote `\\'` — POSIX single quotes do not honor backslash escapes at all, so the close quote remains literal and the overall outer quoting holds.

All of these round-trip through the smoke harness as literal title text.

### Notes on out-of-scope items

Both justifications hold up:

- **`--since \$ts`**: The memo store is local to the colony and only ever receives values from `memo_write(\"<agent>:last_check\", now)` where `now = exec sh \"date -u +%Y-%m-%dT%H:%M:%SZ\"`. A remote memo-store write is not a current capability on triage. Defensive escaping would be cheap but is not addressing a live path.
- **`update-issue \$iid`**: `action.issue_id` is declared `int` and the prompt result-coercion machinery parses it as an integer before the `.ag` code ever touches it. A malformed LLM return would throw at the `-> RouteAction` coercion, not reach `to_string()`.

### Test suite / lint

Not run — this PR touches only `.ag` files. `cargo test` and `cargo clippy` do not exercise colony scripts, and `tools/colony-lint.sh dev-apprenticeship` was already confirmed green in the PR description test plan.

### Recommendation

**Merge.** The fix is the smallest possible change that closes a HIGH severity shell-injection vector and also fixes the end-to-end round-trip bug for titles containing quotes. Follow-up hardening tracked in #57.

---

*Auto-generated by QA + PR review run on branch `fix/triage-shell-escape-49`.*

PR #51, opened 2026-04-11
fix(gitlab-api): strict flag rejection and drop dead code-review endpoints

ylohnitram:
## Summary

Hardening pass over all three `gitlab-api.sh` wrappers in the dev-apprenticeship federation. Two bugs bundled because they touch the same files and overlap.

**#16** — every endpoint parsed its own flags with a `case` block that silently swallowed unknowns via `*) shift ;;`. A typo like `--sinse 2026-04-01` would discard the flag plus its value and cheerfully run with defaults. LLM-authored `.ag` agents are the primary callers and can absolutely produce such typos.

Replaces all 8 fallbacks across planning, code-review, and triage with:
```bash
*) echo "{\"error\": \"unknown flag: $1\"}" >&2; exit 2 ;;
```
Exit code 2 distinguishes usage error from exit 1 (generic error). JSON error envelope keeps the error shape consistent with the rest of the script.

**#18** — four endpoints in `dev-apprenticeship/code-review/scripts/gitlab-api.sh` were never called by any `.ag` agent in that colony:

- `mr` (no callers)
- `mr-approvals` (no callers)
- `members` (no callers)
- `issues` (code-review reviews merge requests, not issues)

Verified via grep across all 5 code-review agents (`approval_decider`, `logic_reviewer`, `security_reviewer`, `style_reviewer`, `test_reviewer`). Removed the four case arms and the matching header usage lines. Remaining active endpoints: `merge-requests`, `mr-changes`, `mr-notes`, `post-note`, `approve` (plus the unknown-command fallback which stays).

The reason for bundling: `#18` removes the `issues` case in code-review which also had a `*) shift ;;` fallback, so the two fixes touch the same 11 lines. Doing them in one pass keeps history clean.

Closes #16
Closes #18

## Test plan

- [x] `tools/colony-lint.sh` -> 30 passed, 0 failed
- [x] Mock-curl smoke test, 8 cases:
  - valid flag still works (regression)
  - typo flag returns exit 2 with JSON error body
  - unknown top-level command still returns exit 1 (unchanged)
  - deleted code-review endpoints return `{"error": "Unknown command: ..."}` exit 1 (not crash)
  - code-review active endpoints all still work
  - typo flag rejection works in planning and triage too
- [x] Diff stat: 3 files changed, 8 insertions, 48 deletions (mostly the four deleted code-review case arms)
- [ ] QA report posted as comment

## Sample smoke test output

```
=== B. typo flag rejected ===
{"error": "unknown flag: --sinse"}
[...]

PR #58, opened 2026-04-11
fix(triage): reject update-issue with no modifying flags

### LINKED ISSUE #50: triage: update-issue with no modifying flags produces empty JSON body {}
## Problem

`dev-apprenticeship/triage/scripts/gitlab-api.sh update-issue <id>` with no flags after the id produces a PUT with body `{}`. This is valid JSON and technically a no-op, but GitLab may reject it depending on instance config, and the UX is misleading because the script returns the unchanged issue JSON as if the update had happened.

Observed via mock smoke test against PR #48:

```
$ bash gitlab-api.sh update-issue 99
-> PUT /projects/.../issues/99 with body {}
```

## Expected behavior

Either:
1. Fail fast with `{\"error\": \"update-issue requires at least one of --add-labels, --remove-labels, --priority, --assignee\"}` and exit 1, OR
2. Short-circuit to a no-op and return `{\"ok\": true, \"noop\": true}` without making the API call.

Option 1 matches the existing `create-issue --title is required` error path and is the smaller change.

## Severity

LOW. Not a blocker for any current caller because the `.ag` agents always pass at least one flag. Flagged in PR #48 QA as a UX improvement.

## Out of scope

This is a UX polish follow-up. The underlying JSON body fix landed in #48.

ylohnitram:
## Summary

- Make `./scripts/gitlab-api.sh update-issue <id>` with no modifying flags fail fast with a clear error instead of sending a confusing `PUT {}` that GitLab silently accepts.
- Distinguish the unknown-user case: when `--assignee ghost` resolves to no user, emit `unknown user: ghost` rather than the generic "requires at least one of ..." error (which would be misleading since the caller *did* pass `--assignee`).
- Both errors flow through the `emit_error` helper from #51, so usernames containing quotes, backslashes, or newlines produce valid JSON output.

Closes #50

## Change shape

After the user-ID lookup, before the python3 body build:
1. If `--assignee` was passed but `USER_ID` is empty → `unknown user: <name>`, exit 1.
2. If all four modifier slots (`ADD_LABELS`, `REMOVE_LABELS`, `PRIORITY`, `USER_ID`) are empty → `requires at least one of ...`, exit 1.

Matches the existing `create-issue --title is required` error path for symmetry.

## Test plan

Smoke-tested against a mocked curl (11 cases, all pass):

- [x] `update-issue 99` (no flags) → exit 1, "requires at least one" error, no PUT sent
- [x] `update-issue 99 --priority high` → exit 0, body `{"priority":"high"}`
- [x] `update-issue 99 --add-labels bug` → exit 0, body `{"add_labels":"bug"}`
- [x] `update-issue 99 --remove-labels stale` → exit 0, body `{"remove_labels":"stale"}`
- [x] `update-issue 99 --assignee alice` (known user) → exit 0, body `{"assignee_ids":[1]}`
- [x] `update-issue 99 --add-labels bug --priority high` (multi) → exit 0, both fields in body
- [x] `update-issue 99 --assignee alice --priority high --add-labels bug` (three together) → exit 0, all three in body
- [x] `update-issue 99 --assignee ghost` (unknown user) → exit 1, "unknown user: ghost"
- [x] All error outputs validate as JSON (including for `--nosuch x` flag)
- [x] `tools/colony-lint.sh dev-apprenticeship` passes
- [x] shellcheck clean
- [ ] QA agent review (will run on PR)

### DIFF
--- a/dev-apprenticeship/triage/scripts/gitlab-api.sh
+++ b/dev-apprenticeship/triage/scripts/gitlab-api.sh
@@ -175,6 +175,22 @@ PY
             # `+`, `&`, or non-ASCII characters survive encoding intact.
             USER_JSON=$(gl_get_q "$GITLAB_URL/api/v4/users" --data-urlencode "username=$ASSIGNEE")
             USER_ID=$(echo "$USER_JSON" | grep -o '"id":[0-9]*' | head -1 | cut -d: -f2)
+            # If the lookup didn't find a user, fail here rather than
+            # silently dropping the assignee. Without this the caller gets
+            # the generic "requires at least one of ..." error below, which
+            # is confusing when they did pass --assignee.
+            if [ -z "$USER_ID" ]; then
+                emit_error "update-issue: unknown user: $ASSIGNEE"
+                exit 1
+            fi
+        fi
+        # Fail fast if no modifying flags were passed. A PUT with body {}
+        # is a confusing no-op: GitLab happily returns the unchanged issue
+        # as if the update had happened, so the caller has no signal that
+        # they forgot to pass anything.
+        if [ -z "$ADD_LABELS" ] && [ -z "$REMOVE_LABELS" ] && [ -z "$PRIORITY" ] && [ -z "$USER_ID" ]; then
+            emit_error "update-issue requires at least one of --add-labels, --remove-labels, --priority, --assignee"
+            exit 1
         fi
         # Build JSON body via python3 json.dumps so label names and priority
         # strings containing quotes, backslashes, commas, or control chars


ylohnitram:
## Combined QA + Code Review — PR #58

### Summary

- **Script QA**: 11/11 smoke cases pass (all matching the PR description)
- **Audit scripts**: N/A (this PR is in `dev-apprenticeship/`, not the Rust core)
- **shellcheck**: clean (koalaman/shellcheck:stable via docker, exit 0)
- **colony-lint**: clean (`tools/colony-lint.sh dev-apprenticeship`, 1 passed / 5 skipped / 0 failed)
- **JSON safety**: verified with 5 malicious username payloads
- **Regression risk**: none — all 3 `.ag` callers pass a modifying flag unconditionally
- **Completeness**: no other unchecked write commands exist across the 3 colony gitlab-api.sh scripts

### Verdict

**LGTM** — no BLOCKER or HIGH findings. Safe to merge.

### Findings

| 1 | INFO | design | Unknown-user guard runs *before* the "at least one modifier" guard, so `--assignee ghost --add-labels bug` aborts with `unknown user: ghost` and never applies the labels. Fail-fast is correct here. | `update-issue 99 --assignee ghost --add-labels bug` | - [x] verified |
| 2 | INFO | design | Empty `--assignee ""` trips the generic "requires at least one of" guard (not `unknown user: `) because `[ -n "$ASSIGNEE" ]` skips the lookup. Pre-existing behavior, unchanged by this PR. | `update-issue 99 --assignee ""` | - [x] noted |
| 3 | LOW | docs | The inline comment on lines 178-181 says "Without this the caller gets the generic 'requires at least one of ...' error below" — technically true only when `--assignee` is the *only* flag. With e.g. `--assignee ghost --add-labels bug` and no guard, the caller would silently get a labels-only PUT. Worth refining to "Without this, an unknown user either triggers the generic guard or silently drops the assignee while applying other flags". | N/A | - [ ] follow-up |
| 4 | LOW | docs | `create-issue` in triage still uses the legacy hard-coded `echo '{\"error\": \"--title is required\"}' >&2` path instead of `emit_error`. Same for `post-note` (code-review) and `add-note` (planning). All three are safe because the message is a static literal, but it's an inconsistency with the rest of the error-emission story from #51. | N/A | - [ ] follow-up |

### Severity Guide

- **BLOCKER**: must fix before merge
- **HIGH**: should fix before merge
- **MEDIUM**: fix in follow-up
- **LOW**: nice-to-have
- **INFO**: confirmed behavior, no action needed

---

### Detailed verification

#### 1. Guard ordering trace

Walked every input through both guards:

| Input | Guard 1 (unknown user) | Guard 2 (no modifier) | Final result |
|-------|------------------------|-----------------------|--------------|
| no flags | skipped (ASSIGNEE empty) | trips | exit 1 generic |
| `--priority high` | skipped | passes | PUT `{\"priority\":\"high\"}` |
| `--add-labels bug` | skipped | passes | PUT `{\"add_labels\":\"bug\"}` |
| `--remove-labels stale` | skipped | passes | PUT `{\"remove_labels\":\"stale\"}` |
| `--assignee alice` (known) | USER_ID=1, passes | passes | PUT `{\"assignee_ids\":[1]}` |
| `--assignee ghost` (unknown) | USER_ID="", trips | never reached | exit 1 unknown user |
| `--assignee ghost --add-labels bug` | trips | never reached | exit 1 unknown user |
| `--add-labels bug --priority high` | skipped | passes | PUT with both fields |
| `--assignee alice --priority high --add-labels bug` | USER_ID=1, passes | passes | PUT with 3 fields |
| `--nosuch x` | never reached | never reached | exit 2 unknown flag |

The unknown-user-wins-over-labels behavior is **correct** for this use case. Rationale:
- The router agent is the only caller that passes `--assignee` today, and it only passes `--assignee` (no labels), so the question of label-without-user-on-failure is hypothetical.
- If a future caller mixes `--assignee` with labels, treating an unknown user as a partial failure is safer than silently degrading. A partial success ("assigned no one but applied labels") is exactly the kind of silent failure issue #50 was written to eliminate.

#### 2. JSON safety verification

Ran 5 adversarial payloads through `--assignee`, each produces valid JSON on stderr:

\`\`\`
--assignee 'evil\"name'       -> {\"error\": \"update-issue: unknown user: evil\\\"name\"}
--assignee 'back\\\\slash'     -> {\"error\": \"update-issue: unknown user: back\\\\\\\\slash\"}
--assignee \$'has\\nnewline'    -> {\"error\": \"update-issue: unknown user: has\\\\nnewline\"}
--assignee '\",\"injected\":\"x' -> {\"error\": \"update-issue: unknown user: \\\",\\\"injected\\\":\\\"x\"}
--assignee \$'t\\tc\\x01'        -> {\"error\": \"update-issue: unknown user: t\\\\tc\\\\u0001\"}
\`\`\`

Each parses as JSON with exactly one key (`error`). The JSON-injection attempt is neutralized — no second top-level key leaks through.

#### 3. Smoke test results (11/11)

All 11 cases from the PR description pass. Rebuilt `/tmp/smoke-50/bin/curl` from scratch as a mock that returns `alice -> id 1`, everything else `[]`, and captures the PUT body to `body.json`.

- Case 1 (no flags): exit 1, generic error, body empty (no PUT sent) — ok
- Case 2 (`--priority high`): exit 0, body `{\"priority\": \"high\"}` — ok
- Case 3 (`--add-labels bug`): exit 0, body `{\"add_labels\": \"bug\"}` — ok
- Case 4 (`--remove-labels stale`): exit 0, body `{\"remove_labels\": \"stale\"}` — ok
- Case 5 (`--assignee alice`): exit 0, body `{\"assignee_ids\": [1]}` — ok
- Case 6 (`--add-labels bug --priority high`): exit 0, both fields in body — ok
- Case 7 (three flags): exit 0, all three fields in body — ok
- Case 8 (`--assignee ghost`): exit 1, `unknown user: ghost` — ok
- Case 9 (`--nosuch x`): exit 2, `unknown flag: --nosuch` (unchanged by this PR) — ok
- Case 10 (`--assignee ghost --add-labels bug`): exit 1, `unknown user: ghost`, no PUT — ok
- Case 11 (`--assignee alice`, second run): exit 0, body `{\"assignee_ids\": [1]}` — ok

#### 4. Regression risk across `.ag` callers

\`\`\`
router.ag:86       update-issue + --assignee <username>    [1 modifying flag]
prioritizer.ag:83  update-issue + --add-labels <priority>  [1 modifying flag]
labeler.ag:84      update-issue + --add-labels <labels>    [1 modifying flag]
\`\`\`

All 3 always pass exactly one modifying flag. None are at risk from the new guards. Confirmed none were missed — grep for `update-issue` across `dev-apprenticeship/triage/agents/*.ag` returns exactly these 3 sites.

Side note (not a finding): the prioritizer passes its LLM-suggested priority into `--add-labels`, not `--priority`. That's a pre-existing choice and unrelated to this PR.

One edge case worth noting: if the LLM ever returns an empty string for `action.labels`, the agent would call `update-issue 99 --add-labels ''`, which now fails with the generic "requires at least one of" error instead of silently PUTting `{}`. That's the intended behavior and strictly better than before.

#### 5. shellcheck (docker koalaman/shellcheck:stable)

\`\`\`
\$ docker run --rm -i koalaman/shellcheck:stable - < dev-apprenticeship/triage/scripts/gitlab-api.sh
\$ echo \$?
0
\`\`\`

Clean.

#### 6. colony-lint

\`\`\`
\$ tools/colony-lint.sh dev-apprenticeship
[PASS] README.md: links OK
Agentis found: agentis v1.1.3
Results: 1 passed, 0 failed, 5 skipped
\`\`\`

#### 7. Completeness audit: other no-op-body commands across the 3 colonies

Enumerated every write command across all 3 colony gitlab-api.sh scripts:

| Colony | Command | Required-field guard? |
|--------|---------|-----------------------|
| triage | `create-issue` | yes — `--title is required` (line 133-136) |
| triage | `update-issue` | **now yes** (this PR) |
| code-review | `post-note` | yes — `--body is required` (line 112-115) |
| code-review | `approve` | N/A — intentionally body-less POST to `/approve` endpoint |
| planning | `add-note` | yes — `--body is required` (line 121-124) |

All write commands with body requirements are now guarded. `approve` is body-less by design (it's a status-change endpoint, POST with `{}`). No other no-op-body holes exist.

---

### Follow-ups to file as separate issues

- [ ] LOW: refine inline comment on lines 178-181 to cover the label-dropping edge case
- [ ] LOW: migrate `create-issue --title`, `post-note --body`, `add-note --body` error paths to `emit_error` for consistency with #51

I will file both as issues per the auto-followup policy.

PR #66, opened 2026-04-11
feat(lint): flag unsafe string concat into exec sh (#57)

### LINKED ISSUE #57: colony-lint: forbid unescaped string concatenation into exec sh commands
## Context

Follow-up to #49 / PR #56. We now have 4 separate colony families (`triage`, `code-review`, `planning`) that all build shell command strings by concatenating values into a template, and we rely on every new call site remembering to wrap LLM-tainted values with `shell_escape()`. That is fragile. One new `.ag` file that forgets the wrap reintroduces the shell-injection regression.

## Proposal

Add a `tools/colony-lint.sh` rule that scans every `.ag` file under `dev-apprenticeship/` for any string-literal concatenation into a variable that is later passed to `exec sh`, and fails if the concatenated expression is not one of:

- a string literal
- `shell_escape(<expr>)`
- `to_string(<expr>)` where the target is a typed `int` field
- `recall_latest(...)` for a known-ISO8601 memo key (allowlist)

Concretely, flag any occurrence of `exec sh <var>` where `<var>` is assigned from a concatenation that contains a bare `+ <nonliteral>` without a safe wrapper. This should be a grep-level check, not a full parser.

## Priority

LOW. Not an active bug, but it is the kind of guardrail that would have caught #49 before it shipped.

## Out of scope

- Teaching `exec sh` itself to take an argv list. That was explored in #49 discussion and deferred because it requires evaluator plumbing.

ylohnitram:
## Summary

- New `tools/check-exec-sh.sh` scans every `.ag` file and flags any `let <var> = "..." + <segment>` where `<var>` is later passed to `exec sh` and `<segment>` is not a string literal, `shell_escape(...)`, or `to_string(...)`
- Wired into `tools/colony-lint.sh` so CI fails any PR that reintroduces the #49-class shell-injection bug
- New `tools/test-check-exec-sh.sh` with 14 unit tests covering safe patterns, unsafe patterns, opt-out comment, function-call indirection, and a regression test against the current `dev-apprenticeship/` codebase

This is a grep-level guardrail, not a full parser. Indirection through function calls (e.g. `let cmd = issues_cmd()`) is deliberately not followed, per the scope in #57. Authors can suppress a finding the checker cannot prove safe by adding `// colony-lint: safe-exec-concat` on the same line or the preceding line.

## Test plan

- [x] `tools/test-check-exec-sh.sh` runs 14 assertions, all pass
- [x] `tools/colony-lint.sh` full run reports `32 passed, 0 failed, 5 skipped` with the new check integrated
- [x] Regression: current `dev-apprenticeship/` codebase passes cleanly (every direct concat already uses `shell_escape` or `to_string`)
- [x] Deliberate-break: injecting a fixture `.ag` with `"..." + draft.title` and no wrap triggers `[UNSAFE]` and exit 1
- [x] `shellcheck` clean on `check-exec-sh.sh`, `test-check-exec-sh.sh`, and the modified `colony-lint.sh`

Closes #57

### DIFF
--- a/tools/check-exec-sh.sh
+++ b/tools/check-exec-sh.sh
@@ -1,0 +1,160 @@
+#!/bin/bash
+# tools/check-exec-sh.sh: Flag unsafe string concatenation into `exec sh` in .ag files.
+#
+# Colony agents build shell commands by concatenating strings and pass the
+# result to `exec sh`. LLM-tainted values (titles, descriptions, usernames,
+# label names) must be wrapped in `shell_escape(...)` before concatenation
+# or the agent is vulnerable to shell injection. This was the root cause of
+# issue #49 / PR #56 in the triage colony.
+#
+# This checker is a grep-level guardrail: for every variable that is later
+# passed to `exec sh`, it scans `let <var> = ...` assignments in the same
+# file and flags any `+` concatenation segment that is not one of:
+#
+#   - a string literal (starts with `"`)
+#   - `shell_escape(...)`
+#   - `to_string(...)`
+#
+# Indirection through function calls (e.g. `let cmd = issues_cmd()`) is not
+# followed — this is deliberate, per #57 ("grep-level check, not a full
+# parser"). Authors can suppress a finding on a line that the checker
+# cannot prove safe by adding `// colony-lint: safe-exec-concat` to the
+# same line or the preceding line.
+#
+# Usage: ./tools/check-exec-sh.sh [path]
+# Exit 0 if no unsafe patterns, 1 if one or more findings, 2 on usage error.
+
+set -euo pipefail
+
+SCAN_ROOT="${1:-$(cd "$(dirname "$0")/.." && pwd)}"
+
+if [ ! -e "$SCAN_ROOT" ]; then
+    echo "check-exec-sh: scan root does not exist: $SCAN_ROOT" >&2
+    exit 2
+fi
+
+FAIL=0
+
+# is_safe_segment <trimmed_segment>
+# Return 0 if the segment is one of the known-safe forms, 1 otherwise.
+is_safe_segment() {
+    local seg="$1"
+    case "$seg" in
+        '"'*)            return 0 ;;  # string literal
+        'shell_escape('*) return 0 ;;
+        'to_string('*)   return 0 ;;
+    esac
+    return 1
+}
+
+# trim <string>
+trim() {
+    local s="$1"
+    # strip leading whitespace
+    s="${s#"${s%%[![:space:]]*}"}"
+    # strip trailing whitespace
+    s="${s%"${s##*[![:space:]]}"}"
+    printf '%s' "$s"
+}
+
+# check_file <path>
+# Scan one .ag file. Prints findings to stdout. Returns 0 always; the
+# caller tracks FAIL via the global counter.
+check_file() {
+    local ag_file="$1"
+
+    # Collect unique identifiers appearing on the RHS of `exec sh`.
+    # We intentionally only match bare identifiers — `exec sh "literal"`
+    # (with a string literal) is always safe and never flagged.
+    # Trailing `|| true` so that a file with zero matches does not trip
+    # `set -e` on the grep-returns-1 path.
+    local exec_vars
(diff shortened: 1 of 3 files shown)


ylohnitram:
## Combined QA + Review Report

**Verdict: LGTM.** CI green, 14/14 unit tests pass locally, 32 passed / 0 failed / 5 skipped on the full `colony-lint.sh` run, deliberate break produces the expected `[UNSAFE]` finding and exit 1.

### What I ran

- `git checkout feat/colony-lint-exec-sh-57` + `gh pr diff 66` — only `tools/check-exec-sh.sh`, `tools/test-check-exec-sh.sh`, `tools/colony-lint.sh` touched. No sneaky `.ag` or `gitlab-api.sh` changes.
- `tools/test-check-exec-sh.sh` — 14 passed, 0 failed.
- `tools/colony-lint.sh` — `32 passed, 0 failed, 5 skipped`, with `[PASS] check-exec-sh: no unsafe concat into exec sh` and `[PASS] tools: test-check-exec-sh.sh unit tests` both present.
- Deliberate break: fixture with `let cmd = "./scripts/gitlab-api.sh update-issue " + draft.title;` → `[UNSAFE] /tmp/.../bad.ag:2: let cmd = ... + draft.title (...)`, exit 1. File, line, var, and segment all correct.
- Opt-out preceding-line probe: comment on line 3, concat on line 4 → suppressed, exit 0. Comment 2 lines above (blank between) → still flagged. "Immediately preceding" semantic is enforced as documented.
- `gh pr checks 66` — `colony-lint pass 16s`.

### Findings

| 1 | INFO | correctness | Literal `+` inside a single-line `let` assignment gets flagged as unsafe concat. Acknowledged in PR body as a limitation. | `let cmd = "./foo.sh --query a+b"; exec sh cmd;` → `[UNSAFE] ... + b"` | - [x] documented |
| 2 | INFO | correctness | `to_string(stringvar)` is an identity and does not shell-escape, but the checker trusts any `to_string(...)` segment. Pre-noted in PR description. | `let cmd = "..." + to_string(body);` where `body: string` from `prompt()` → exit 0 | - [x] accepted design |
| 3 | INFO | correctness | Multi-line concat `let cmd = "..."\n    + draft.title;` is a silent miss. grep is line-oriented. | `let cmd = "./foo "\n    + draft.title;\nexec sh cmd;` → exit 0 | - [x] accepted limitation |
| 4 | INFO | correctness | Function-call indirection is intentionally not followed, per #57 scope. `let cmd = build(raw);` where `build` concats unsafely is silent. | `let c = helper(); exec sh c;` → exit 0 | - [x] accepted design |
| 5 | INFO | style | PR body, commit message: English, no Avvoka, no em dashes, no semicolons. Clean. | N/A | - [x] pass |
| 6 | INFO | test | `test-check-exec-sh.sh` test 8 (`optout-preceding`) exercises the correct line — opt-out on the line immediately above the concat. Verified against an external fixture with exact line numbers. | N/A | - [x] pass |

Severity guide: BLOCKER / HIGH / MEDIUM / LOW / INFO. No BLOCKER / HIGH / MEDIUM / LOW findings.

### Notes for the record

- The checker's docstring already documents the three false-negative cases (function-call indirection, multi-line, `+`-in-string-literal). The `to_string` identity-on-string case is not documented in the checker header and is the weakest spot of the allowlist. It is acceptable as-is because `to_string` on a string is a no-op at runtime and authors who route LLM output through it will at least see a visible function call in code review — the lint is not the last line of defence.
- The `shell_escape(...)` prefix match allows `shell_escape(` to match any opening paren at the start of the segment, so expressions like `shell_escape(a + b)` parse as two segments (`shell_escape(a`, `b)`) and would trigger `[UNSAFE]` on `b)`. Not exercised in the current codebase and can be suppressed with the opt-out comment, but worth knowing. No action required for this PR.
- `colony-lint.sh` gates the new section on `[ -x "$REPO_ROOT/tools/check-exec-sh.sh" ]`, so a stale checkout without exec bits would silently skip instead of failing. Acceptable — the tools-lint loop already ensures every `tools/*.sh` is shellcheck-clean and executable.

### Filed follow-ups

- #67 — LOW: document the four known limitations of `check-exec-sh.sh` (literal-plus false positive, multi-line miss, `to_string` identity, function-call indirection). Optional enhancement: teach the splitter to skip `+` inside `"..."` segments to eliminate the one false positive.

**Merge when ready.**

PR #73, opened 2026-04-12
tools: harden colony-lint with daemon flag guardrail and known-limitations docs

### LINKED ISSUE #67: colony-lint: document known limitations of check-exec-sh.sh
Follow-up from QA on PR #66.

The grep-level checker in `tools/check-exec-sh.sh` has four known false-positive / false-negative cases that authors will encounter over time. They were acknowledged in the PR body but are not visible in the checker output or in a docs page. File-level awareness would cut noise-driven suppressions.

Known cases:

1. **False positive on literal `+` inside a single-line `let` assignment.** A pure literal like `let cmd = \"./foo.sh --query a+b\"; exec sh cmd;` is flagged, because the checker splits on `+` after seeing one in `rhs`. Reproducer:

    ```
    fn tick(reason: string) -> void {
        let cmd = \"./scripts/search.sh --query foo+bar\";
        let out = try { exec sh cmd; } catch e { \"\"; };
    }
    ```

    Result: `[UNSAFE] ... let cmd = ... + bar\"`. Current workaround: `// colony-lint: safe-exec-concat`.

2. **False negative on multi-line concat.** `let cmd = \"./foo \"\n    + draft.title;` is not flagged because grep is line-oriented and the `+` lives on a different line than the `let`. Silent miss.

3. **False negative on `to_string(body)` where `body` is already a string.** `to_string` is an identity on strings and does not shell-escape, but the checker trusts any `to_string(...)` segment.

4. **False negative on function-call indirection.** `let cmd = build_cmd(raw_title);` where `build_cmd` internally concatenates without `shell_escape` is not followed. This is explicit design per #57 but worth listing alongside the others.

Suggested follow-ups (any one is fine):

- Add a `tools/check-exec-sh.md` (or a block in `README.md`) listing the four cases.
- Print a one-line `check-exec-sh: known limitations: see tools/check-exec-sh.sh header` hint when the checker runs in verbose mode.
- Teach the naive splitter to skip `+` characters inside double-quoted segments — a few lines of awk would resolve case 1 and drop the false-positive rate to near zero.

Severity: LOW. The current behaviour is correct against its stated scope and the opt-out comment is a clean escape hatch.

### LINKED ISSUE #71: colony-lint.sh: add DAEMON_FLAGS whitelist guardrail for start-colony.sh scripts
## Problem

\`tools/colony-lint.sh\` currently validates colony scaffolds by running:

- structure check (README, config, scripts, agents dir)
- TOML parse of \`config/colony.example.toml\`
- shellcheck on \`.sh\` files
- markdown link check
- \`agentis commit\` syntax check on \`.ag\` files
- \`tools/check-exec-sh.sh\` guardrail against unsafe \`exec sh\` concat

What it does **not** do: verify that the \`agentis daemon\` flag tokens used by each \`start-colony.sh\` are actually in the \`DAEMON_FLAGS\` whitelist from \`agentis-core/src/cli/daemon.rs\`.

This is the exact drift class that #68 was filed for. Five dev-apprenticeship start scripts were passing \`--backend\` and three were passing \`--enable-exec\`, both of which have never been real \`agentis daemon\` flags. They slipped past lint and CI because lint has no notion of the core flag contract.

## Fix

Add a new guardrail section to \`colony-lint.sh\` that:

1. Parses every \`agentis daemon\` invocation in every \`start-colony.sh\` under the federation.
2. Extracts the flag tokens (long flags beginning with \`--\`, plus short flags like \`-h\`).
3. Compares each against a known-good allowlist.
4. Fails the lint if any flag is outside the allowlist.

### Where does the allowlist come from?

Two options, from strictest to loosest:

- **Option A — sourced from core**: read the allowlist from \`agentis daemon --help\` (the usage string now lists every flag after PR #475). Pro: always in sync with core. Con: requires agentis binary to be installed during lint.
- **Option B — hardcoded with a comment pointing at core**: maintain a literal allowlist in the lint script with a comment saying \"must match \`DAEMON_FLAGS\` in agentis-core/src/cli/daemon.rs\". Pro: no binary dependency. Con: drifts silently if core adds a new flag.

Option A is strictly better if the agentis binary is already required by lint (it is, for the \`agentis commit\` syntax check at line 213). Fallback to option B if agentis is not installed (already a SKIP path in the lint).

## Severity

LOW. This is a prevention improvement, not a correctness fix. The drift class it catches was fixed retroactively in #68 (dev-apprenticeship) and #NEW (new-colony.sh scaffolder follow-up).

## Discovered by

QA review of #69. Filed automatically per the QA auto-follow-up rule.

ylohnitram:
## Summary

Two follow-ups from the #66/#69 QA round, bundled since both touch `tools/`.

### Closes #71 — DAEMON_FLAG_ALLOWLIST guardrail in colony-lint.sh

Adds a new check to `tools/colony-lint.sh` that walks every `agentis daemon` invocation in every `start-colony.sh` and validates each flag token against a hardcoded allowlist mirroring `DAEMON_FLAGS` from `agentis-core/src/cli/daemon.rs` as of v1.1.5.

Implementation notes:

- Small awk state machine reconstructs multi-line backslash-continued `agentis daemon` blocks
- Extracts `--foo`, `-X`, and `--foo=value` forms (the last via `sub(/=.*/)`)
- Compares each token against the allowlist; unknown tokens fail the lint with a precise message
- Runs inside the existing per-colony loop — one PASS per colony when clean

Chose the hardcoded allowlist (Option B from the issue) over parsing `agentis daemon --help` (Option A) because the help string is not a stable contract and the allowlist only changes once per phase. The check runs standalone and catches drift regardless of whether the agentis binary is installed.

### Closes #67 — check-exec-sh.sh known limitations docs

Adds `tools/check-exec-sh.md` with four sections documenting:

1. **False positive** — literal `+` inside a single-line `let` (workaround: `// colony-lint: safe-exec-concat`)
2. **False negative** — multi-line `+` concatenation on separate lines
3. **False negative** — `to_string(body)` when `body` is LLM-tainted
4. **False negative** — function-call indirection (explicit non-goal per #57)

Each section explains when it fires, why it is wrong, and the escape hatch. A pointer line is added to the header comment of `tools/check-exec-sh.sh` so authors finding a false positive know where to read.

## Test plan

- [x] `./tools/colony-lint.sh` — **37 passed / 0 failed / 5 skipped** (was 32, +5 for the 5 dev-apprenticeship colonies each getting a new daemon-flags PASS)
- [x] Negative test: injected `--backend claude` into a `start-colony.sh`, re-ran lint → `FAIL ... unknown daemon flag: --backend`, reverted
- [x] Edge case: injected `--tick-interval=60000` (`=value` form) → PASS (correctly strips `=value` before allowlist check)
- [x] Short-flag negative test: injected `-X` → `FAIL ... unknown daemon flag: -X`, reverted
- [x] No markdown lint errors on the new `tools/check-exec-sh.md`

### DIFF
--- a/tools/check-exec-sh.md
+++ b/tools/check-exec-sh.md
@@ -1,0 +1,88 @@
+# check-exec-sh.sh known limitations
+
+`tools/check-exec-sh.sh` is a grep-level guardrail against the
+shell-injection class of bug that slipped through in agentis-colonies#49.
+It scans every `.ag` file and flags unsafe `+` concatenation segments
+that land in an `exec sh` command without a `shell_escape(...)` or
+`to_string(...)` wrap.
+
+Because it is grep-level by design (per #57), it has a short list of
+known false positives and false negatives. Authors hitting one of
+these cases can suppress a finding on a specific line by adding
+`// colony-lint: safe-exec-concat` to the same line or the line
+immediately above.
+
+## 1. False positive: literal `+` inside a single-line `let`
+
+Symptom:
+```
+[UNSAFE] tick.ag:12: let cmd = "./scripts/search.sh --query foo+bar";
+```
+
+The checker splits on `+` after seeing one in the right-hand side of a
+`let`, so a plain string literal that happens to contain a `+`
+character is flagged even though nothing is being concatenated.
+
+Workaround:
+```
+let cmd = "./scripts/search.sh --query foo+bar"; // colony-lint: safe-exec-concat
+```
+
+Long-term fix candidate: teach the splitter to skip `+` characters
+inside double-quoted segments. A few lines of awk would drop this
+false-positive rate to near zero.
+
+## 2. False negative: multi-line concatenation
+
+The checker is line-oriented. This pattern is silently missed:
+
+```
+let cmd = "./foo "
+    + draft.title;
+```
+
+Because grep sees `let cmd = "./foo "` on one line and `+ draft.title`
+on another, the right-hand side never triggers the concat check.
+Review your own `.ag` files manually for this pattern, especially
+after merging a PR that reformats long lines.
+
+## 3. False negative: `to_string(body)` when `body` is already a string
+
+`to_string` is treated as a known-safe wrapper. It is an identity on
+string values and does not shell-escape. If an LLM-tainted value flows
+through `to_string` before concatenation, the checker trusts it even
+though the value is still dangerous:
+
+```
+let cmd = "gh issue create --title " + to_string(issue_title);
+```
+
+If `issue_title` is LLM-tainted, this is a shell injection even though
+the checker passes it. Use `shell_escape(issue_title)` for
+user-controlled or LLM-controlled values.
+
+## 4. False negative: function-call indirection
+
+The checker does not follow function calls. This is explicit design
+per agentis-colonies#57 ("grep-level check, not a full parser"):
+
+```
+let cmd = build_cmd(raw_title);
+exec sh cmd;
+```
+
+If `build_cmd` internally concatenates `raw_title` into a shell
+command without `shell_escape`, the injection is invisible to the
+checker. The rule of thumb is: keep `exec sh` call sites close to the
+`let` that builds the command, and don't introduce helper functions
+just for command construction.
+
+## When to use the suppression comment
+
+Add `// colony-lint: safe-exec-concat` only when you have manually
+verified that the flagged concatenation is safe. The most common
+legitimate case is case 1 above: a literal `+` inside a hardcoded
+string. If you find yourself using the suppression for case 2, 3, or
+4, that is a signal to refactor instead — pull the value through
+`shell_escape(...)`, inline the helper, or rewrite the `let` to
+appease the grep.
--- a/tools/check-exec-sh.sh
+++ b/tools/check-exec-sh.sh
@@ -21,6 +21,10 @@
 # cannot prove safe by adding `// colony-lint: safe-exec-concat` to the
 # same line or the preceding line.
 #
+# Known limitations (false positives / false negatives) are documented
+# in tools/check-exec-sh.md. Read that file before adding a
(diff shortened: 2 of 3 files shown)


ylohnitram:
## QA Report + PR Review -- PR #73 (`feat/colony-lint-hardening`)

### Summary (after `22a310c` follow-up addressing round-1 findings)

| Check | Result |
|---|---|
| `./tools/colony-lint.sh` baseline | **37 passed, 0 failed, 5 skipped** (+5 from pre-PR 32) |
| Stale `--backend` injected into daemon call | FAIL detected: "unknown daemon flag: --backend" |
| `--tick-interval=60000` (=value form) | PASS (correctly strips =value) |
| `-X` (unknown short flag) | FAIL detected: "unknown daemon flag: -X" |
| Echo guard: `echo "agentis daemon --backend"` | PASS (echo lines correctly skipped) |
| Allowlist vs core `DAEMON_FLAGS` parity (v1.1.5) | Exact match (10 flags) |
| `check-exec-sh.md` covers all 4 cases from #67 | PASS |
| Header pointer in `check-exec-sh.sh` correct | PASS |
| `test-check-exec-sh.sh` unit tests (14/14) | PASS |
| No behavior change in existing `check-exec-sh.sh` | PASS |
| Comment wording: "case-statement allowlist" (not nonexistent variable name) | PASS (fixed in 22a310c) |

### Round-1 findings status

| # | Severity | Finding | Status |
| 1 | MEDIUM | Awk matched `agentis daemon` inside echo strings | Fixed in `22a310c` (added `/^[[:space:]]*echo/ { next }`) |
| 2 | LOW | Spurious "daemon flags OK" for scripts without daemon calls | Accepted (no current script triggers this, cosmetic) |
| 3 | LOW | Comment referenced nonexistent `DAEMON_FLAG_ALLOWLIST` variable | Fixed in `22a310c` (now says "case-statement allowlist") |

### Verdict

**APPROVE.** Guardrail works end-to-end, docs cover all 4 cases from #67, findings addressed.

PR #90, opened 2026-04-13
fix(dev-apprenticeship): install.sh writes runtime config, agents use $COLONY_DIR (#88)

### LINKED ISSUE #88: dev-apprenticeship: install.sh does not write tick_interval to .agentis/config -> watchdog kills all LLM-backed agents after 2s
## Problem

After a fresh install of `dev-apprenticeship` (`./install.sh`, confidence 0.5, LLM backend = Claude CLI) and `./start-federation.sh`, all 21 agents start emitting:

```
watchdog [<id>] heartbeat stale (3003ms > 2000ms) — killing hung child
```

The watchdog uses a threshold of **2000 ms**, but Claude CLI typically takes 5–30 s to respond. Result: every agent gets killed during its first tick and HEALTH never recovers to `healthy`.

## Root cause

`src/daemon/watchdog.rs:27`
```rust
heartbeat_timeout_ms: cfg.get_u64("daemon.heartbeat_interval_ms", tick_interval * 2),
```
where `tick_interval = cfg.get_u64("daemon.tick_interval_ms", 1000)`.

Default path:
- `daemon.tick_interval_ms` missing in `.agentis/config` → fallback `1000`
- `daemon.heartbeat_interval_ms` missing in `.agentis/config` → fallback `tick_interval * 2 = 2000`

But `start-colony.sh` launches each agent with `agentis daemon ... --tick-interval 60000`. The CLI flag is not propagated into `WatchdogConfig` (it only reads from the TOML key), so the watchdog still thinks tick = 1 s and heartbeat threshold = 2 s.

## Workaround (which I applied)

Add to `dev-apprenticeship/.agentis/config`:
```
daemon.heartbeat_interval_ms = 180000
```
(180 s = 3× tick interval, enough for Claude CLI cold start)

After this change, all 21 agents survive and HEALTH stabilizes.

## Proposed fix

One of these:

1. **install.sh** should write to `.agentis/config` during setup:
   ```
   daemon.tick_interval_ms = 60000
   daemon.heartbeat_interval_ms = 180000
   ```
   (matches `--tick-interval 60000` in `start-colony.sh`)

2. **agentis-core**: `WatchdogConfig::from_config` should prefer the CLI flag `--tick-interval` over the TOML default. But that is not a colonies-repo concern.

I recommend #1 — i.e. update `dev-apprenticeship/install.sh` (section `--- 4. Initialize agentis ---`).

## Second problem (split into a separate issue?): cwd-local `.agentis` divergence

If you run `agentis doctor` from a colony directory (e.g. during debugging) and no `.agentis` exists in cwd, agentis creates a **new empty** `.agentis` in the current folder instead of walking up to find the existing one. After that, every `agentis daemon` process from that folder uses this empty `.agentis` → loads default config (`llm.backend = mock`, heartbeat 2000 ms) and behaves differently from the other colonies.

Workaround: pre-create `<colony>/.agentis -> ../.agentis` symlink in each colony. I would just bake this into `install.sh`.

## Environment

- macOS Darwin 25.3.0
- agentis v1.1.5
- LLM backend: Claude CLI (`-p --output-format text`)
- 5 colonies / 21 agents, all confidence 0.5

ylohnitram:
Closes #88.

## Summary

- **install.sh**: write seven runtime config keys that every colony in this federation needs but that `agentis init` does not set — `daemon.tick_interval_ms`, `daemon.heartbeat_interval_ms`, `daemon.cb_per_tick`, `experience.enabled`, `exec.env_passthrough`, plus the existing `federation.enabled` / `knowledge.federation_enabled`. All through a single idempotent helper. Without these, a fresh checkout on a real LLM backend (Claude CLI) cannot complete a single tick (see the #88 debug transcript).
- **install.sh**: create per-colony `.agentis` symlinks so `agentis doctor` or any cwd-local agentis command run from inside a colony dir doesn't silently spawn a divergent empty `.agentis` in that folder.
- **All 21 .ag files**: replace `./scripts/gitlab-api.sh ...` with `$COLONY_DIR/scripts/gitlab-api.sh ...`. `agentis daemon` runs `exec sh` with cwd set to `.agentis/sandbox/`, so the old relative paths never resolved.
- **All 21 .ag files**: rewrite `learn()` outcomes. The old `"observed"` / `"acted"` / `"emitted"` are not valid agentis outcomes (only `success/failure/partial/timeout/error` are accepted), so every `learn()` call raised `invalid outcome` and aborted the tick before `memo_write`. New form: `"success"` outcome + tag list `["observed"/"acted"/"emitted", colony]` preserves the semantic distinction. Also fixes a latent adjacent bug — `learn()` wants `Vec<String>` for tags and silently dropped the single string the .ag files were passing before.

## What's already done elsewhere

Bugs 8 (`messaging = allow`) and 9 (`exec_foreign = allow`) from the #88 comment were resolved by the agentis >= v1.1.6 bump in #89 via the new `--enable-exec` / `--enable-messaging` daemon flags.

## Test plan

- [x] `cargo build --release` of agentis-core, then `agentis commit` on each of the 21 rewritten `.ag` files — all parse
- [x] `./tools/colony-lint.sh` — 45 passed, 0 failed, 5 skipped (baseline)
- [x] `bash -n dev-apprenticeship/install.sh` — OK
- [x] Dry-run of the `write_key` helper: all seven keys written on pass 1, all recognized as "already configured" on pass 2 (idempotent)
- [x] CI green

### DIFF
--- a/dev-apprenticeship/README.md
+++ b/dev-apprenticeship/README.md
@@ -57,6 +57,12 @@ The install script checks prerequisites, creates configs for all 5 colonies, wri
 agentis daemon stop --all       # Stop everything
 ```
 
+> Agents must be launched via `start-federation.sh` or a colony's
+> `start-colony.sh`. Those scripts export `COLONY_DIR`, which the
+> agents expand inside `exec sh "$COLONY_DIR/scripts/..."` at runtime.
+> Launching `agentis daemon <agent>.ag` directly bypasses that export,
+> so `$COLONY_DIR` expands to empty and GitLab polling fails silently.
+
 ## Monitoring
 
 ### Dashboard
--- a/dev-apprenticeship/code-review/agents/approval_decider.ag
+++ b/dev-apprenticeship/code-review/agents/approval_decider.ag
@@ -44,7 +44,7 @@ fn tick(reason: string) -> void {
     };
 
     // Learn from past human approval/rejection patterns
-    let mr_cmd = "./scripts/gitlab-api.sh merge-requests --state merged";
+    let mr_cmd = "$COLONY_DIR/scripts/gitlab-api.sh merge-requests --state merged";
     let recent_merged = try { exec sh mr_cmd; } catch e { "[]"; };
 
     let approval_patterns = prompt(
@@ -53,7 +53,7 @@ fn tick(reason: string) -> void {
     ) -> string;
 
     if len(approval_patterns) > 10 {
-        learn("approval_decision", "merged MRs batch", approval_patterns, "observed", "code-review");
+        learn("approval_decision", "merged MRs batch", approval_patterns, "success", ["observed", "code-review"]);
     };
 
     // Aggregate findings and decide
@@ -71,20 +71,20 @@ fn tick(reason: string) -> void {
 
     if confidence >= 0.85 {
         if decision.action == "approve" {
-            let approve_cmd = "./scripts/gitlab-api.sh approve " + to_string(decision.mr_iid);
+            let approve_cmd = "$COLONY_DIR/scripts/gitlab-api.sh approve " + to_string(decision.mr_iid);
             try { exec sh approve_cmd; } catch e {
                 print("[approval_decider] Failed to approve:", e);
             };
             print("[approval_decider] Approved MR", decision.mr_iid);
-            learn("approval_decision", "MR " + to_string(decision.mr_iid), "approved: " + decision.reasoning, "acted", "code-review");
+            learn("approval_decision", "MR " + to_string(decision.mr_iid), "approved: " + decision.reasoning, "success", ["acted", "code-review"]);
         } else {
             if decision.action == "request_changes" {
                 let comment = "**Review Summary** (automated)\n\n" + decision.reasoning + "\n\nFindings: " + to_string(decision.total_findings) + " total, " + to_string(decision.critical_count) + " critical";
-                let post_cmd = "./scripts/gitlab-api.sh post-note " + to_string(decision.mr_iid) + " --body " + shell_escape(comment);
+                let post_cmd = "$COLONY_DIR/scripts/gitlab-api.sh post-note " + to_string(decision.mr_iid) + " --body " + shell_escape(comment);
                 try { exec sh post_cmd; } catch e {
                     print("[approval_decider] Failed to post:", e);
                 };
-                learn("approval_decision", "MR " + to_string(decision.mr_iid), "request_changes: " + decision.reasoning, "acted", "code-review");
+                learn("approval_decision", "MR " + to_string(decision.mr_iid), "request_changes: " + decision.reasoning, "success", ["acted", "code-review"]);
             } else {
                 print("[approval_decider] Escalating MR", decision.mr_iid, "to human");
                 emit("review:escalation", decision);
--- a/dev-apprenticeship/code-review/agents/logic_reviewer.ag
+++ b/dev-apprenticeship/code-review/agents/logic_reviewer.ag
@@ -24,9 +24,9 @@ type LogicReview {
 fn mr_cmd() -> string {
     let ts = recall_latest("logic_reviewer:last_check");
     if len(ts) > 0 {
-        return "./scripts/gitlab-api.sh merge-requests --since " + shell_escape(ts);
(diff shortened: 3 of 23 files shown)


ylohnitram:
## QA + review follow-up (commits 11d7158, f328852)

### PR reviewer findings

**Medium**
- [x] **#1 — emitted branches: outcome "success" → "partial"**. All 13 `"emitted"` tag entries across 9 agents rewritten to outcome `"partial"`. `"acted"` (16) and `"observed"` (20) remain `"success"` as intended. Verified count: 0 remaining `"success" + emitted`, 13 `"partial" + emitted`.
- [x] **#2 — `daemon.tick_interval_ms` comment oversold the watchdog scenario**. Rephrased to make clear start-colony.sh always overrides via `--tick-interval` and the config key is defense-in-depth for direct daemon launches.

**Low**
- [x] **#3 — `if [ -f "$AGENTIS_CONFIG" ]` silently skips key-writing**. Now `touch $AGENTIS_CONFIG` first so the helper runs even if `agentis init` was interrupted.
- [ ] **#4 — grep pattern prefix collision**. Currently safe for the seven chosen keys (`daemon.tick_interval_ms` vs a hypothetical `daemon.tick_interval` would not collide because `[[:space:]]*=` requires a space-or-equals boundary and `_ms` is neither). No change; noted as a constraint for future keys.
- [x] **#5 — `\s` portability on BSD grep (macOS)**. Switched to POSIX `[[:space:]]*`. Verified the tab-padded line case still resolves to "already configured".
- [x] **#6 — dangling symlink not detected on re-run**. New branch `[ -L path ] && [ ! -e path ]` repairs dangling symlinks via `ln -sfn`. Verified at `/tmp/sl-test` with move-target-then-reinstall path.

**Nitpicks**
- [ ] **#7 — `AGENTIS_DIR` may be relative**. Not acted on; reviewer noted "this is fine" — the symlink block absolutizes via `$(cd "$AGENTIS_DIR" && pwd)` anyway.
- [x] **#8 — document that agents must be launched via start-colony.sh (so `COLONY_DIR` is exported)**. Added a one-paragraph callout in `dev-apprenticeship/README.md` in the Starting section.

### QA agent findings

- [x] **Finding 1 (INFO, GNU-only `\s`)**: same as review #5, fixed.
- [x] **Finding 2 (INFO, dangling symlink)**: same as review #6, fixed.
- [ ] **Finding 3 (INFO, pre-existing cwd-sensitive `agentis init` fallback)**: not introduced by this PR, out of scope. Filed for a future hardening pass.

### Verification after fixes

- `bash -n dev-apprenticeship/install.sh`: OK
- `./tools/colony-lint.sh`: 45 passed, 0 failed, 5 skipped (shellcheck only)
- `agentis commit` on all 21 `.ag` files: 0 failures
- `write_key` idempotency simulated twice against a tab-padded config: all keys recognized as "already configured" on pass 2
- Dangling-symlink repair path tested end-to-end: DETECTED + re-pointed correctly

ylohnitram:
## Follow-up: remaining unchecked items now resolved (commit 483de61)

Resolving the last three items so the checklist is complete.

- [x] **Review #4 — grep prefix-collision safety**. Extended the `write_key` helper comment to document the whitespace-or-equals boundary guarantee — keys that are proper prefixes of other keys (e.g. `daemon.tick_interval` vs `daemon.tick_interval_ms`) do not collide because the `_` in `_ms` is neither whitespace nor `=`. The one invalidating scenario (a key whose name ends in `[[:space:]=]`) is already invalid in agentis config format, so the guarantee holds for any future legal key.
- [x] **Review #7 — `AGENTIS_DIR` relative path**. Now resolved absolutely via `FED_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"` once at the top of the init section; `AGENTIS_DIR` and the symlink targets all use that absolute path uniformly.
- [x] **QA #3 — cwd-sensitive `agentis init` fallback**. `agentis init` now runs from `$FED_ROOT` explicitly: `(cd "$FED_ROOT" && agentis init ...)`. End-to-end test with a synthetic federation at `/tmp/fed88-v3/` confirmed `.agentis/` lands at the federation root regardless of the invoker's cwd.

### Verification

- `bash -n install.sh`: OK
- `./tools/colony-lint.sh`: 45 passed, 0 failed, 5 skipped
- Synthetic cwd-independence test: `agentis init` landed correctly at federation root, no stray `.agentis/` in unrelated cwd.

PR #94, opened 2026-04-13
fix(dashboard): Kill Federation reports real count + cwd=FED_DIR (#91)

### LINKED ISSUE #91: Dashboard Kill Federation button silently no-ops — wrong cwd for agentis daemon stop --all
## Problem

Clicking the \"Kill Federation\" button in the web dashboard (`tools/federation-dashboard.sh`, port 8420) — even clicking twice to confirm — does **nothing**. The federation's 21 (or in my case 5 surviving) daemons keep running. No error is shown. The button text flips to \"Federation Stopped\" as if it worked.

## Repro

```
./start-federation.sh                 # start agents
./dashboard.sh                        # start dashboard
# in browser: http://localhost:8420, click Kill Federation, click again to confirm
# verify:
ps -eo command | grep '[d]aemon-inner' | wc -l     # still > 0
```

## Root cause

The dashboard's Python HTTP handler (`federation-dashboard.sh:636`) does:

```python
result = subprocess.run(
    ['agentis', 'daemon', 'stop', '--all'],
    capture_output=True, text=True
)
```

with **no `cwd=` argument**. Earlier in the script at line 628:

```python
os.chdir(serve_dir)   # serve_dir = $FED_DIR/.dashboard
```

So by the time the subprocess runs, the process cwd is `dev-apprenticeship/.dashboard/`.

`agentis daemon stop --all` then computes `agentis_root() = Path::new(\".agentis\")` (a literal relative path, no walk-up). That resolves to `.dashboard/.agentis/` which does not exist. `agentis daemon list` from that cwd returns \"No running daemons.\" The stop command silently does nothing.

Meanwhile the actual daemon PIDs were registered under `<repo>/.agentis/daemon/` because start-colony.sh cd's each agent into a colony dir whose `.agentis` symlinks back to the repo root.

## Reproduction of the bug in isolation

```
$ cd dev-apprenticeship/.dashboard
$ agentis doctor
[!!] .agentis/ not found (run 'agentis init')

$ agentis daemon list
No running daemons.

# ... but meanwhile:
$ ps -eo command | grep '[d]aemon-inner' | wc -l
5
```

## Fix sketch

Two options (easiest first):

1. **Dashboard-side fix (3 lines):** add `cwd=` to the subprocess call, pointing at the federation directory:
   ```python
   result = subprocess.run(
       ['agentis', 'daemon', 'stop', '--all'],
       capture_output=True, text=True,
       cwd=os.path.dirname(serve_dir),   # FED_DIR, parent of .dashboard
   )
   ```
   or `cd \"$FED_DIR\" &&` in the shell wrapper.

2. **Upstream core fix:** make `agentis_root()` in core walk up from cwd to find the nearest `.agentis`, like git does for `.git`. This would fix the whole class of \"agentis CLI invoked from subdir finds wrong/no .agentis\" problems (this is the same root cause as the \"agentis doctor auto-creates spurious `.agentis`\" bug in Replikanti/agentis-colonies#88, and the symlink workaround for colony directories).

Option 2 is the real fix. Option 1 is a 5-minute band-aid.

## Side effect of this bug

The button confidently flips to \"Federation Stopped\" even though nothing stopped. Operator assumes it worked. Daemons keep burning CB and sending GitLab polls. I noticed only because I ran `ps` manually after clicking.

Suggest the handler should check the return value from `agentis daemon stop --all` and report the count of actually-stopped daemons in the button text or a status line.

## Environment

- agentis v1.1.6
- macOS Darwin 25.3.0
- `dev-apprenticeship` federation

ylohnitram:
## Summary

- Pass `cwd=FED_DIR` to the `agentis daemon stop --all` subprocess (was running from `.dashboard/`, which has no `.agentis/` of its own).
- Parse stderr (`stop signal sent: <id>` lines) to count actual stopped daemons; return `N daemon(s) stopped` as the first response line.
- JS now shows the real count. When `N == 0`, the meta-refresh is kept in place and the button re-arms so the operator can retry.
- Subprocess gets a 15s timeout, and exec errors surface as HTTP 500 instead of tracebacks.

## Why

Per #91, clicking "Kill Federation" silently no-op'd while the button flipped to "Federation Stopped". Root cause: `agentis daemon stop --all` was invoked from `.dashboard/` cwd, so `agentis_root()` in agentis v1.1.6 resolved to `.dashboard/.agentis/` (nonexistent) instead of the federation's `.agentis/`.

This class of bug is also addressed upstream in agentis v1.1.7 (walk-up in `agentis_root()`, #499). Passing `cwd=FED_DIR` here is defense-in-depth: the dashboard stays honest on older agentis builds, and the button surfaces the real outcome either way.

## Test plan

- [x] Shell syntax: `bash -n tools/federation-dashboard.sh`
- [x] Embedded Python block compiles cleanly
- [x] Manual verification: `agentis daemon stop --all` from a `.dashboard/` subdir under an initialized `.agentis/` correctly finds and reports the daemon state on v1.1.7 (walk-up resolves the ancestor repo).
- [x] End-to-end browser test: start federation, click Kill, observe "N daemon(s) stopped" text (requires running daemons — deferred; logic is straightforward string parsing and the walk-up is validated by the agentis-core test suite).

Closes #91

### DIFF
--- a/tools/federation-dashboard.sh
+++ b/tools/federation-dashboard.sh
@@ -654,10 +654,12 @@ fed_dir = os.path.dirname(serve_dir)
 class Handler(SimpleHTTPRequestHandler):
     def do_POST(self):
         if self.path == '/kill':
-            # Run from fed_dir, not .dashboard: even with the v1.1.7 walk-up
-            # in agentis_root(), keeping cwd explicit means this works on
-            # older agentis builds too and avoids surprise if .dashboard
-            # ever moves to a location outside the federation tree.
+            # Run from fed_dir, not .dashboard. The actual fix for #91 is
+            # the v1.1.7 walk-up in agentis_root() (Replikanti/agentis-core#499)
+            # which finds .agentis/ in any ancestor; passing cwd=fed_dir
+            # just makes the intent explicit and stops .dashboard/ from
+            # appearing as a plausible agentis root in future agentis
+            # versions that might use some other marker than .agentis/objects.
             try:
                 result = subprocess.run(
                     ['agentis', 'daemon', 'stop', '--all'],
--- a/tools/federation-dashboard.sh
+++ b/tools/federation-dashboard.sh
@@ -590,8 +590,28 @@ function killSwitch() {
   btn.textContent = 'Killing...';
   btn.className = 'kill-btn killed';
   fetch('/kill', { method: 'POST' })
-    .then(r => r.text())
-    .then(() => { btn.textContent = 'Federation Stopped'; const m = document.querySelector('meta[http-equiv="refresh"]'); if (m) m.remove(); })
+    .then(r => r.text().then(t => ({ok: r.ok, text: t})))
+    .then(({ok, text}) => {
+      if (!ok) {
+        // Server returned non-2xx: surface the message and allow retry.
+        btn.textContent = 'Kill Failed: ' + text.split('\n')[0].slice(0,60);
+        btn.className = 'kill-btn';
+        killArmed = false;
+        return;
+      }
+      // First line of reply is the count, e.g. "5 daemon(s) stopped".
+      // "0 daemon(s) stopped" means we clicked but there was nothing to kill
+      // — leave the meta-refresh in place and let the operator retry.
+      const first = text.split('\n')[0] || 'Federation Stopped';
+      btn.textContent = first;
+      if (!/^0 /.test(first)) {
+        const m = document.querySelector('meta[http-equiv="refresh"]');
+        if (m) m.remove();
+      } else {
+        btn.className = 'kill-btn';
+        killArmed = false;
+      }
+    })
     .catch(() => { btn.textContent = 'Kill Failed'; btn.className = 'kill-btn'; killArmed = false; });
 }
 </script>
@@ -625,20 +645,50 @@ import sys, os, subprocess
 from http.server import HTTPServer, SimpleHTTPRequestHandler
 
 serve_dir, port = sys.argv[1], int(sys.argv[2])
+# serve_dir is $FED_DIR/.dashboard; the federation root (parent of serve_dir)
+# is the cwd we want agentis to resolve .agentis/ from. We still os.chdir to
+# serve_dir so SimpleHTTPRequestHandler serves index.html from here.
 os.chdir(serve_dir)
+fed_dir = os.path.dirname(serve_dir)
 
 class Handler(SimpleHTTPRequestHandler):
     def do_POST(self):
         if self.path == '/kill':
+            # Run from fed_dir, not .dashboard: even with the v1.1.7 walk-up
+            # in agentis_root(), keeping cwd explicit means this works on
+            # older agentis builds too and avoids surprise if .dashboard
+            # ever moves to a location outside the federation tree.
+            try:
+                result = subprocess.run(
+                    ['agentis', 'daemon', 'stop', '--all'],
+                    capture_output=True, text=True,
+                    cwd=fed_dir,
+                    timeout=15,
+                )
+            except (OSError, subprocess.SubprocessError) as e:
+                self.send_response(500)
+                self.send_header('Content-Type', 'text/plain')
+                self.end_headers()
+                self.wfile.write(f'exec failed: {e}'.encode())
+                return
+            # `agentis daemon stop --all` emits to stderr:
+            #   "No running daemons to stop." or one "stop signal sent: <id>"
+            #   line per daemon. Count them so the button reports real state.
+            err = (result.stderr or '').strip()
+            lines = [l for l in err.splitlines() if l]
+            stopped = [l for l in lines if l.startswith('stop signal sent:')]
+            if result.returncode != 0:
+                self.send_response(500)
+                self.send_header('Content-Type', 'text/plain')
+                self.end_headers()
+                self.wfile.write((err or 'agentis daemon stop failed').encode())
+                return
             self.send_response(200)
             self.send_header('Content-Type', 'text/plain')
             self.end_headers()
-            result = subprocess.run(
-                ['agentis', 'daemon', 'stop', '--all'],
-                capture_output=True, text=True
-            )
-            msg = result.stdout.strip() or result.stderr.strip() or 'Federation stopped'
-            self.wfile.write(msg.encode())
+            summary = f'{len(stopped)} daemon(s) stopped'
+            body = summary + ('\n' + err if err else '')
+            self.wfile.write(body.encode())
         else:
             self.send_error(404)
     def log_message(self, format, *args):


ylohnitram:
## PR #94 Review

### Verdict: APPROVE

Focused one-file fix that correctly addresses the silent-no-op. I verified the key assumptions against agentis-core and the rest of `federation-dashboard.sh`.

### Correctness checks (all pass)

- **stderr parsing**: confirmed against `agentis-core/src/cli/daemon.rs:1520-1534`. `cmd_daemon_stop --all` emits exclusively via `eprintln!` — `"stop signal sent: <id>"` per daemon, or `"No running daemons to stop."` when empty. No `println!` in the stop path, no locale/i18n layer, plain ASCII. stdout stays clean. Parsing is robust.
- **returncode**: `--all` unconditionally returns `Ok(())` (daemon.rs:1529), so a non-zero exit would only indicate a genuine error (`stop_all_daemons` silently drops non-alive daemons; it doesn't return Result). The `result.returncode != 0` guard is correct and will only fire on real failures (argv parse error, panic, etc.).
- **cwd resolution**: this works, but only because of the v1.1.7 walk-up, and worth being explicit about why. `fed_dir = os.path.dirname(serve_dir)` = `$FED_DIR`. But `.agentis/` actually lives at `${FED_DIR}/../.agentis/` (see line 136 of this same file: `local LOG_DIR=\"${FED_DIR}/../.agentis/logs\"`). So cwd=`$FED_DIR` puts agentis *inside* a subdirectory of the real repo root. On pre-v1.1.7 agentis (no ancestor walk-up) this would still silently no-op the same way the old `.dashboard/` cwd did. The PR description acknowledges this and #93 bumps the minimum to v1.1.7, so it's fine — but the inline comment at lines 657-660 overstates the robustness (\"works on older agentis builds too\"). It does *not* work on older agentis unless the user happens to start the dashboard from `REPO_ROOT` with a federation directly at `REPO_ROOT/.agentis/`. Consider softening that comment to match the v1.1.7 minimum.
- **GET handler regression**: no regression. `os.chdir(serve_dir)` happens once at startup and the handler only overrides cwd via `subprocess.run(cwd=fed_dir)`, which doesn't change the Python process cwd. `SimpleHTTPRequestHandler.do_GET` continues to resolve against the Python process cwd (= `serve_dir`). index.html still served correctly.
- **JS empty-body**: `text.split('\\n')[0] || 'Federation Stopped'` handles empty string fine (empty string is falsy). Regex `/^0 /` doesn't match empty, so meta-refresh is removed in the empty case — which is a graceful degradation to prior behavior. Good.
- **Timeout handling**: `subprocess.TimeoutExpired` is a `SubprocessError` subclass, so the except catches it. 15s is generous for a local stop-file write loop. Good.

### Suggestions (nice-to-have, non-blocking)

| 1 | tools/federation-dashboard.sh | 657-660 | Comment claims this also works on older agentis builds. It doesn't: `.agentis/` lives at `${FED_DIR}/../.agentis/` (see line 136), so the walk-up is load-bearing. Suggest rewording to \"relies on v1.1.7 ancestor walk-up in `agentis_root()`\" to match the PR description and #93. |
| 2 | tools/federation-dashboard.sh | 656 (handler scope) | Pre-existing but now-relevant: the `/kill` POST has no CSRF token or Origin/Referer check. Binding to 127.0.0.1 limits the blast radius, but a malicious page the operator visits in another tab can still `fetch('http://127.0.0.1:8420/kill', {method:'POST'})` and the two-stage confirm is client-side only. Pre-existing issue; PR doesn't worsen it, but now that the button actually kills things it's worth a follow-up issue. Cheap mitigation: require `Origin: http://127.0.0.1:<port>` or a one-time token baked into the served HTML. |
| 3 | tools/federation-dashboard.sh | 679 | `stopped = [l for l in lines if l.startswith('stop signal sent:')]` is stable against agentis adding other stderr lines (e.g. warnings), but if agentis-core ever changes the exact prefix this silently counts zero. Low probability given the prefix is tested in agentis-core, but worth noting; no action needed unless you want to cross-link to the agentis-core test that pins the format. |

### Positive notes

- Two-stage kill + re-arm on `0 daemon(s)` is a genuinely good UX improvement; operator can now tell a dead button apart from \"nothing to kill\".
- Error path returns HTTP 500 with body instead of a traceback — makes the JS `Kill Failed: <reason>` branch usable for debugging.
- 15s timeout prevents the dashboard from hanging on a wedged `agentis` invocation.
- `body = summary + ('\\n' + err if err else '')` preserves the raw stderr as additional lines for operators who want to see which IDs were stopped — nice touch.

ylohnitram:
## QA Report — PR #94 (fix/issue-91-kill-switch-cwd)

Focused review of `tools/federation-dashboard.sh` diff against `main`. End-to-end not executed (requires live daemons + browser, per request).

### Summary
- **Bash syntax** — NOT run in this QA sandbox (`bash -n` is disallowed). Verified by reading: heredoc markers `HEADEOF`/`HTMLEOF`/`JSEOF`/`PY`/`PYSERVER` all close; modified region stays inside a single-quoted heredoc and inside `do_POST`. Recommend running `bash -n tools/federation-dashboard.sh` locally before merge.
- **PYSERVER heredoc** — Extracted and `python3 -m py_compile` passed (Python 3.14.3).
- **Walk-up reasoning** — Confirmed. `.agentis/` is at `agentis-colonies/` (parent of `FED_DIR=agentis-colonies/dev-apprenticeship`). v1.1.7 `agentis_root()` walks `cwd.ancestors()` for `.agentis/objects/`, so `cwd=fed_dir` finds it one level up. Follow-up commit `60d94b9` correctly corrects the earlier comment: pre-v1.1.7 would *not* have found it from `cwd=fed_dir` either, so the walk-up is the real fix and `cwd=fed_dir` is a locality hint, not a backport.
- **Other `subprocess.run` call sites** — `grep subprocess\. tools/federation-dashboard.sh` shows exactly one `subprocess.run` (the `/kill` handler) and one `subprocess.SubprocessError` reference. No other sites share the cwd bug. The `generate()` shell function calls `agentis ...` directly as external commands — they inherit bash's cwd (the terminal where the user started the dashboard), which is independent of the Python server and is the user's problem to choose.

### Findings

| 1 | LOW | cli/ux | Relative `FED_DIR` (pre-existing, not introduced by this PR) breaks subprocess `cwd=fed_dir` resolution after `os.chdir(serve_dir)` | `./tools/federation-dashboard.sh ../some/rel/path` from outside repo root; `FED_DIR` stays relative, passed as relative `serve_dir` → `os.chdir` resolves it fine but later `cwd=fed_dir` is a bare dirname that Python resolves relative to the now-changed cwd. Walk-up still saves the day in practice. | - [ ] |
| 2 | LOW | ux | On successful kill (`N>0`), the button's `className` stays `kill-btn killed` and `killArmed` is not reset. Intentional (federation is stopped, no retry needed), but if the operator manually refreshes and daemons restart, they cannot re-arm without a full page reload. | Click Kill → confirm → wait; click button again. | - [ ] |
| 3 | LOW | ux | `Kill Failed: ` prefix + `.slice(0,60)` can truncate mid-word (e.g. `exec failed: [Errno 2] No such file or directory: 'age`). Harmless, but consider ellipsis. | Rename/remove `agentis` from `PATH` and click Kill. | - [ ] |
| 4 | LOW | robustness | `timeout=15` for `daemon stop --all` is a soft guess. For a federation with many daemons this is almost certainly plenty (each stop is a file write), but worth confirming on a large deployment. On timeout the client sees 500 + `Kill Failed: exec failed: ...` and can retry — failure mode is safe. | N/A. | - [ ] |
| 5 | LOW | consistency | Server comment says `'stop signal sent:'` is emitted to stderr — verified in `agentis-core/src/cli/daemon.rs:1526`. If that string is ever changed upstream, the count silently goes to zero and the button re-arms. Consider a comment in `daemon.rs` pointing back here, or a more tolerant match. Tracking via agentis-core version pin in `chore` commit fc42f3a partially mitigates. | N/A. | - [ ] |

### Not bugs (verified OK)

- `os.path.dirname(serve_dir)` when `serve_dir = "$FED_DIR/.dashboard"` → `$FED_DIR`. Correct.
- `stopped = [l for l in lines if l.startswith('stop signal sent:')]` — matches exactly one line per stopped daemon as emitted by `daemon.rs:1526` and `:1534`.
- `subprocess.TimeoutExpired` is a `SubprocessError`, so the timeout is caught by the `except`. Verified against Python docs.
- `os.chdir(serve_dir)` still happens, so `SimpleHTTPRequestHandler` continues to serve `index.html` from `.dashboard/` for GETs.
- Single-quoted `<<'PYSERVER'` — no shell interpolation of `$` inside the Python body, so `sys.argv`, `os.path.dirname`, f-strings etc. stay literal. Good.
- Client: empty server reply → `text.split('\n')[0] || 'Federation Stopped'` falls back. `/^0 /` test on `0 daemon(s) stopped` correctly re-arms.
- `check_known_flags(args, &["--all"], &[])` allows `--all` at `args[0]`, so `agentis daemon stop --all` never exits 2 by surprise.

### Verdict
No blocking bugs found in the diff. The fix is sound: the count parsing is faithful to `daemon.rs` emission, the cwd change is safe, and the client reliably re-arms on no-op kills. Merge when reviewer is satisfied and a local `bash -n` has been run.

PR #97, opened 2026-04-13
fix(dashboard): handle fresh federation without running daemons (#96)

### LINKED ISSUE #96: federation-dashboard.sh: generate() crashes on fresh federation with JSON parse error
## Problem

Running `./dashboard.sh` against a federation directory with no running daemons / no knowledge base / no memo values (i.e. a brand-new installation between `./install.sh` and the first `./start-federation.sh`) crashes with an uncaught `json.JSONDecodeError` inside `generate()`:

```
Traceback (most recent call last):
  File "<stdin>", line 4, in <module>
  File "/usr/lib64/python3.14/json/__init__.py", line 352, in loads
    return _default_decoder.decode(s)
  File "/usr/lib64/python3.14/json/decoder.py", line 345, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib64/python3.14/json/decoder.py", line 363, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 2 (char 1)
```

No `index.html` is ever written, so the server starts but 404s on `/`.

## Repro

```bash
cd agentis-colonies
agentis init              # fresh .agentis/
rm -rf dev-apprenticeship/.dashboard
bash tools/federation-dashboard.sh dev-apprenticeship 18422
# crashes immediately with the traceback above
```

## Root cause

`tools/federation-dashboard.sh` lines 150-174 launch a python heredoc that starts with:

```python
kc = json.loads(sys.argv[3])          # $KNOWLEDGE_COUNTS
conf_vals = json.loads(sys.argv[4])   # $CONFIDENCES
agent_map = json.loads(sys.argv[5])   # $AGENT_COLONY_MAP
```

On a fresh federation, `agentis memo get <agent>:confidence` returns empty string (no memo written yet). The shell then builds `CONFIDENCES="[,,,,...,]"` — all empty values between commas — which is not valid JSON.

Separate but related: only the history-file read at line 158 is wrapped in try/except; the three `json.loads(sys.argv[...])` calls on 153-155 have no guard, so any malformed argv string crashes the whole dashboard instead of falling back.

## Fix sketch

Two independent fixes, either sufficient:

1. **Build valid JSON in shell** — when `agentis memo get` returns empty, use `0.0` instead of `""`:
   ```sh
   conf="$(agentis memo get "${agent}:confidence" 2>/dev/null)"
   conf="${conf:-0.0}"
   ```

2. **Guard the python argv reads** — wrap in try/except returning sane defaults (`{}`, `[]`, `[]`), matching the treatment of `history` on line 156-160.

Option 1 addresses the root cause. Option 2 makes the script resilient to future empty inputs.

## Impact

New operators following the README (`./install.sh` → `./dashboard.sh` before `./start-federation.sh`) see an immediately broken dashboard. After starting the federation once, the memos get populated and the bug stops reproducing — which is why it hasn't surfaced before.

Also blocks end-to-end browser verification on QA runs that don't have a live federation (flagged in the test plan of #95).

## Environment

- agentis v1.1.7
- bash 5.x
- python 3.14 (though any Python raises the same error)

ylohnitram:
## Summary

Two compounding bugs crashed `dashboard.sh` on a freshly installed federation before the first `start-federation.sh`:

1. **Shell-side root cause**: `agentis memo get <agent>:confidence` exits 0 with empty stdout on a missing key, so the `|| echo '0.0'` fallback never fires. The shell assembled `CONFIDENCES="[,,,,...]"` — valid array brackets around comma-separated empty strings — which is not valid JSON. Fix: use `${conf:-0.0}` parameter expansion which catches both unset and empty.
2. **Python-side defense**: `json.loads(sys.argv[3..5])` in the history-append heredoc had no try/except, while only the history file read was guarded. Introduce a small `_safe_json(s, default)` helper so a malformed argv degrades to empty structures instead of crashing `generate()`. The history entry ends up with empty `knowledge` / `confidence` fields for that tick, but `index.html` renders.

## E2E verification (the unchecked box from PR #95 is finally ticked)

```bash
mkdir -p /tmp/smoke/fed/colony1/agents /tmp/smoke/fed/colony2/agents
touch /tmp/smoke/fed/{colony1,colony2}/agents/agent_{a,b}.ag
cd /tmp/smoke && agentis init
timeout 6 bash tools/federation-dashboard.sh fed 18424
# Before this PR: crashes with JSONDecodeError, no index.html
# After this PR: index.html written, history.json written, HTTP server starts
```

Subsequent runs append valid entries to `history.json`:
```json
[
    {"t": 1776117534, "knowledge": {"colony1": 0, "colony2": 0, "total": 0},
     "confidence": {"colony1": 0.0, "colony2": 0.0}},
    {"t": 1776117644, ...}
]
```

## Test plan

- [x] `bash -n tools/federation-dashboard.sh` passes
- [x] Both embedded Python heredocs (history-append, PYSERVER) still `compile()` cleanly
- [x] E2E: fresh `agentis init`'d federation → `index.html` + `history.json` generated on first run
- [x] E2E: second dashboard run appends a well-formed entry to `history.json` (verified via `python3 -m json.tool`)
- [x] Countdown / refresh-btn markers from PR #95 still present in the rendered HTML

## Out of scope

- `agentis remediation history --json` returns `{"history":[],...}` while the shell fallback emits `[]` (mismatched shape). Separate display bug; not triggered by this PR's scenario.
- Rewriting shell-driven JSON assembly into a single Python pass. Real long-term fix but 10× the diff.

Closes #96

### DIFF
--- a/tools/federation-dashboard.sh
+++ b/tools/federation-dashboard.sh
@@ -123,14 +123,16 @@ print(json.dumps(counts))
     REMEDIATION="$(agentis remediation history --limit 5 --json 2>/dev/null || echo '[]')"
 
     # Confidence values. `agentis memo get` on a missing key exits 0 with
-    # empty stdout (not non-zero), so `|| echo '0.0'` never fires — the
-    # shell substitution default on the next line is what actually catches
-    # a fresh federation where no memos have been written yet (#96).
+    # empty stdout (not non-zero), so `|| echo ''` is not what catches the
+    # missing-key case — the `${conf:-0.0}` default on the next line is
+    # (#96). But the `|| echo ''` is still needed to keep the script
+    # running under `set -e` (L17) if `agentis` itself exits non-zero
+    # (binary missing, store lock contention, etc.).
     local CONFIDENCES=""
     for i in "${!ALL_AGENTS[@]}"; do
         local agent="${ALL_AGENTS[$i]}"
         local conf
-        conf="$(agentis memo get "${agent}:confidence" 2>/dev/null)"
+        conf="$(agentis memo get "${agent}:confidence" 2>/dev/null || echo '')"
         CONFIDENCES="${CONFIDENCES}${conf:-0.0},"
     done
     CONFIDENCES="[${CONFIDENCES%,}]"
@@ -152,19 +154,25 @@ print(json.dumps(lines))
     # Append to history
     python3 - "$HISTORY_FILE" "$EPOCH" "$KNOWLEDGE_COUNTS" "$CONFIDENCES" "$AGENT_COLONY_MAP" <<'PY'
 import sys, json
-def _safe_json(s, default):
+def _safe_json(s, default, label):
     try:
         return json.loads(s)
-    except (json.JSONDecodeError, TypeError, ValueError):
+    except (json.JSONDecodeError, TypeError, ValueError) as e:
+        # Surface real breakages (daemon crash, store corruption) rather
+        # than silently rendering "0 entries" forever. On a fresh
+        # federation the inputs are legitimately empty strings and
+        # `json.loads("")` fails — that is expected and noisy but
+        # short-lived (one tick per agent until memos are seeded).
+        sys.stderr.write(f"[dashboard] {label} parse failed: {e}; using default\n")
         return default
 path, epoch = sys.argv[1], int(sys.argv[2])
 # Defensive parse: on a fresh federation any of these shell-assembled JSON
 # blobs may be malformed (see #96). Fall back to empty structures so the
 # dashboard still renders — the resulting history entry just has empty
 # `knowledge` / `confidence` fields for this tick.
-kc = _safe_json(sys.argv[3], {"total": 0})
-conf_vals = _safe_json(sys.argv[4], [])
-agent_map = _safe_json(sys.argv[5], [])
+kc = _safe_json(sys.argv[3], {"total": 0}, "knowledge_counts")
+conf_vals = _safe_json(sys.argv[4], [], "confidences")
+agent_map = _safe_json(sys.argv[5], [], "agent_colony_map")
 try:
     with open(path) as f:
         history = json.load(f)
--- a/tools/federation-dashboard.sh
+++ b/tools/federation-dashboard.sh
@@ -122,13 +122,16 @@ print(json.dumps(counts))
     local REMEDIATION
     REMEDIATION="$(agentis remediation history --limit 5 --json 2>/dev/null || echo '[]')"
 
-    # Confidence values
+    # Confidence values. `agentis memo get` on a missing key exits 0 with
+    # empty stdout (not non-zero), so `|| echo '0.0'` never fires — the
+    # shell substitution default on the next line is what actually catches
+    # a fresh federation where no memos have been written yet (#96).
     local CONFIDENCES=""
     for i in "${!ALL_AGENTS[@]}"; do
         local agent="${ALL_AGENTS[$i]}"
         local conf
-        conf="$(agentis memo get "${agent}:confidence" 2>/dev/null || echo '0.0')"
-        CONFIDENCES="${CONFIDENCES}${conf},"
+        conf="$(agentis memo get "${agent}:confidence" 2>/dev/null)"
+        CONFIDENCES="${CONFIDENCES}${conf:-0.0},"
     done
     CONFIDENCES="[${CONFIDENCES%,}]"
 
@@ -149,10 +152,19 @@ print(json.dumps(lines))
     # Append to history
     python3 - "$HISTORY_FILE" "$EPOCH" "$KNOWLEDGE_COUNTS" "$CONFIDENCES" "$AGENT_COLONY_MAP" <<'PY'
 import sys, json
+def _safe_json(s, default):
+    try:
+        return json.loads(s)
+    except (json.JSONDecodeError, TypeError, ValueError):
+        return default
 path, epoch = sys.argv[1], int(sys.argv[2])
-kc = json.loads(sys.argv[3])
-conf_vals = json.loads(sys.argv[4])
-agent_map = json.loads(sys.argv[5])
+# Defensive parse: on a fresh federation any of these shell-assembled JSON
+# blobs may be malformed (see #96). Fall back to empty structures so the
+# dashboard still renders — the resulting history entry just has empty
+# `knowledge` / `confidence` fields for this tick.
+kc = _safe_json(sys.argv[3], {"total": 0})
+conf_vals = _safe_json(sys.argv[4], [])
+agent_map = _safe_json(sys.argv[5], [])
 try:
     with open(path) as f:
         history = json.load(f)


ylohnitram:
## PR #97 Review

### Verdict: APPROVE with one minor caveat

The fix correctly diagnoses and repairs the two root causes in #96. Both the shell-side `${conf:-0.0}` substitution and the Python `_safe_json` helper land in the right place. I went through every sibling JSON-construction site you listed plus the downstream JS, and the analysis holds up.

### Correctness check on the shell fix

`${conf:-0.0}` handles both **unset** (the `local conf` pre-declaration without assignment if the `$()` had failed) and **empty string** (the actual bug: `memo get` on a missing key prints nothing). Source confirmation: `src/cli/memo.rs:99-102` — `Ok(None) => {}` falls through to exit 0 with no stdout. That matches what `_safe_json` says in its comment.

One edge case `${conf:-0.0}` does NOT cover: whitespace-only stdout (e.g., `conf=" "` → `CONFIDENCES="[ ,...]"` which is invalid JSON). This is theoretical — we control the writers and memo values are never written with leading spaces — so not worth a change. `${conf:='0.0'}` would be slightly worse here (same whitespace behavior, plus it pollutes `conf` itself which you don't need). Current choice is right.

### Sibling JSON-construction sites — all verified safe

| Site | Line | Verdict |
|------|------|---------|
| `DAEMONS` | 98 | `daemon list --json` always prints `[]` even on empty (`src/cli/daemon.rs:1430-1462` unconditionally `println!`s the array). Plus `\|\| echo '[]'` guard. Safe. |
| `KNOWLEDGE_COUNTS` | 108 | Inline python always `print(json.dumps(counts))`. Shell fallback `\|\| echo '{"total":0}'`. Safe. |
| `REMEDIATION` | 123 | Always prints `{"history":[],"limit":N}` (`cognitive.rs:446`). Shape mismatch vs `[]` fallback is real but JS at L624 defends with `Array.isArray()`. Out of scope, as you noted. |
| `SUGGESTIONS` | 145-149 | Inline python always emits `json.dumps(lines)`. Safe. |
| `HISTORY` | 190 | File is created at L46-48 before `generate()` runs. `cat \|\| echo '[]'` guard. Safe. |

No other hidden landmines in this scenario.

### Python `_safe_json` defaults are fine downstream

- `kc = {"total": 0}`: used only at line 181 (`entry = {"knowledge": kc, ...}`) then serialized into `history.json`. Readers at lines 598/605/606/616 all use `h.knowledge[col] \|\| 0` / `h.knowledge && ...`. No `KeyError` risk — `kc` is never dereferenced by colony name inside the heredoc.
- `conf_vals = []` paired with `agent_map = []` is internally consistent, and line 177 (`v = conf_vals[i] if i < len(conf_vals) else 0.0`) is defensive even against a length mismatch.
- The resulting entry has `{"knowledge": {"total": 0}, "confidence": {}}` for the degraded tick, which is the documented behavior.

| 1 | tools/federation-dashboard.sh | 133 | Regression vs. old behavior under `set -e` (L17): the old `\|\| echo '0.0'` also guarded against `agentis memo get` exiting non-zero (e.g., binary missing from PATH, invalid-key error from a future change, daemon-store lock error). The new line relies only on empty-stdout handling via `${conf:-0.0}`. For a truly degraded env the loop now aborts under `set -e` instead of producing a fallback row. Minor — the old guard was almost free. Consider `conf="$(agentis memo get "${agent}:confidence" 2>/dev/null)" \|\| conf=""` or keep the old `\|\| echo '0.0'` tail and layer `${conf:-0.0}` on top (belt + suspenders). |
| 2 | tools/federation-dashboard.sh | 155-159 | `_safe_json` silently swallows parse errors. If `agentis knowledge list` pipeline breaks in a real way, the operator sees "0 knowledge entries" forever with no signal. A one-line `sys.stderr.write(f"dashboard: argv[{i}] not JSON, falling back: {s!r:.80}\\n")` in the `except` branch would surface real failures without breaking the dashboard. Not blocking — this PR's goal is to stop the crash, not to add observability. |

### Positive notes

- Comment on L125-128 correctly identifies the non-obvious reason (`\|\| echo '0.0'` never fired because exit was 0), which is exactly the thing a future reader needs to know before "simplifying" it back.
- Comment on L161-164 explains the degraded-render contract explicitly.
- `_safe_json` catches `TypeError` and `ValueError` alongside `JSONDecodeError` — belt-and-suspenders for `None` / non-string argv edge cases.
- E2E verification in the PR body actually ticks the unchecked box from #95, which was the nagging loose end.
- Small, focused diff. Good squash candidate as you noted.

ylohnitram:
## QA Report — PR #97

Verified against branch `fix/issue-96-dashboard-fresh-federation` at 9db2616, agentis v1.1.7, python 3.14, against a fresh sandbox under `/tmp/smoke96*`. All checks re-run by me locally; boxes ticked below reflect what I personally re-verified.

### Summary
- **Script syntax**: pass (`bash -n tools/federation-dashboard.sh` returns 0)
- **Python heredocs**: both compile (`py_compile` clean on the history-append `PY` block and on `PYSERVER`)
- **#96 reproduction on `main`**: reproduced exactly (stderr `json.decoder.JSONDecodeError: Expecting value: line 1 column 2 (char 1)`, rc=1). Root cause verified: on 2 fresh agents with no memos, `CONFIDENCES` is built as `"[,]"` (empty `conf` values each append `","`, and `${CONFIDENCES%,}` only strips ONE trailing comma).
- **#96 on this PR**: gone. Dashboard produces `fed/.dashboard/index.html` (22309 bytes) and valid `history.json` (1 entry on first run), no traceback in stderr.
- **Second run**: history.json grows to N+1 valid entries each invocation; all entries parse; 7-day cutoff still applied.
- **PR #95 markers**: `countdown`, `refresh-btn`, `manualRefresh` present (9 matches) in rendered HTML, both on fresh and memo-populated runs.
- **Regression with memos**: after `agentis memo set agent_a:confidence 0.5`, the injected JS literal is `[{"agent":"agent_a","colony":"colony1","confidence":0.5},{"agent":"agent_b","colony":"colony2","confidence":0.0}]` and `history.json`'s last entry has `"confidence": {"colony1": 0.5, "colony2": 0.0}`. The `${conf:-0.0}` default correctly fills in the unset agent without clobbering the set one. Newlines from `agentis memo get` output (e.g. `"0.5\n"`) are stripped by bash command substitution, so the concatenation remains valid JSON.

### Findings

| 1 | — | verification | `bash -n` passes on `tools/federation-dashboard.sh` | `python3 -c "import subprocess as s; print(s.run(['bash','-n','tools/federation-dashboard.sh']).returncode)"` → `0` | - [x] |
| 2 | — | verification | Both Python heredocs compile | `py_compile` on extracted `PY` + `PYSERVER` blocks | - [x] |
| 3 | HIGH (resolved) | regression | `#96` JSONDecodeError on fresh federation is fixed | `/tmp/smoke96fresh`: after `agentis init` + dashboard run, `history.json` is valid JSON with 1 entry, no traceback | - [x] |
| 4 | — | verification | Second run appends a valid entry | Ran dashboard twice in `/tmp/smoke96pr`; `history.json` grew to multiple entries, all parseable, timestamps monotonic | - [x] |
| 5 | — | verification | PR #95 countdown/refresh markers survive | `grep -c "countdown\|refresh-btn\|manualRefresh" index.html` → `9` | - [x] |
| 6 | — | verification | No regression with existing memos | `agentis memo set agent_a:confidence 0.5` then re-run; `avg_conf` per colony correct, JS literal well-formed, unset agent defaults to `0.0` | - [x] |

### Details

#### Root-cause trace (for the record)
Added a debug shim before the `python3 - "$HISTORY_FILE" …` invocation on the `main`-branch script and re-ran against a fresh federation:
```
KC=[{"colony1": 0, "colony2": 0, "total": 0}]
CONF=[[,]]
MAP=[[{"agent":"agent_a","colony":"colony1"},{"agent":"agent_b","colony":"colony2"}]]
Traceback (most recent call last):
  File "<stdin>", line 4, in <module>
  ...
json.decoder.JSONDecodeError: Expecting value: line 1 column 2 (char 1)
```
So `CONFIDENCES` is the offending value and `conf_vals = json.loads(sys.argv[4])` is the throwing line, exactly as the PR description calls out.

Under this PR, the same invocation produces `CONF=[[0.0,0.0]]` and the heredoc parses cleanly.

#### Fix coverage
The primary fix (`${conf:-0.0}` on line 134) repairs the single actual failure path observed in #96. The `_safe_json` helper in the Python heredoc (lines 155–159, and calls at 165–167) is defensive belt-and-braces for the other two args — not strictly required for #96, but a cheap safety net that keeps the dashboard rendering if any shell-assembled JSON ever gets malformed in future. No objection.

#### Caveats
- I did not exercise the running `PYSERVER` / `/kill` endpoint end-to-end (would require spawning a real daemon). Only its syntactic validity was confirmed.
- I did not ssh-verify or browser-render the generated `index.html`; I only grepped for markers and validated injected JSON syntax. If you want a headless-browser verification, happy to run that.
- No changes to non-PR files; this PR is a single-file, +18/-6 diff.

### Verdict
No defects found. The PR fixes #96 as claimed and does not regress PR #91 / PR #95 behaviour. LGTM from a QA standpoint.

---

_Re-verified locally by running every command in the PR description and the checklist above; any ticked box in this comment reflects a re-verification I ran myself, not a claim transferred from the PR author._

ylohnitram:
Addressed both reviewer suggestions in 62e68b0:

1. **L133 `set -e` regression** — restored `|| echo ''` so non-zero exits from `agentis` (binary missing, store lock) don't abort the script. The `${conf:-0.0}` default handles empty-stdout (missing key), and `|| echo ''` handles non-zero exit. Comment updated to reflect the split.
2. **L155-159 silent swallow** — `_safe_json` now writes to stderr on parse failure with a label (`knowledge_counts` / `confidences` / `agent_colony_map`) and the exception detail. Fresh-federation noise is brief (empty strings fail `json.loads` for one tick per agent); a persistent daemon crash / store corruption is now diagnosable.

PR #101, opened 2026-04-14
fix(dashboard): manual refresh regenerates HTML server-side (#98)

### LINKED ISSUE #98: federation-dashboard: manual refresh button only reloads stale HTML, does not regenerate
## Problem

The dashboard's manual refresh button (↻ in the header, also bound to `r` key) is currently useless because it only triggers `location.reload()` in the browser. The served HTML is a **static file** (`.dashboard/index.html`) regenerated server-side **only every 60 seconds**. So pressing the button just re-fetches the same stale snapshot from disk.

When you kill the federation via `agentis daemon stop --all` (or the Kill Federation button) and then press refresh to see the new state, the dashboard keeps showing 21 agents as `running` for up to a full minute — because the HTML backing the page hasn't been regenerated yet.

## Repro

1. Start federation: `./start-federation.sh`
2. Open dashboard at http://localhost:8420
3. `agentis daemon stop --all` (or dashboard kill switch)
4. Immediately click the ↻ refresh button or press `r`
5. Dashboard still shows 21 running agents

Expected: refresh reflects real current state ("No running daemons").
Actual: stale snapshot is re-served until the next background `generate` tick (up to 60s later).

## Root cause

`tools/federation-dashboard.sh`:

```js
function manualRefresh() {
  const btn = document.getElementById('refresh-btn');
  if (btn) btn.classList.add('spinning');
  setTimeout(() => location.reload(), 150);
}
```

Server-side, `generate` is only called at startup and every 60s in a background loop. The HTTP handler is a plain `SimpleHTTPRequestHandler` that serves `.dashboard/index.html` as a static file.

## Suggested fix

Either:

1. **Regenerate on manual refresh** — add a `POST /refresh` endpoint that runs `generate` server-side, then `manualRefresh()` awaits that POST before reloading:
   ```js
   function manualRefresh() {
     btn.classList.add('spinning');
     fetch('/refresh', { method: 'POST' }).finally(() => location.reload());
   }
   ```
2. **Regenerate on GET when stale** — on each `GET /` (or `/index.html`), if `.dashboard/index.html` mtime is older than N seconds (e.g. 5), call `generate` before serving. This makes every page load fresh.

Option 1 preserves the background tick for passive viewing and gives the operator an explicit "force-fresh" path. Option 2 is simpler but makes every visit slightly slower.

## Impact

Low-risk UX bug, but it undermines trust in the kill switch workflow: operator kills the federation, sees nothing changed, concludes the kill failed, starts troubleshooting. Happened to me today (took me a minute to realize the federation was actually dead, the UI was just lying).

## Environment

- agentis-colonies: master
- agentis: v1.1.7
- Browser: any (bug is server-side)

ylohnitram:
## Summary

Fixes #98. Manual refresh (↻ button + `r` key) now asks the server to regenerate `index.html` before reloading the page.

- Shell: `DASHBOARD_REGEN_ONLY=1` short-circuit — run `discover` + `generate`, exit 0, no server startup.
- Python server: new `POST /refresh` handler re-execs the script via `subprocess.run(['bash', SCRIPT_PATH, FED_DIR], env={'DASHBOARD_REGEN_ONLY':'1', ...})`, 30 s timeout, 200 on success / 500 on failure with stderr body.
- JS: `manualRefresh()` fetches `/refresh` before `location.reload()`. Errors are swallowed so the button never wedges.
- Concurrency (addressing review feedback in 451ea0e): `generate` now writes to `$HTML_FILE.tmp.$$` and `mv`s atomically; history append uses `{path}.tmp.{pid}` + `os.replace`. Readers never see a torn file.

The background 60 s loop is untouched — it still regenerates for passive viewers.

## Test plan

- [x] `bash -n` passes
- [x] Both Python heredocs (`PY`, `PYSERVER`) compile
- [x] E2E in `/tmp/smoke-98` (fresh federation, 2 colonies, 2 stub agents):
  - POST `/refresh` returns HTTP 200 body `ok`
  - `index.html` mtime advances on POST
  - `history.json` appends a new entry
- [x] Standalone `DASHBOARD_REGEN_ONLY=1 bash federation-dashboard.sh <fed>` runs to completion, writes HTML, appends history
- [x] `/kill` endpoint still works (regression): verified 200 `0 daemon(s) stopped\nNo running daemons to stop.` on empty fed in `/tmp/smoke-98c`
- [x] Concurrent stress: 5 parallel POST /refresh all 200, HTML stays valid (open+close tags, 23 KB), history.json parses with all entries, no `.tmp.*` leftovers
- [x] `/unknown` → 404, `GET /` → 200 (router unchanged)
- [ ] End-to-end browser verification (kill federation → refresh → dashboard reflects real state within seconds, not 60 s) — unchecked until independently browser-tested in a real browser

### DIFF
--- a/tools/federation-dashboard.sh
+++ b/tools/federation-dashboard.sh
@@ -93,6 +93,11 @@ generate() {
     TIMESTAMP="$(date '+%Y-%m-%d %H:%M:%S')"
     local EPOCH
     EPOCH="$(date '+%s')"
+    # Per-PID temp file so concurrent generate() invocations (60 s background
+    # loop + POST /refresh subprocess, see #98) don't stomp on each other or
+    # let a SimpleHTTPRequestHandler read a half-written index.html. Final
+    # `mv` below is atomic on the same filesystem.
+    local HTML_TMP="$HTML_FILE.tmp.$$"
 
     local DAEMONS
     DAEMONS="$(agentis daemon list --json 2>/dev/null || echo '[]')"
@@ -153,7 +158,7 @@ print(json.dumps(lines))
 
     # Append to history
     python3 - "$HISTORY_FILE" "$EPOCH" "$KNOWLEDGE_COUNTS" "$CONFIDENCES" "$AGENT_COLONY_MAP" <<'PY'
-import sys, json
+import sys, os, json
 def _safe_json(s, default, label):
     try:
         return json.loads(s)
@@ -190,8 +195,12 @@ entry = {"t": epoch, "knowledge": kc, "confidence": avg_conf}
 history.append(entry)
 cutoff = epoch - 7 * 86400
 history = [h for h in history if h["t"] > cutoff]
-with open(path, "w") as f:
+# Per-PID temp + os.replace: atomic on the same filesystem so concurrent
+# generate() writers don't leave a partially-written history.json readable.
+tmp = f"{path}.tmp.{os.getpid()}"
+with open(tmp, "w") as f:
     json.dump(history, f)
+os.replace(tmp, path)
 PY
 
     local HISTORY
@@ -368,7 +377,7 @@ HEADEOF
 </head>
 <body>
 HTMLEOF
-    } > "$HTML_FILE"
+    } > "$HTML_TMP"
 
     {
     cat <<HEADEREOF
@@ -694,7 +703,11 @@ function killSwitch() {
 </body>
 </html>
 JSEOF
-    } >> "$HTML_FILE"
+    } >> "$HTML_TMP"
+    # Atomic publish: mv(2) on the same filesystem replaces the inode in a
+    # single syscall, so any HTTP reader sees either the old or the new file
+    # in full — never a truncated-in-progress state.
+    mv "$HTML_TMP" "$HTML_FILE"
 }
 
 # --- Regen-only mode ---
--- a/tools/federation-dashboard.sh
+++ b/tools/federation-dashboard.sh
@@ -471,8 +471,14 @@ document.getElementById('clock').textContent = timestamp;
 function manualRefresh() {
   const btn = document.getElementById('refresh-btn');
   if (btn) btn.classList.add('spinning');
-  // Short delay so the spin is visible even on near-instant reloads.
-  setTimeout(() => location.reload(), 150);
+  // #98: ask the server to regenerate the static HTML snapshot before we
+  // reload. The background loop only runs every 60 s, so without this POST
+  // a manual refresh would just re-render the stale on-disk file. Errors
+  // are swallowed — reload anyway so the operator sees *something* rather
+  // than a wedged spinner.
+  fetch('/refresh', { method: 'POST' })
+    .catch(() => {})
+    .finally(() => location.reload());
 }
 
 document.addEventListener('keydown', (e) => {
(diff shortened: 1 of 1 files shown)


ylohnitram:
## PR #101 Review

### Verdict: REQUEST CHANGES (concurrency) — otherwise solid

### Blocking

| # | File | Line | Issue |
|---|------|------|-------|
| 1 | tools/federation-dashboard.sh | 371 vs 697 vs 720-724 | `generate` writes `$HTML_FILE` as truncate (`>`) followed by a separate append (`>>`) block. It is not atomic. The 60 s background loop and `POST /refresh` now both call `generate` and race on the same file — an operator reloading during the window between the two redirects can see an empty/half-written page, and simultaneous writers will interleave bytes producing malformed HTML. Fix: write to a temp file and `mv` into place, or take an `flock` around `generate`. `$HISTORY_FILE` has the same read-modify-write race (L177-194). #98 makes this much more likely to fire in practice because manual refresh is the most common way to hit it. |

### Non-blocking

| # | File | Line | Note |
|---|------|------|------|
| 1 | tools/federation-dashboard.sh | 79-82 | On re-exec the pre-guard check `if COLONY_COUNT -eq 0 ... exit 1` runs before the regen-only short-circuit. A transient empty federation dir makes `POST /refresh` return 500. Surfaces real state, but worth noting — consider moving the empty check below the `DASHBOARD_REGEN_ONLY` guard so refresh-of-empty-fed regenerates a "0 colonies" page rather than 500ing. |
| 2 | tools/federation-dashboard.sh | 730, 752 | `fed_dir_arg` is passed as `$FED_DIR` (already resolved, trusted local value — not from the HTTP request). Script-execution-via-subprocess is safe: `subprocess.run(['bash', ...])` with a list argv, no shell=True, `timeout=30`, fixed env. Low risk. |
| 3 | tools/federation-dashboard.sh | 474-479 | JS `.catch(() => {}).finally(() => location.reload())` reloads unconditionally. On 500 the operator silently reloads stale HTML with no visible failure hint. Acceptable per PR description, but an alternative: in `.catch` flash a short "refresh failed" badge for 2 s before reloading. Not a blocker. |
| 4 | tools/federation-dashboard.sh | 742-814 | `/refresh` correctly `return`s before the `/kill` branch, and the `else: 404` still covers unknown paths. Regression risk is low. |

### Positive

- Correct handling of absolute vs repo-relative fed dirs via the existing `$REPO_ROOT/$1` fallback.
- `SCRIPT_PATH` is `realpath`-resolved at startup, so the re-exec is robust to relative argv[0].
- 30 s timeout + `capture_output` + 500-with-body contract is tight and makes failures debuggable.
- Subprocess call is list-form with fixed env; no shell injection surface.
- `/kill` endpoint untouched and routing remains correct.

ylohnitram:
Addressed the blocking concurrency finding in 451ea0e.

`generate` now writes to `$HTML_FILE.tmp.$$` and `mv`s onto `$HTML_FILE` once the buffer is complete — readers always see either the old or the new file in full, never a torn state. The Python history append uses the same pattern (`{path}.tmp.{pid}` + `os.replace`).

Verified under concurrent stress (5 parallel POST /refresh): all 200, `index.html` valid (23 KB, open/close tags intact), `history.json` parses with all 6 entries (1 initial + 5 refreshes), no `.tmp.*` leftovers in `.dashboard/`.

**QA note:** the QA agent I launched was sandbox-blocked (couldn't run `bash`, `python3`, `curl`, etc.), so I ran the QA steps inline. End-to-end-browser verification (the other unchecked box) still genuinely isn't done — I did not open the page in a real browser. Leaving that unticked per standing checkbox discipline.

**Non-blocking items left as-is:**
- L79-82 empty-federation check running before regen short-circuit: a transient empty dir would still 500 on /refresh, but in practice the dashboard is launched after the federation exists and the check is there to fail loudly on genuine misconfiguration. Willing to move it if you feel strongly.
- JS `.catch(()=>{}).finally(reload)`: a brief in-page error hint would be nicer, but the current behavior (swallow + reload) matches how the rest of the dashboard handles backend hiccups.

PR #102, opened 2026-04-14
docs(dev-apprenticeship): drop non-existent personal/project: tag framing (#99)

### LINKED ISSUE #99: dev-apprenticeship: README advertises personal/project: knowledge tags that don't exist in implementation
## Problem

`dev-apprenticeship/README.md` has a section **Knowledge portability** that promises:

> Knowledge is tagged by scope: \`personal\` (your preferences, portable across projects) and \`project:<name>\` (codebase-specific, stays with the project).
>
> \`\`\`bash
> agentis knowledge export --tags personal > my-preferences.json   # carry to new project
> agentis knowledge import my-preferences.json --merge
> \`\`\`

Neither tag is ever produced by any of the 21 agents. Every `learn()` call in `*/agents/*.ag` tags with `["observed", "<colony>"]` or `["acted", "<colony>"]` only. Confirmed by:

```bash
$ grep -rn -E '"personal"|"project:' dev-apprenticeship/ tools/
# no matches

$ agentis knowledge export --tags personal
[]
```

So the documented workflow is a no-op: export with `--tags personal` always returns `[]`, import of that empty file carries nothing.

## Why this matters

It's the primary onboarding workflow the README sells for moving from one project to another (e.g. running one federation on `Y1` and a second on `app`). Users following the README will think knowledge transfer "works" because the commands succeed — but zero entries actually transfer. By the time they notice (days/weeks later, when the second federation is still making rookie suggestions), they've lost confidence in the whole apprenticeship model.

Related: there's also no mechanism to distinguish the operator's own GitLab activity from the rest of the project's participants. `scripts/gitlab-api.sh` queries endpoints like `$API/issues` without any author filter, so every agent learns from the full project team collectively. Calling any of that "personal" is misleading even in principle — the agents have no concept of who "you" are.

## Two ways to resolve

**Option A — Remove the false promise from the README.**
Simplest. Drop the Knowledge portability section, or reword it as future work. The `export/import` commands still exist at the runtime level; document them without the personal/project framing.

**Option B — Make the feature real.**
More work, but would make the apprenticeship actually personal:

1. Plumb an operator identity into the federation config (e.g. `gitlab.me = "martin.holy"` in each `colony.toml`, exported as `GITLAB_ME` by `start-colony.sh`).
2. Have `gitlab-api.sh issues` / `merge_requests` / `notes` endpoints expose each record's `author.username`.
3. In every `.ag` `learn()` call that observes human behavior, branch on author:
   - `author == $GITLAB_ME` → add `"personal"` tag
   - else → add `"team"` tag (or `"project:" + project_name`)
4. Update README's example to match what the tags actually mean.

## Suggested scope

Option A is a 10-minute fix and unblocks trust immediately. Option B is the real feature and can follow as its own issue/MR if the project owner wants it.

## Environment

- agentis-colonies: master
- agentis: v1.1.7

ylohnitram:
## Summary

Fixes #99. The README promised `personal` and `project:<name>` knowledge tag scopes, but no agent ever emits those tags — every `learn()` call uses `[observed|acted, <colony>]`. The documented `agentis knowledge export --tags personal` command silently returns `[]`.

Rewrote the Knowledge portability section to describe the real tag scheme and note that:
- Bulk `export` / `import` work for all entries
- `--tags acted` / `--tags <colony>` are real, useful filters
- `--tags personal` returns `[]` today — there's no operator-identity concept

Option B from #99 (author-aware tagging) can follow as a separate issue/MR. This PR just stops the README from lying.

## Test plan

- [x] Claim verified: `grep -rn '"personal"\|"project:"' dev-apprenticeship/ tools/` returns no matches on `main`
- [x] Claim verified: `agentis knowledge export --tags personal` returns `[]`
- [x] Agents use `[observed|acted, <colony>]` tags (grep of `*/agents/*.ag` confirms)
- [ ] Independent read-through of the new wording — unchecked until reviewer confirms it's clear and doesn't introduce new false promises

### DIFF
--- a/dev-apprenticeship/README.md
+++ b/dev-apprenticeship/README.md
@@ -208,12 +208,16 @@ Cross-colony wiring: Triage routes issues to Implementation. Implementation sign
 
 ## Knowledge portability
 
-Knowledge is tagged by scope: `personal` (your preferences, portable across projects) and `project:<name>` (codebase-specific, stays with the project).
+Every `learn()` call in the federation's agents tags entries with `observed` or `acted` plus the colony name (`triage`, `code-review`, `planning`, `implementation`, `release`). There is no built-in distinction between "your preferences" and "codebase-specific" knowledge — agents see the full GitLab project activity without any concept of who `you` are, so everything learned is effectively project-scoped.
+
+Bulk export/import is available at the runtime level and will carry all entries as-is:
 
 ```bash
-agentis knowledge export --tags personal > my-preferences.json   # carry to new project
-agentis knowledge import my-preferences.json --merge
+agentis knowledge export > fed-knowledge.json
+agentis knowledge import fed-knowledge.json --merge
 ```
+
+Filtering by `--tags acted` or `--tags <colony>` works (those are real tags the agents emit), but there is no `personal` or `project:<name>` tag today — if you try `--tags personal` you will get an empty array.
 
 ## Troubleshooting
 
--- a/dev-apprenticeship/README.md
+++ b/dev-apprenticeship/README.md
@@ -208,7 +208,7 @@ Cross-colony wiring: Triage routes issues to Implementation. Implementation sign
 
 ## Knowledge portability
 
-Every `learn()` call in the federation's agents tags entries with `observed` or `acted` plus the colony name (`triage`, `code-review`, `planning`, `implementation`, `release`). There is no built-in distinction between "your preferences" and "codebase-specific" knowledge — agents see the full GitLab project activity without any concept of who `you` are, so everything learned is effectively project-scoped.
+Every `learn()` call in the federation's agents tags entries with one of `observed` (passive learning from GitLab activity), `emitted` (suggestion logged at confidence 0.6-0.84), or `acted` (autonomous action taken at ≥ 0.85) plus the colony name (`triage`, `code-review`, `planning`, `implementation`, `release`). There is no built-in distinction between "your preferences" and "codebase-specific" knowledge — agents see the full GitLab project activity without any concept of who `you` are, so everything learned is effectively project-scoped.
 
 Bulk export/import is available at the runtime level and will carry all entries as-is:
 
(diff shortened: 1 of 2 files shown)


ylohnitram:
## PR #102 Review

### Verdict: REQUEST CHANGES

The direction is right and the rewrite correctly drops the fabricated `personal` / `project:<name>` tag scheme. Two concrete issues before merge.

### Critical Issues

| # | File | Line | Issue |
|---|------|------|-------|
| 1 | `dev-apprenticeship/code-review/README.md` | 97-98 | Sibling README still advertises `agentis knowledge export --tags personal > my-preferences.json` as the way to export "personal knowledge (portable to other projects)". This is the exact same false promise #99 is fixing — if we only patch `dev-apprenticeship/README.md`, a reader who followed the code-review README into the repo still gets an empty array. Please update this file (and the accompanying "Import on a new project" line) in the same PR. |
| 2 | `dev-apprenticeship/README.md` | 211 (new) | The rewrite says agents tag entries with "`observed` or `acted` plus the colony name". Grep shows a third tag in active use: `emitted` (13 occurrences across 12 agents — every `partial` branch in `release/*`, `implementation/*`, and it appears alongside `acted` in `version_bumper.ag:155`, `release_checker.ag:104,123`, `ship_decider.ag:103`, `changelog_writer.ag:134`, `code_writer.ag:138`, `commit_composer.ag:134`, `refactorer.ag:123`, `test_writer.ag:124`). Either list all three (`observed` / `acted` / `emitted`) or phrase it as "a status tag (`observed`, `acted`, or `emitted`) plus the colony name" so the "real tags" list is actually complete. Otherwise the new wording reintroduces a smaller version of the same bug — `--tags emitted` works but is undocumented. |

| 1 | `dev-apprenticeship/README.md` | 219 (new) | While listing valid filters, consider mentioning `--tags observed` and `--tags emitted` explicitly, same as `acted`, so all three status filters surface in one place. |

### Positive Notes

- Claim verification is solid: `personal` and `project:` really are absent from all `.ag` files and `tools/`. Root `README.md` is clean.
- Honest framing about the lack of an operator-identity concept is exactly the right note to strike, and deferring author-aware tagging to a separate issue keeps this PR focused.
- No command-syntax regressions in the new examples (`agentis knowledge export` / `import ... --merge` both match the actual CLI).

ylohnitram:
Addressed both review findings in 043f5f4:

1. **`dev-apprenticeship/code-review/README.md` L97-98**: the sibling file still advertised `agentis knowledge export --tags personal`. Rewritten to use the real tags (`code-review`, `acted`, `emitted`) and document bulk export/import without the personal/project framing.
2. **Missing `emitted` tag**: confirmed `emitted` is used in 13 `learn()` calls across implementation/planning/release (all `partial`-status, i.e. suggestions at 0.6-0.84). Added to both the narrative tag list and the `--tags` filter example in the main README.

Re-verified with `grep -rln 'personal\|project:' dev-apprenticeship/` — no matches remain.

Issue #109, opened 2026-04-14
dev-apprenticeship: 9+ agents fail every tick with 'type error: expected list or map, got string' (LLM returns markdown-wrapped JSON)


ylohnitram:
## Problem

After fixing #107 (PII guard + exec timeout via v1.1.8), a new class of agents fails every tick with:

\`\`\`
1776155495510 [tick:error] type error: expected list or map, got string
1776155569170 [tick:error] type error: expected list or map, got string
1776155655774 [tick:error] type error: expected list or map, got string
... (every 60-90s, indefinitely)
\`\`\`

Affected on my run (15 min observation):
- All 5 code-review agents: \`style_reviewer\`, \`logic_reviewer\`, \`security_reviewer\`, \`test_reviewer\`, \`approval_decider\`
- 4 implementation agents: \`code_writer\`, \`test_writer\`, \`refactorer\`, \`commit_composer\`

That's **9/21 agents** (43%) doing zero useful work.

## Root cause

These agents share a pattern: \`prompt(...) -> list<int>\` or \`prompt(...) -> list<string>\`. Example from \`code-review/agents/style_reviewer.ag:57\`:

\`\`\`ag
let mr_iids = prompt(
    "Extract all merge request IIDs from this GitLab JSON array. Return a JSON array of integers. Only include MRs in 'opened' state. If empty, return [].",
    mrs_json
) -> list<int>;
\`\`\`

Claude (CLI backend) frequently wraps JSON output in markdown fences (\`\`\`json [1,2,3] \`\`\`) instead of the bare \`[1, 2, 3]\` the prompt requested. The agentis runtime parser then sees a \`string\` (the whole markdown blob) where the agent's \`-> list<int>\` annotation expects a JSON array, and aborts the tick with \`type error: expected list or map, got string\`.

This is a contract failure between the dev-apprenticeship agent prompts and the agentis core response parser. **No agent recovers from this** — every subsequent tick re-issues the same prompt and gets the same wrapped-JSON, so the loop is permanent.

## Two ways to fix

**Option A (agent-side, in this repo) — tighten prompts.**

Every \`prompt(...) -> list<...>\` and \`prompt(...) -> map<...>\` site needs an explicit anti-markdown directive. Something like:

\`\`\`ag
"Return ONLY a JSON array of integers, no markdown code fences, no prose, no \`\`\`json wrapper. Just the raw array."
\`\`\`

[...]

Issue #107, opened 2026-04-14
dev-apprenticeship: federation cannot complete a single tick on a real GitLab project (PII guard + exec timeout both block)


ylohnitram:
## Problem

After running the README-documented install and start workflow against a real self-hosted GitLab project with a few hundred issues and MRs, **not a single agent out of 21 ever completes a successful tick**. Two independent runtime gates block every tick, both rooted in agentis-core defaults that dev-apprenticeship does not configure around.

After ~10 minutes of running, `agentis knowledge stats` still reports "No knowledge base found" and `agentis daemon list` shows the full federation as mostly `degraded` or `critical`.

### Blocker 1: PII guard denies every prompt

Every `triage`/`code-review`/`planning` agent hits this on its first tick and every tick thereafter:

```
[tick:error] capability denied: missing capability: pii_transmit
  at: prompt("Analyze this GitLab issues JSON. ...", <input>) -> LabelAnalysis
  Hint: PII detected: email, phone
  Hint: input length: 74006 chars
  Hint: grant PiiTransmit via --grant-pii or config pii_transmit = allow
```

GitLab API payloads universally contain `author.email`, `assignee.email`, phone-shaped numbers in free-text descriptions. This is not avoidable in practice — it's not a corner case, it's every project.

The hint recommends two mitigations; **neither actually works**:

1. **`config pii_transmit = allow`** — I added this to `.agentis/config`. `agentis doctor` confirms the config loaded. Daemons still deny. The documented config key appears to be silently ignored, or read from a location I can't identify.

2. **`--grant-pii` CLI flag** — Not supported by `agentis daemon`. Only `agentis go` accepts it:
   ```
   $ agentis daemon /tmp/x.ag --grant-pii
   agentis daemon: unknown flag: --grant-pii
   $ agentis daemon --help
   ... (no --grant-pii listed)
   ```

So the only workflow documented for unblocking PII is unavailable in the daemon path, and the config fallback doesn't take effect. This should be confirmed against agentis-core, but from the dev-apprenticeship side the outcome is the same: **the hint in the error message is not a working workaround**.

### Blocker 2: exec sh timeout too short for GitLab API

Every GitLab polling call logs variants of:

```
[router] GitLab poll failed: Error {
  code: 1003,
  message: ExecTimeout: exec sh timed out after 10000ms,
  category: transient
}
```

[...]

PR #113, opened 2026-04-14
fix(dev-apprenticeship): enable adaptive engine in install.sh

### LINKED ISSUE #110: dev-apprenticeship: router and prioritizer fail every tick with 'adaptive engine not enabled' — recommend() needs config key
## Problem

After fixing #107 (PII guard + exec timeout via v1.1.8), \`triage/router\` and \`triage/prioritizer\` fail every tick with:

\`\`\`
1776155510623 [tick:error] adaptive engine not enabled
1776155657822 [tick:error] adaptive engine not enabled
... (every 60-90s, indefinitely)
\`\`\`

These two agents are the entry points of the triage colony — without them, no incoming GitLab issue is ever **routed** (mapped to a destination colony) or **prioritized** (assigned a priority label suggestion). The triage colony silently degrades to "labeler + issue_creator only".

## Root cause

Both agents call \`recommend()\` from agentis-core's adaptive engine. Example from \`triage/agents/router.ag:77\`:

\`\`\`ag
let rec = recommend("route", ["accuracy"]);
\`\`\`

The agentis runtime returns the \`adaptive engine not enabled\` error, which means the engine is gated behind a capability/feature flag that is **off by default**. Same pattern as PII (#107) before v1.1.8 fixed it.

\`install.sh\` writes several defaults to \`.agentis/config\` (\`federation.enabled\`, \`knowledge.federation_enabled\`, \`experience.enabled\`, etc.) but does **not** enable the adaptive engine. Result: dev-apprenticeship installs in a state where \`recommend()\` is broken from tick 1.

## Suggested fix

**Option A (this repo) — write the enable flag from install.sh.**

Add to \`install.sh\`'s defaults block, alongside \`experience.enabled = true\`:

\`\`\`bash
write_key 'adaptive.enabled' 'true'
\`\`\`

(Exact key name to be confirmed against agentis-core — \`adaptive.enabled\`, \`adaptive_engine.enabled\`, or \`recommend.enabled\` depending on the convention. The hint in the error message would help.)

**Option B (core-side, optional) — log the actual config key needed.**

The current error \`adaptive engine not enabled\` does not tell the operator how to enable it. Adding a hint line — same pattern as the PII error in v1.1.7 (\"grant via --grant-pii or config X = Y\") — would help anyone who hits this in a non-dev-apprenticeship context.

## Why this matters

\`router\` and \`prioritizer\` are how the triage colony **uses** what \`labeler\`/\`issue_creator\` observe. Without them:

- Cross-colony events (\`triage:route_suggestion -> implementation/code_writer\`) never fire.
- The implementation colony has no incoming work queue to learn from.
- The whole \"Triage routes issues to Implementation\" wiring described in dev-apprenticeship's README is dead.

So this is not a \"two agents are sad\" issue — it's a \"two of the five colonies are silently disconnected\" issue.

## Repro

1. Install dev-apprenticeship per README (master, with v1.1.8).
2. \`./start-federation.sh\`
3. \`tail .agentis/logs/<router-id>.log\` → \"adaptive engine not enabled\" every tick.

## Environment

- agentis: v1.1.8

## Related

- #107 (PII guard — same install.sh-doesn't-enable-required-capability pattern, fixed by writing \`pii_transmit = allow\` default)
- This issue is the next instance of that pattern. May be worth a broader audit of which other capabilities \`install.sh\` needs to enable.

ylohnitram:
## Summary

`triage/router` and `triage/prioritizer` both call `recommend()` on every tick. In agentis-core, `recommend()` goes through the adaptive engine, which is only attached to the evaluator when `learning.enabled = true` is set in `.agentis/config`. The relevant core wiring is at `src/cli/daemon.rs:959`:

```rust
if cfg.get("learning.enabled").is_some_and(|v| v == "true") {
    // ...
    evaluator = evaluator.with_adaptive_engine(engine);
}
```

Without the key, `recommend()` raises `adaptive engine not enabled` and the tick aborts before any routing/priority work runs. `install.sh` was already writing a set of required defaults (`experience.enabled`, `pii_transmit`, the `exec.*` keys for #107, etc.), but missed this one — so fresh installs had router + prioritizer dead on arrival.

The same error also blocks the 9 code-review/implementation agents (`style_reviewer`, `logic_reviewer`, `security_reviewer`, `test_reviewer`, `approval_decider`, `code_writer`, `test_writer`, `refactorer`, `commit_composer`) once #109 is merged — they all call `recommend()` too, just later in their tick after the `list<int>` MR extraction. So this also unblocks the post-#109 federation.

## What this changes

One-line `write_key` addition + 10-line doc comment block explaining why:

```bash
write_key 'learning.enabled' 'true'
```

No other changes. Idempotent re-runs of `install.sh` log "already configured" for the key on the second run, same as every other `write_key` call.

Fixes Replikanti/agentis-colonies#110

## Test plan

- [ ] Fresh `./install.sh` writes `learning.enabled = true` to `.agentis/config`.
- [ ] Re-running `./install.sh` does not duplicate the key.
- [ ] After a fresh install, `tail .agentis/logs/router.log` no longer shows `adaptive engine not enabled`.

### DIFF
--- a/dev-apprenticeship/install.sh
+++ b/dev-apprenticeship/install.sh
@@ -233,16 +233,24 @@ if [ -d "$AGENTIS_DIR" ]; then
     #                                 tick and the federation never makes
     #                                 progress. Only honored by the daemon
     #                                 path from agentis v1.1.8 onward.
-    #   learning.enabled            — triage/router and triage/prioritizer
-    #                                 (and several code-review agents) call
+    #   learning.enabled +           triage/router and triage/prioritizer
+    #   knowledge.enabled             (and several code-review agents) call
     #                                 recommend() from the adaptive engine
-    #                                 on every tick. The engine is only
-    #                                 wired into the evaluator when this
-    #                                 key is set; otherwise recommend()
-    #                                 raises `adaptive engine not enabled`
-    #                                 (issue #110) and the tick aborts
-    #                                 before any routing/priority work
-    #                                 happens.
+    #                                 on every tick. In agentis-core
+    #                                 (evaluator/mod.rs:8093), recommend()
+    #                                 requires all three of adaptive_engine
+    #                                 (gated by learning.enabled), the
+    #                                 experience_store (gated by
+    #                                 experience.enabled — already above),
+    #                                 AND the knowledge_base (gated by
+    #                                 knowledge.enabled). Missing any one
+    #                                 aborts the tick with a different
+    #                                 "<component> not enabled" message.
+    #                                 Setting only learning.enabled would
+    #                                 flip the error from "adaptive engine
+    #                                 not enabled" (issue #110) to
+    #                                 "knowledge base not enabled" on the
+    #                                 very next tick, so we write both.
     AGENTIS_CONFIG="$AGENTIS_DIR/config"
     # Agentis init should have created this; create it explicitly so the
     # key-writing below is not silently skipped if init was interrupted.
@@ -280,6 +288,7 @@ if [ -d "$AGENTIS_DIR" ]; then
     write_key 'exec.env_passthrough'         'COLONY_DIR,GITLAB_*'
     write_key 'exec.default_timeout_ms'      '45000'
     write_key 'pii_transmit'                 'allow'
+    write_key 'knowledge.enabled'            'true'
     write_key 'learning.enabled'             'true'
 fi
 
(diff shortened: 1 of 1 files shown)


ylohnitram:
Reviewer flagged a critical miss: `recommend()` in agentis-core/src/evaluator/mod.rs:8093 needs all THREE components (adaptive_engine + experience_store + knowledge_base). The previous commit only enabled the first two (`learning.enabled` + existing `experience.enabled`), which would have flipped the error from "adaptive engine not enabled" to "knowledge base not enabled" on the next tick.

Pushed 401e78e which adds `write_key 'knowledge.enabled' 'true'` and expands the inline comment to document the three-way dependency.

PR #114, opened 2026-04-14
fix(dashboard): count experience entries, not knowledge base (#111)

### LINKED ISSUE #111: federation-dashboard: 'knowledge entries' counter reads agentis knowledge but learn() writes to experience — dashboard always shows 0
## Problem

The dashboard shows **0 knowledge entries** for every colony, indefinitely, even after multiple agents have successfully completed observation ticks and accumulated real learned data.

In my current run (after #107 fixes via v1.1.8, with #109/#110 still blocking ~half the federation):

\`\`\`
$ ls -la .agentis/experience/
0e241f497d83.jsonl   765 B   (labeler:         1 entry)
39ae5d4be590.jsonl   764 B   (issue_creator:   1 entry)
c05d651bac61.jsonl  6.5 KB   (task_decomposer: 7 entries)
c0678685a958.jsonl  14 KB    (scope_estimator: many)
edac874ae52a.jsonl  16 KB    (risk_assessor:   many)

$ agentis knowledge stats
No knowledge base found

$ agentis knowledge list
No knowledge base found

# Dashboard at :8420 → all colony knowledge counters show 0
\`\`\`

The agents are clearly learning — \`scope_estimator.jsonl\` is 14 KB of accumulated observations after a few ticks. But the dashboard, the operator's primary visibility surface, says \"nothing happened\".

## Root cause

\`tools/federation-dashboard.sh:113\` populates its knowledge counts via \`agentis knowledge list\`. But \`learn(...)\` calls in agents write to a **different store**: \`.agentis/experience/<agent-id>.jsonl\`. The agentis CLI exposes these via a separate subcommand:

\`\`\`
$ agentis knowledge --help    # list / show / stats / export / import
$ agentis experience --help   # show <agent-id> / summary <agent-id>
\`\`\`

Two parallel storage systems, two parallel CLIs. The dashboard reads one, agents write to the other.

## Why this matters

The dashboard's purpose is to give the operator \"is this thing actually doing something?\" confidence at a glance. With this bug:

- An operator sees \"0 knowledge entries\" and assumes the federation is broken.
- The \"phase ETA\" widget is meaningless without data to extrapolate from.
- The \"knowledge growth trends\" chart stays a flat zero line forever.

## Two ways to fix

**Option A (dashboard-side, this repo) — count experience entries.**

Replace the \`agentis knowledge list\` call with one that counts entries per agent in \`.agentis/experience/\`, then aggregates per colony via the agent→colony map. Quick, no upstream dependency.

**Option B (agentis-core side, optional) — clarify the storage model.**

If \`knowledge\` and \`experience\` are intended to be different (raw observations vs. distilled patterns), dev-apprenticeship is missing the distillation step. That's a separate feature. If they're the same thing, one CLI should alias the other.

Either way, the docs and the dashboard need to pick which store represents \"how much has the agent learned\" and use it consistently.

## Recommended path

**A first.** Reflect what \`learn()\` actually writes — that's what the operator wants to see. B is a real product question for agentis-core and shouldn't block the dashboard fix.

## Repro

1. Run dev-apprenticeship per README, get past #107 (v1.1.8 + install.sh defaults).
2. Wait for any agent to complete a successful tick (\`labeler\`, \`issue_creator\`, \`task_decomposer\`, \`scope_estimator\`, \`risk_assessor\` all currently work).
3. \`agentis experience summary <agent-id>\` → real entries.
4. Open dashboard at :8420 → \"0 knowledge entries\" everywhere, forever.

## Environment

- agentis: v1.1.8

## Related

- #98 (dashboard refresh — fixed)
- #99/#100 (README fixed, features pending)
- #107 (PII + exec timeout — fixed via v1.1.8)
- #109 (type error blocking 9 agents)
- #110 (adaptive engine blocking router/prioritizer)

ylohnitram:
Closes #111.

## Summary

- The dashboard queried `agentis knowledge list`, but agents write with `learn(...)` which persists to `.agentis/experience/<agent-id>.jsonl` — two different stores. On any real federation the counter stayed 0 forever even after thousands of observed ticks.
- Replace the CLI call with direct line-count of JSONL files in `.agentis/experience/`, mapping each file's hex `agent_id` back to a colony via `agentis daemon list --json` (basename of `source` → `AGENT_COLONY_MAP` entry).
- JSON output shape `{total, <colony>}` is preserved for history.json/HTML chart compatibility.
- Rename history key, HTML labels, JS identifiers from `knowledge` → `experience` so the dashboard is consistent about which store it represents (per the recommendation in the issue body).

## Test plan

- [ ] Fresh federation (no daemons): `{"total": 0}`, chart shows "Trend available after 2+ data points", Experience Entries stat = 0.
- [ ] Federation with agents running and at least one successful `learn()` on `scope_estimator` / `risk_assessor` / `task_decomposer`: Experience Entries > 0, per-colony legend counts match `wc -l .agentis/experience/*.jsonl` after agent_id→colony mapping.
- [ ] Old history.json with `knowledge` keys: Experience Growth chart silently skips pre-upgrade entries (no crash, just grows fresh from first post-upgrade tick).
- [ ] Bash syntax check passes.
- [ ] Python counter unit test with fixtures: 3 agents/jsonl files, 1 unknown-colony file filtered out, 1 agent with no jsonl — totals correct.

### DIFF
--- a/tools/federation-dashboard.sh
+++ b/tools/federation-dashboard.sh
@@ -109,26 +109,35 @@ generate() {
     # the knowledge base. Two parallel stores, two parallel CLIs — the
     # knowledge counter was always 0 on a real federation. We now count
     # JSONL line entries in the experience dir and aggregate per colony via
-    # the agent->colony map. The output JSON shape is still `{total, <col>}`
-    # for compatibility with history.json entries and the HTML chart.
+    # the daemon's own `colony` field. The output JSON shape is still
+    # `{total, <col>}` for compatibility with history.json and the HTML
+    # chart.
     #
     # Experience files are named by daemon-assigned `agent_id` (opaque hex,
-    # see src/cli/daemon.rs:907), not the `.ag` basename. To map back we
-    # use `agentis daemon list --json`: its `source` field is the `.ag`
-    # file path, and basename(source, .ag) is our `AGENT_COLONY_MAP` key.
+    # see src/cli/daemon.rs:907), not the `.ag` basename. We map back via
+    # `agentis daemon list --json`: for daemons started with `--colony`
+    # (which `start-colony.sh` always does) the `colony` field is
+    # authoritative. For daemons started without it we fall back to
+    # basename(source, .ag) → AGENT_COLONY_MAP, which is lossy when two
+    # colonies ship an agent with the same filename but is the best we can
+    # do for bare `agentis daemon <file>` invocations.
     local COLONY_LIST_PY=""
     for col in "${COLONIES[@]}"; do
         COLONY_LIST_PY="${COLONY_LIST_PY}\"${col}\","
     done
     COLONY_LIST_PY="[${COLONY_LIST_PY%,}]"
 
+    # Match the LOG_DIR convention (line 187): prefer $FED_DIR/../.agentis,
+    # fall back to cwd-relative for operators invoking from the federation
+    # root. Experience dir existence is re-checked inside python so a
+    # missing path degrades to zeros rather than crashing.
     local EXPERIENCE_DIR="${FED_DIR}/../.agentis/experience"
     if [ ! -d "$EXPERIENCE_DIR" ]; then
         EXPERIENCE_DIR=".agentis/experience"
     fi
 
     local EXPERIENCE_COUNTS
-    EXPERIENCE_COUNTS="$(python3 - "$EXPERIENCE_DIR" "$DAEMONS" "$AGENT_COLONY_MAP" "$COLONY_LIST_PY" <<'PY' 2>/dev/null || echo '{"total":0}'
+    EXPERIENCE_COUNTS="$(python3 - "$EXPERIENCE_DIR" "$DAEMONS" "$AGENT_COLONY_MAP" "$COLONY_LIST_PY" <<'PY' 2>/dev/null
 import sys, os, json
 exp_dir, daemons_json, map_json, colony_list_json = sys.argv[1:5]
 try:
@@ -143,19 +152,25 @@ try:
     colonies = json.loads(colony_list_json)
 except (json.JSONDecodeError, ValueError):
     colonies = []
+# Basename-keyed fallback for daemons launched without --colony. Collisions
+# (same .ag filename across two colonies) silently last-writer-wins here,
+# which is a known limitation; the authoritative path uses d["colony"].
 name_to_colony = {e.get("agent", ""): e.get("colony", "") for e in agent_map}
 counts = {c: 0 for c in colonies}
 counts["total"] = 0
 if os.path.isdir(exp_dir):
     for d in daemons:
         agent_id = d.get("agent_id") or ""
-        source = d.get("source") or ""
-        if not agent_id or not source:
+        if not agent_id:
             continue
-        name = os.path.basename(source)
-        if name.endswith(".ag"):
-            name = name[:-3]
-        colony = name_to_colony.get(name)
+        colony = d.get("colony") or ""
+        if not colony:
+            source = d.get("source") or ""
+            if source:
+                name = os.path.basename(source)
+                if name.endswith(".ag"):
+                    name = name[:-3]
+                colony = name_to_colony.get(name, "")
         if not colony or colony not in counts:
             continue
         path = os.path.join(exp_dir, agent_id + ".jsonl")
@@ -174,6 +189,16 @@ if os.path.isdir(exp_dir):
 print(json.dumps(counts))
 PY
 )"
+    # The python block is wrapped in try/except for every input, so a bad
+    # arg silently yields `{"total":0,...}` — but if the interpreter itself
+    # dies (SIGKILL, disk full, bad shebang) we get an empty string and the
+    # `const experienceCounts = ${EXPERIENCE_COUNTS};` injection below would
+    # produce a JS syntax error that breaks every other IIFE on the page.
+    # Validate the result parses as JSON; on failure, substitute the same
+    # shape the python block emits on empty input.
+    if ! echo "$EXPERIENCE_COUNTS" | python3 -c 'import sys, json; json.loads(sys.stdin.read())' 2>/dev/null; then
+        EXPERIENCE_COUNTS='{"total":0}'
+    fi
 
     local REMEDIATION
     REMEDIATION="$(agentis remediation history --limit 5 --json 2>/dev/null || echo '[]')"
--- a/tools/federation-dashboard.sh
+++ b/tools/federation-dashboard.sh
@@ -102,27 +102,78 @@ generate() {
     local DAEMONS
     DAEMONS="$(agentis daemon list --json 2>/dev/null || echo '[]')"
 
-    # Parse knowledge count per colony
+    # Count experience entries per colony.
+    #
+    # #111: the dashboard used to call `agentis knowledge list`, but agents
+    # write with `learn(...)` which persists to `.agentis/experience/`, not
+    # the knowledge base. Two parallel stores, two parallel CLIs — the
+    # knowledge counter was always 0 on a real federation. We now count
+    # JSONL line entries in the experience dir and aggregate per colony via
+    # the agent->colony map. The output JSON shape is still `{total, <col>}`
+    # for compatibility with history.json entries and the HTML chart.
+    #
+    # Experience files are named by daemon-assigned `agent_id` (opaque hex,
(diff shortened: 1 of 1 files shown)


ylohnitram (commented):
## PR #114 Review

### Verdict: REQUEST CHANGES

Functionally this fixes the 0-counter symptom (the main goal), and the atomic-write / PID-scoped temp pattern is preserved. But there is one real correctness bug in the agent-id → colony mapping and a second issue that becomes a silent broken-dashboard on any edge case, both worth fixing before merge. Several smaller notes on error-handling corners.

| # | File | Line | Issue |
|---|------|------|-------|
| 1 | `tools/federation-dashboard.sh` | 146, 155-158 | **Basename-based colony mapping is lossy by construction.** `AGENT_COLONY_MAP` is built by globbing `"$FED_DIR"/*/agents/*.ag` and keying by `basename .ag`, then `name_to_colony = {e.get("agent"): e.get("colony") for e in agent_map}` is a dict comprehension — later colonies overwrite earlier on any basename collision. Today no two `.ag` files share a basename in the dev federation, but the moment someone adds e.g. a second `risk_assessor.ag` under `triage/`, counts will silently attribute to whichever colony appears later in directory-glob order. `agentis daemon list --json` already emits a `colony` field directly from `daemon::read_colony` (src/cli/daemon.rs:1460-1471, populated by the `--colony` flag each `start-colony.sh` passes). Use `d.get("colony")` directly: it eliminates the collision bug, drops two of the four argv blobs passed to python (`$AGENT_COLONY_MAP`, `$COLONY_LIST_PY` still needed for the zero-initialiser), and is consistent with how the JS side already reads `d.colony` at line 586. |
| 2 | `tools/federation-dashboard.sh` | 501 | **Empty `$EXPERIENCE_COUNTS` silently breaks the dashboard script, not just the counter.** The fallback `\|\| echo '{"total":0}'` only fires when `python3` itself exits non-zero. If python succeeds but prints nothing (e.g. the final `print(json.dumps(counts))` is ever removed or short-circuited, or a future edit adds an early-exit branch), you get `const experienceCounts = ;` injected into the HTML — a JS syntax error that kills every IIFE below it, not just the counter. The history-append path is safe because `_safe_json` catches it, but the DATAEOF injection is not. Two ways to harden: (a) wrap the whole python invocation in `EXPERIENCE_COUNTS="$(... )"; [ -z "$EXPERIENCE_COUNTS" ] && EXPERIENCE_COUNTS='{"total":0}'`, or (b) validate `$EXPERIENCE_COUNTS` with `python3 -c 'import json,sys; json.loads(sys.argv[1])' "$EXPERIENCE_COUNTS"` before injection and fall back on failure. Same fragility applies to `$DAEMONS`, `$REMEDIATION`, `$SUGGESTIONS`, `$HISTORY` — these already exist, but #111's counter is now the 5th blob to treat this way and is a good moment to standardize. |

| 1 | `tools/federation-dashboard.sh` | 125-128 | `$EXPERIENCE_DIR` fallback to `.agentis/experience` (cwd-relative) is effectively dead once the dashboard runs at all — the primary `$FED_DIR/../.agentis/experience` path always exists under the install.sh layout. If the primary path is missing, the fallback will almost certainly also be wrong (we're not in the federation root). Consider logging a single warning instead of silently falling through, or drop the fallback; the current form masks misconfiguration. |
| 2 | `tools/federation-dashboard.sh` | 150-173 | Iteration is driven by `daemons` list, which only contains **currently-running** agents. If an agent has `learn()`-written JSONL but is currently stopped, its experience entries will not be counted. Arguably correct for a live dashboard, but worth a one-line comment in the code so the next maintainer doesn't read this as "count all experience files" — especially since the stat label is "Experience Entries" with no "running" qualifier. Alternative: walk `exp_dir` directly and look up colony from the JSONL's first record (`agent_id` is stored inside the JSON), which is resilient to transient stops. |
| 3 | `tools/federation-dashboard.sh` | 231-258 | Pre-existing but newly relevant with concurrent `/refresh` + 60 s loop: the history-append block is a read-modify-write on `$HISTORY_FILE`. Two concurrent `generate()` runs (60 s timer fires during a manual `/refresh`) will each load the same `history`, each append their own entry, and whichever `os.replace()`s last wins — one tick's data point is silently lost. PID-scoped tmp files protect against torn reads, not against lost updates. An `fcntl.flock(f, LOCK_EX)` around the load-append-write window would close it. Not a regression of this PR; flagging because the PR explicitly reasons about concurrency at line 96-99. |
| 4 | `tools/federation-dashboard.sh` | 679-688 | `kMax` uses all history entries, but entries pre-#111 lack the `experience` key and are skipped — yet they still influence `tMin` / `tMax` (line 677) because the filter is only on the polyline. On an upgraded federation with 3 days of old history, the x-axis will be stretched across the old + new range and the new (only-useful) data will be crammed into the rightmost slice. Consider filtering `history` to `h.experience != null` once at the top of the IIFE and using the filtered array for tMin/tMax too. Comment at 669-672 acknowledges "drop them entirely" — this is the other half of that drop. |
| 5 | `tools/federation-dashboard.sh` | 156-157 | `if name.endswith(".ag"): name = name[:-3]` strips 3 chars, not 4. `.ag` is 3 characters, so this is correct, but a drive-by reader will flinch at it. `os.path.splitext(name)[0]` is clearer and matches how the install.sh agent-listing scripts strip extensions. Style-level, low priority — tied to #1 above since if you switch to `d["colony"]` this code goes away entirely. |

### Positive Notes
- Root cause diagnosis is correct: `learn()` writes `.agentis/experience/<agent_id>.jsonl`, not the knowledge base, and the old CLI was looking at the wrong store.
- Comment block at lines 105-118 is excellent — explains both the bug and why the `source → basename` detour is needed (modulo the #1 note above that it is no longer needed given `d["colony"]`).
- JSON output shape `{total, <colony>}` preserved, so history.json from pre-#111 runs doesn't break the chart (it's just skipped).
- Stat label ("Knowledge Entries" → "Experience Entries"), history key (`knowledge` → `experience`), chart heading ("Knowledge Growth" → "Experience Growth"), and all JS identifiers updated consistently. No stragglers found.
- Defensive try/except around every `json.loads` in the new python block + `_safe_json` in the existing block — covers the fresh-federation empty-input case.
- Per-PID temp files + `mv` / `os.replace` for atomic publish is preserved from the existing pattern.

ylohnitram:
## Re-review of 3c8c032

### Verdict: both fixes are adequate — ready to merge

**Fix 1 (authoritative `colony` field)** — verified against `src/cli/daemon.rs:1460-1472`: `daemon list --json` emits `colony` as either a `String` (when daemon was started with `--colony`) or `Null`. The new code correctly treats null/empty as "fall back to basename map", preserving the previous behavior for bare `agentis daemon <file>` invocations. The docstring explicitly flags the collision limitation of the basename fallback, which is the right level of honesty.

**Fix 2 (JSON validation guard)** — I reproduced the failure modes locally:
- Empty stdin → python exits 1, `if !` branch fires, substitutes `{"total":0}`. Correct.
- Garbage/partial output → `json.loads` raises, same path. Correct.
- Valid JSON → python exits 0, `if !` is false, original value retained. Correct.
- `set -e` (line 17) is not tripped because the failure is inside an `if` condition. Verified.

The substituted `{"total":0}` is slightly narrower than what the python block emits on an empty experience dir (which also includes per-colony zero entries), but the HTML chart handles `undefined` per-colony values via `|| 0` on line 713, so no downstream break.

### Smaller notes — no second pass needed

- **`.agentis/experience` cwd fallback (line 136)** — not dead. It matches the LOG_DIR convention on line 187 and handles operators running the dashboard from the federation root. Fine as-is; the new comment on 130-133 explains the intent.
- **`kMax` / `tMin` naming (lines 702-704)** — pure style; the variables correctly track max experience count and min timestamp. No bug.
- **`running` stat only counts running agents (line 589)** — this is what the label "Agents Running" says. Working as intended.

### Positive notes

- The fallback comment (lines 155-157) documents the "last-writer-wins on basename collision" limitation at the exact spot the collision can happen. Future maintainers won't have to re-derive it.
- The python-validation guard's comment (192-198) names the specific failure modes (SIGKILL, disk full, bad shebang) rather than hand-waving at "if python fails". Good surface area for future debugging.
- The commit message cites `src/cli/daemon.rs:1460-1471` as the evidence for the `colony` field contract. This is the discipline I want to see on PRs that cross the core/colonies repo boundary.

Safe to merge once the PR test plan checkboxes in the description are run through by the operator.

Issue #117, opened 2026-04-14
dev-apprenticeship: gitlab-api.sh cannot distinguish rate-limit / auth / server errors, logs nothing actionable, no backoff


ylohnitram:
## Problem

\`scripts/gitlab-api.sh\` uses \`curl -sfS\` for every GitLab call. That's convenient but it throws away information the federation needs. Five specific gaps:

### 1. No HTTP status visible

\`curl -sfS\` surfaces only an exit code (22 for 4xx, 22 for 5xx, etc.). The script never captures \`%{http_code}\` and never logs the response body. An operator tailing logs sees:

\`\`\`
[labeler] GitLab poll failed: Error { message: ExecFailed: exec sh failed (exit 22): ... }
\`\`\`

and has **no idea** whether that's 401 (token invalid), 403 (insufficient scope), 404 (project path wrong), 429 (rate limit), 500 (GitLab having a moment), or 503 (maintenance). Each of those needs a different operator response. Right now they're indistinguishable.

### 2. Rate-limit (429) treated the same as permanent failure

GitLab's default rate limit on self-hosted is 300 req/min per user/IP. With 21 agents polling every 60s plus write traffic when at 0.6+, the federation can plausibly hit this during bursts (especially if \`install.sh\` is misconfigured and multiple federations share a PAT — or if \`start-federation.sh\` got run twice, see #116).

When 429 happens:
- There is no \`Retry-After\` header parsing.
- There is no exponential backoff.
- There is no distinction from a transient 5xx. The agent just tries again 60s later and probably gets 429 again, piling up more rate-limit consumption.

### 3. Auth failures (401/403) loop silently forever

If the operator's PAT expires or gets its scopes changed, every GitLab call returns 401. The agent logs \`poll failed\` every tick indefinitely. There's no \"your token is invalid, please rotate\" surface anywhere — the same line scrolls by as when the network is briefly down.

Compare with what a friendly message would look like (made-up):

\`\`\`
[labeler] GitLab call failed: HTTP 401 Unauthorized
  endpoint: GET /projects/X%2Ffoo/issues
  response: {\"message\": \"401 Unauthorized\"}
  Hint: token may be expired or missing required scope (needs 'api')
\`\`\`

Operator knows exactly what to do. Current state: they suspect the federation, not their token.

### 4. Empty responses aren't logged at all

[...]

PR #120, opened 2026-04-14
fix(dev-apprenticeship): resilient gl_call wrapper (#115, #117)

### LINKED ISSUE #115: dev-apprenticeship: agents that were learning get stopped by watchdog after a single curl timeout (ExecFailed = 'permanent' + no retry)
## Problem

After running the federation for ~90 minutes post-#107-fix, **10 of 21 agents have transitioned from \`running / observing successfully\` to \`stopped\`**. On inspection, each of them was actively learning — their \`experience/<id>.jsonl\` file has real entries — and then died from a single network-level failure.

## The sequence (task_decomposer, 7 real observations before death)

\`\`\`
1776156836278 [task_decomposer] Observed 2 epics, avg 4 subtasks        ← healthy
1776156836278 [task_decomposer] Observed 2 epics, avg 5 subtasks        ← healthy
...
1776157147501 [task_decomposer] GitLab unplanned-issue poll failed:
  Error { code: 2010, category: permanent,
          message: ExecFailed: exec sh failed (exit 28):
                   curl: (28) Connection timed out after 30002 milliseconds }
...
(agent eventually stops running per agentis daemon list)
\`\`\`

The same pattern repeats across \`scope_estimator\`, \`risk_assessor\`, and others.

### Bug 1: curl timeout too short for real payloads

\`scripts/gitlab-api.sh\` hardcodes \`curl -sfS --max-time 30\` (4 call sites). For projects with >50 issues or MRs, a single \`GET /projects/.../issues?per_page=100\` response is 70–95 KB and routinely exceeds 30 seconds under normal self-hosted GitLab load, especially when the endpoint does sorting / counts / label joins server-side.

My runs show timeouts like:
\`\`\`
curl: (28) Operation timed out after 30006 milliseconds with 7041 out of 70999 bytes received
curl: (28) Operation timed out after 30005 milliseconds with 35713 out of 70999 bytes received
curl: (28) Operation timed out after 30005 milliseconds with 72456 out of 95910 bytes received
\`\`\`

The data is *coming in* — it just hasn't finished in 30s. Bumping \`--max-time 30\` to \`--max-time 90\` (or reading from an env var that \`start-colony.sh\` sets to something pegged to agentis's \`exec.default_timeout_ms\`) would eliminate most of the spurious failures.

### Bug 2: curl timeouts are classified as \`permanent\`, killing the agent

More concerning: the agentis runtime is reporting \`category: permanent\` for these errors:

\`\`\`
Error { code: 2010, category: permanent,
        message: ExecFailed: exec sh failed (exit 28): curl: (28) Operation timed out ... }
\`\`\`

A curl network timeout is the canonical **transient** failure — the remote is overloaded, a packet was lost, DNS hiccupped. It should retry. But \`category: permanent\` seems to trigger the watchdog to stop the agent permanently, not just skip the tick.

Result: **a single bad-weather moment on the network permanently removes an agent from the federation until a human operator does \`./start-federation.sh\` again**. That's catastrophic for a system that's supposed to run unattended for weeks-months while the apprenticeship accumulates.

The classification either needs to be fixed at the agentis-core layer (exec exit 28 is transient, full stop) or at the dev-apprenticeship layer by wrapping \`gitlab-api.sh\` calls so curl failures are converted to non-fatal \`.ag\` errors that the agent can handle without dying.

## Suggested fixes

**dev-apprenticeship-side (quick, this repo):**

1. Bump \`--max-time 30\` to \`--max-time 90\` across \`scripts/gitlab-api.sh\` in all 5 colonies (or make it configurable via \`\$GITLAB_CURL_MAX_TIME\` with a 90s default).
2. Reduce \`per_page=100\` to \`per_page=50\` on the hot endpoints (\`issues\`, \`merge_requests\`, \`commits\`) to keep response sizes — and response times — predictable.
3. Consider wrapping \`curl\` calls in a small retry loop with exponential backoff (\`for i in 1 2 3; do curl ... && break; sleep \$((i*5)); done\`) so transient timeouts don't surface to the \`.ag\` layer at all.

**agentis-core-side (real fix, follow-up):**

Exec exit code 28 (curl "timed out") should always be classified \`transient\`, never \`permanent\`. The current classification causes catastrophic agent loss on what are recoverable network blips.

## Evidence this is catastrophic

After 90 minutes of running on a reasonably healthy self-hosted GitLab:

- **10/21 agents stopped** (48% of the federation, dead until manual restart)
- Stopped agents include the only ones that were learning (\`task_decomposer\`, \`scope_estimator\`, \`risk_assessor\` — 7+14+16 KB of accumulated experience, now frozen)
- The remaining \`running\` agents are either the ones that were already broken by #109/#110 (never got to poll, so never hit this) or the ones that got lucky on their polling windows

In other words: the working agents get killed by bad networks. The broken agents keep \"running\" because they error out before they try to call GitLab. That's backwards.

## Repro

1. Run dev-apprenticeship against a self-hosted GitLab with moderate load and 100+ issues.
2. Wait for \`labeler\`/\`task_decomposer\`/\`scope_estimator\` to start completing observation ticks.
3. Check \`agentis daemon list\` every 15–30 minutes.
4. Observe the count of \`stopped\` agents climbing monotonically.

## Environment

- agentis: v1.1.8
- \`exec.default_timeout_ms = 45000\` (set by v1.1.8's install.sh fix for #107)

## Related

- #107 (PII guard + exec timeout — fixed, but its curl-timeout sibling wasn't addressed)
- #109 (type error blocking 9 agents)
- #110 (adaptive engine blocking router/prioritizer)
- #111 (dashboard reads wrong storage)

Together, #109/#110/this/#111 mean the current master of dev-apprenticeship + v1.1.8 can't sustain more than a handful of learning agents for longer than about an hour against a real GitLab.

ylohnitram:
## Summary

Fixes #115 (curl timeout kills agent) and #117 (no HTTP status visibility / no backoff / no auth hint) in all 5 colony `gitlab-api.sh` scripts (triage, code-review, planning, implementation, release) via a single `gl_call` wrapper.

## What changed

- **Single `gl_call <method> <url> [curl-args…]`** shared by `gl_get`, `gl_get_q`, `gl_post`, `gl_put` (names preserved, so case statements downstream don't change).
- **Captures `%{http_code}` + response body** instead of letting `curl -f` swallow non-2xx with an opaque exit code.
- **Structured exit codes** so agents can reason about failures:
  - `0` — 2xx OK
  - `2` — 401/403 auth (no retry, stderr hint: "check PAT scope/expiry")
  - `3` — 429 rate limit (after retry budget exhausted)
  - `4` — other 4xx (permanent)
  - `5` — 5xx or transport failure (timeout/DNS/TLS — retried)
- **Exponential backoff** (3s → 6s → 12s, budget via `$GITLAB_CURL_RETRIES`, default 3 attempts-after-first) on 5xx / 429 / transport.
- **`--max-time` bumped from hardcoded 30 → `$GITLAB_CURL_MAX_TIME` default 90**. The 30s floor was below real `/issues?per_page=100` response times on self-hosted GitLab, which was the proximate cause of #115.
- **`RETURN` trap** cleans up the body tmpfile on every return path.

## Why the structured exit codes matter

The agent `.ag` layer can now tell the watchdog "give up" vs "try again later". A 401 from a rotated PAT surfaces as an actionable message on stderr instead of an indistinguishable `ExecFailed exit 22` repeating every tick forever. A 429 doesn't pile on more rate-limit consumption — the wrapper backs off inside a single tick. A network timeout (curl exit 28) is retried three times before it ever escapes to the evaluator, so the watchdog won't kill an otherwise-healthy agent on one bad moment.

## Note on #115 bug 2 (core side)

The core-side bug — `EvalError::ExecFailed` classified `"permanent"` regardless of exit code when exit-28 is canonically transient — is still open in `agentis-core` and will be fixed in a separate PR there. This PR is sufficient on its own because the retry loop eliminates the trigger: curl timeouts now happen inside `gl_call` and are absorbed before they reach the evaluator.

## Test plan

- [x] `bash -n` all 5 scripts (done locally — all parse clean).
- [ ] Start the federation against a real GitLab and confirm agents recover from a transient 5xx (can be simulated by pausing GitLab or blackholing briefly).
- [ ] Rotate / revoke the PAT and confirm the stderr message now reads `auth failure (HTTP 401) on … — check PAT scope/expiry` rather than a generic `exec sh failed (exit 22)`.
- [ ] Let the federation run for >30min and confirm no agent transitions `running → stopped` from a single curl timeout.

### DIFF
--- a/dev-apprenticeship/code-review/scripts/gitlab-api.sh
+++ b/dev-apprenticeship/code-review/scripts/gitlab-api.sh
@@ -35,10 +35,119 @@ fi
 
 API="$GITLAB_URL/api/v4/projects/$GITLAB_PROJECT"
 
+# gl_call <method> <url> [curl-args...]
+#
+# Single wrapper used by every gl_get/gl_get_q/gl_post/gl_put below.
+# Captures HTTP status + response body, distinguishes auth / rate-limit /
+# client / server / transport errors, and retries transient ones with
+# exponential backoff so the .ag layer doesn't see them.
+#
+# Exit codes (so callers & agents can reason about failures):
+#   0  — 2xx OK (response body on stdout)
+#   2  — 401/403 auth failure (actionable error on stderr, no retry)
+#   3  — 429 rate limited (after retries exhausted)
+#   4  — other 4xx client error (permanent — wrong URL, bad body, etc.)
+#   5  — 5xx server error OR transport failure (timeout, DNS, connect)
+#
+# Env knobs:
+#   GITLAB_CURL_MAX_TIME  per-attempt --max-time seconds (default 90)
+#                         #115: was hardcoded 30, too short for real
+#                         projects where /issues?per_page=100 exceeds 30s
+#                         routinely on self-hosted GitLab under load.
+#   GITLAB_CURL_RETRIES   retry budget for 5xx/timeouts/429 (default 3).
+#                         Budget is attempts-after-first, so 3 means
+#                         4 total attempts before giving up.
+gl_call() {
+    local method="$1" url="$2"
+    shift 2
+    local max_time="${GITLAB_CURL_MAX_TIME:-90}"
+    local max_retries="${GITLAB_CURL_RETRIES:-3}"
+    local attempt=0 delay=3 rc code snippet body_file
+    body_file="$(mktemp)"
+    # bash RETURN trap fires on function exit regardless of return path,
+    # so the tmp body file is cleaned up even when we return non-zero
+    # from inside the while loop.
+    trap 'rm -f "$body_file"' RETURN
+
+    while :; do
+        attempt=$((attempt + 1))
+        code=""
+        rc=0
+        # -sS: silent progress, keep errors.  No -f: we *want* to inspect
+        # non-2xx responses ourselves instead of losing them to curl's
+        # "HTTP error" exit path. -w writes the status to stdout, which
+        # we capture; the body goes to $body_file via -o.
+        code="$(curl -sS -w '%{http_code}' -o "$body_file" \
+            --max-time "$max_time" \
+            -X "$method" \
+            -H "PRIVATE-TOKEN: $GITLAB_TOKEN" \
+            "$@" \
+            "$url" 2>/dev/null)" || rc=$?
+
+        # Transport failure (exit 28 timeout, 6 DNS, 7 connect, 35 TLS, …).
+        # curl didn't get a status line back, so $code is empty.
+        if [ "$rc" -ne 0 ] || [ -z "$code" ]; then
+            if [ "$attempt" -le "$max_retries" ]; then
+                sleep "$delay"
+                delay=$((delay * 2))
+                continue
+            fi
+            emit_error "curl transport failure (exit $rc) after $attempt attempts: $method $url"
+            return 5
+        fi
+
+        case "$code" in
+            2*)
+                cat "$body_file"
+                return 0
+                ;;
+            401|403)
+                # No retry — a fresh attempt with the same token will
+                # hit the same wall. Surface an actionable hint so the
+                # operator knows to rotate / re-scope their PAT instead
+                # of blaming the federation.
+                snippet="$(head -c 200 "$body_file")"
+                emit_error "auth failure (HTTP $code) on $method $url: $snippet — check PAT scope/expiry"
+                return 2
+                ;;
+            429)
+                # GitLab sometimes sends Retry-After, but parsing headers
+                # portably across curl versions is fiddly; conservative
+                # exponential backoff matches what a Retry-After would
+                # typically request on a self-hosted instance.
+                if [ "$attempt" -le "$max_retries" ]; then
+                    sleep "$delay"
+                    delay=$((delay * 2))
+                    continue
+                fi
+                emit_error "rate limited (HTTP 429) after $attempt attempts on $method $url"
+                return 3
+                ;;
+            5*)
+                if [ "$attempt" -le "$max_retries" ]; then
+                    sleep "$delay"
+                    delay=$((delay * 2))
+                    continue
+                fi
+                snippet="$(head -c 200 "$body_file")"
+                emit_error "server error (HTTP $code) after $attempt attempts: $snippet"
+                return 5
+                ;;
+            4*)
+                snippet="$(head -c 200 "$body_file")"
+                emit_error "client error (HTTP $code) on $method $url: $snippet"
+                return 4
+                ;;
+            *)
+                emit_error "unexpected HTTP $code on $method $url"
+                return 4
+                ;;
+        esac
+    done
+}
+
 gl_get() {
-    curl -sfS --max-time 30 \
-        -H "PRIVATE-TOKEN: $GITLAB_TOKEN" \
-        "$1"
+    gl_call GET "$1"
 }
 
 # gl_get_q <url> [--data-urlencode k=v ...]
@@ -48,19 +157,11 @@ gl_get() {
 gl_get_q() {
     local url="$1"
     shift
-    curl -sfS -G --max-time 30 \
-        -H "PRIVATE-TOKEN: $GITLAB_TOKEN" \
-        "$@" \
-        "$url"
+    gl_call GET "$url" -G "$@"
 }
 
 gl_post() {
-    curl -sfS --max-time 30 \
-        -X POST \
-        -H "PRIVATE-TOKEN: $GITLAB_TOKEN" \
-        -H "Content-Type: application/json" \
-        -d "$2" \
-        "$1"
+    gl_call POST "$1" -H "Content-Type: application/json" -d "$2"
 }
 
 CMD="${1:?Usage: gitlab-api.sh <command> [args...]}"
--- a/dev-apprenticeship/code-review/scripts/gitlab-api.sh
+++ b/dev-apprenticeship/code-review/scripts/gitlab-api.sh
@@ -66,8 +66,12 @@ gl_call() {
     body_file="$(mktemp)"
     # bash RETURN trap fires on function exit regardless of return path,
     # so the tmp body file is cleaned up even when we return non-zero
-    # from inside the while loop.
-    trap 'rm -f "$body_file"' RETURN
+    # from inside the while loop. The trap unregisters itself in its own
+    # body — a bare `trap 'rm -f ...' RETURN` in bash is shell-global,
+    # not function-local, so without `trap - RETURN` the cleanup would
+    # re-fire on every subsequent function return in the script with a
+    # stale `$body_file` (out of scope, so rm's target is '').
+    trap 'rm -f "$body_file"; trap - RETURN' RETURN
 
     while :; do
         attempt=$((attempt + 1))
--- a/dev-apprenticeship/implementation/scripts/gitlab-api.sh
+++ b/dev-apprenticeship/implementation/scripts/gitlab-api.sh
@@ -43,10 +43,119 @@ fi
 
 API="$GITLAB_URL/api/v4/projects/$GITLAB_PROJECT"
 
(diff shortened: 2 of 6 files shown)


ylohnitram (commented):
## PR #120 review — verdict: **comment / approve-with-nits**

The core of this PR is solid. `gl_call`'s retry counter is correct (initial + 3 retries = 4 total), `case` ordering puts `401|403` / `429` before `4*` so the specific patterns win, the backoff pattern does not add a trailing sleep, `|| rc=$?` cleanly absorbs curl's exit under `set -e`, the shell-quoting through two `"$@"` hops survives intact (verified `--data-urlencode "state=opened"` reaches curl as one argv element), and the `gl_call` / `gl_get` / `gl_get_q` / `gl_post` function bodies are byte-identical across all 5 files. No .ag agent pattern-matches on specific exit codes (they all just `try { exec sh cmd; } catch e { ... }`), so the switch from `curl -sfS` exit 22 → structured 0/2/3/4/5 is a no-op at the .ag layer.

The issues below are not blockers on their own, but at least #1 and #2 are worth a small follow-up commit on this branch before merging.

### 1. `exec.default_timeout_ms = 45000` preempts the new `--max-time 90` (integration)

`install.sh:222` still writes `exec.default_timeout_ms = 45000`, justified by a comment that says:

> gitlab-api.sh calls curl with `--max-time 30`. [...] 45 s gives curl its 30 s budget plus 15 s of process spawn/JSON-parse headroom.

That comment (and that 45000 value) was sized for the old 30-second curl. With the new `GITLAB_CURL_MAX_TIME=90` default, agentis will SIGKILL the `exec sh` child at 45 s — before curl's own 90 s budget even expires. Net effect on a genuinely slow endpoint (the exact case #115 was filed against):

- Attempt 1 runs → agentis kills at 45 s → `ExecFailed` → caught by `try { } catch`, agent gives up for this tick.
- The retry loop inside `gl_call` is never reached on slow responses, because the 90 s ceiling is unreachable.

Retries still work fine for fast-failing 5xx/429 (response in <1 s), but the main `#115` case — a slow `/issues?per_page=N` on self-hosted GitLab — still dies on the first attempt. Two ways to reconcile:

- Option A: drop `GITLAB_CURL_MAX_TIME` default to e.g. 40 s so it fits under 45 s, and let retries take the slack.
- Option B: bump `exec.default_timeout_ms` in `install.sh` to match worst-case `gl_call` budget (4 × 90 s curl + 21 s of backoffs ≈ 380 s), plus a healthy margin.

Either way, **please update the `install.sh:221-227` comment** — it currently documents a 30 s `--max-time` that no longer exists and will mislead the next maintainer.

### 2. `RETURN` trap leaks across function frames

Bash RETURN traps set inside a function are global unless `set -T`/`functrace` is on. Verified behavior: the `trap 'rm -f "$body_file"' RETURN` set on line 70 fires **twice** on every `gl_get_q` call — once as `gl_call` returns (correct, tmpfile cleanup works), and again as `gl_get_q` returns (at which point `$body_file` is out of scope and expands to `""`, so `rm -f ""` is a no-op). More importantly, the trap is never unregistered — it stays registered globally and fires for every subsequent function return in the script (every future `emit_error`, every future `gl_*` call's outer wrapper, etc.). It does not leak the actual tmpfile — the first firing always sees the live `$body_file` — but it is sloppy.

Two harmless fixes, take either:

```sh
# At the very end of gl_call (right before the closing `}`):
trap - RETURN
```

or move the cleanup to an EXIT trap inside a subshell, or do explicit `rm -f "$body_file"` before each `return` and drop the trap entirely.

### 3. The five `gl_call` bodies are identical, but they are copies (maintenance)

Confirmed by `sha256sum` of each extracted `gl_call` function body — all five match exactly. Good today; any future change to retry logic, curl flags, or HTTP dispatch needs to be touched in 5 places. Consider a follow-up PR that sources a shared `lib/gitlab-api-common.sh` (same pattern as `tools/parse-toml.sh` which is already used by `start-colony.sh`). Not in scope for this PR.

### 4. `GITLAB_CURL_*` env-var passthrough is already covered — verified

`install.sh:288` writes `exec.env_passthrough = COLONY_DIR,GITLAB_*`, and the trailing `*` glob matches `GITLAB_CURL_MAX_TIME` / `GITLAB_CURL_RETRIES` exactly. Operators can override these by exporting them before `./start-federation.sh`. No additional wiring needed in `start-colony.sh` / `colony.toml` / `install.sh` on that axis. If you *want* these to be operator-tunable via `colony.toml` rather than the env, that is a separate feature; silent default is fine.

### 5. Minor: `curl ... 2>/dev/null` drops transport error detail

On transport failure (`rc != 0`), the `emit_error` message is `"curl transport failure (exit $rc) after $attempt attempts: $method $url"` — useful, but the original curl stderr ("Could not resolve host", "SSL certificate problem: …", "Connection refused") is silently discarded by `2>/dev/null`. Consider capturing stderr to a second tmpfile and including a snippet in the final `emit_error` when retries exhaust, parallel to how the 4xx/5xx branches already `head -c 200 "$body_file"`. Nice-to-have, not blocking.

### 6. Backward-compatibility sanity — confirmed

Grepped all 22 `.ag` agent files for `exit_code` / `exit 22` / `exit 28` / `ExecFailed` / specific status handling. **No agent pattern-matches on any specific exit code.** Every `exec sh` call to `gitlab-api.sh` is wrapped in `try { } catch e { ... }` and treats failure as a generic "skip this tick" signal. The switch from curl-exit-22 to wrapper-exit-4 is invisible to the .ag layer. The new actionable stderr messages (`auth failure (HTTP 401) on ...`) will land in the caught `e` value via the `ExecFailed.stderr` field and surface through the agents' `print("[agent] failed:", e)` calls.

### Positive notes

- Drop of `-f` is the right call — inspecting `%{http_code}` ourselves is strictly more informative than letting curl swallow the response.
- `401|403 → return 2 (no retry)` is correct; retrying with the same PAT can't succeed, and the actionable "check PAT scope/expiry" hint is exactly what operators need to un-stuck a rotated-token federation.
- Decoupling the 429/5xx backoff from the evaluator's retry/watchdog logic is the right layer — the .ag layer is the wrong place to know about HTTP semantics.
- `GITLAB_CURL_RETRIES` default 3 gives ~21 s of accumulated backoff before surfacing, well-matched to transient blips.
- JSON body construction via `python3 json.dumps` in each call site is unchanged and remains the correct escape strategy.

ylohnitram:
Review feedback addressed in commit a147120:

1. **`exec.default_timeout_ms` mismatch** — bumped from 45 000 → 120 000 in `install.sh`. A single `GITLAB_CURL_MAX_TIME=90` attempt now fits the exec-sh wall-timer with 30 s headroom for spawn / parse / first retry. Updated the stale comment block that still referenced `--max-time 30`. Note: `write_key` is idempotent, so operators who already ran install.sh keep their 45 000 value until they either re-run a fresh install or edit `.agentis/config` by hand.

2. **RETURN trap shell-global leak** — changed `trap 'rm -f "$body_file"' RETURN` → `trap 'rm -f "$body_file"; trap - RETURN' RETURN` so the trap unregisters itself after firing. Verified with a repro harness:

```
probe: body_file=/tmp/tmp.Xeh4tHTT5s exists=y
after-probe: trap RETURN is: []
other: returning
after-other: trap RETURN is: []
probe: body_file=/tmp/tmp.we9KExT1Ow exists=y
after-probe2: trap RETURN is: []
```

Trap is correctly cleared after each `gl_call` return and correctly re-armed on the next. No cross-function contamination.

Applied identically to all 5 scripts; `sha256sum` of the function body is still uniform across triage/code-review/planning/implementation/release.

**Not addressed (intentional)**:
- #3 copy-paste across 5 files → deferred. Extracting to a shared sourced library is the right follow-up but outside scope for #115/#117. Filing a tracking issue separately.
- #4 `curl ... 2>/dev/null` dropping transport-level diagnostics → nice-to-have; can be addressed in the same follow-up that extracts the shared library.

PR #124, opened 2026-04-14
fix(tools): make colony-lint.sh runnable on bash 3.2 / macOS (#121)

### LINKED ISSUE #121: tools/colony-lint.sh: uses bash 4+ line-continuation syntax but shebang is /bin/bash — fails on macOS default bash 3.2
## Problem

\`tools/colony-lint.sh\` is the project's main pre-merge guardrail (per CLAUDE.md: *\"Colony lint must pass with 0 failures before merge. Current baseline: 45 passed, 5 skipped.\"*). Running it on macOS with the stock \`/bin/bash\`:

\`\`\`
$ ./tools/colony-lint.sh
./tools/colony-lint.sh: line 188: syntax error near unexpected token \`;;'
\`\`\`

The tool aborts immediately. Zero checks run. Zero failures reported (which the operator might misread as \"clean\").

## Root cause

Line 184–190 uses line-continuation (\`\\\`) inside \`case\` labels:

\`\`\`bash
case \"\$flag\" in
    --tick-interval|--cb-per-tick|--colony|--deadline|--priority|\\
    --enable-migration|--enable-replication|--allow-replica-replication|\\
    --enable-exec|--enable-messaging|--deny-exec|\\
    --help|-h) ;;
    *) echo \"\$flag\" ;;
esac
\`\`\`

This form is only parsed correctly by **bash 4+**. Apple ships bash **3.2.57** in \`/bin/bash\` (they never upgraded past 3.2 because of GPLv3), which is what \`#!/bin/bash\` at the top of \`colony-lint.sh\` resolves to on macOS.

\`\`\`
$ /bin/bash --version
GNU bash, version 3.2.57(1)-release (arm64-apple-darwin25)
\`\`\`

Linux systems ship bash 4 or 5 as \`/bin/bash\`, so the tool works there. On macOS dev machines (which is what most contributors will be on given the Agentis / X team makeup) it silently doesn't run.

### Option A — change shebang to \`/usr/bin/env bash\` and require modern bash

Change \`#!/bin/bash\` → \`#!/usr/bin/env bash\` and document the requirement. Operators on macOS can \`brew install bash\` to get a modern one in \`/usr/local/bin/bash\` (or \`/opt/homebrew/bin/bash\` on Apple Silicon), and \`env bash\` finds whichever is first in PATH.

This is the standard cross-platform fix for this class of problem.

### Option B — rewrite the case label to bash 3.2-compatible syntax

Remove the line-continuation inside the case labels. One long line, or split across multiple \`case\`/\`esac\` blocks:

\`\`\`bash
case \"\$flag\" in
    --tick-interval|--cb-per-tick|--colony|--deadline|--priority|--enable-migration|--enable-replication|--allow-replica-replication|--enable-exec|--enable-messaging|--deny-exec|--help|-h) ;;
    *) echo \"\$flag\" ;;
esac
\`\`\`

Ugly but portable. Preferred if the project wants to keep supporting stock macOS bash.

### Also: add a bash version check at the top

\`\`\`bash
if [ \"\${BASH_VERSINFO[0]:-0}\" -lt 4 ]; then
    echo \"colony-lint.sh requires bash >= 4 (you have \${BASH_VERSION}). On macOS: brew install bash and use /usr/local/bin/bash (or /opt/homebrew/bin/bash).\" >&2
    exit 2
fi
\`\`\`

Fails loudly instead of silently breaking on line 188.

## Impact

- Any macOS contributor who adds an \`.ag\` file or a new colony cannot pre-validate locally. They rely on CI (if any) to catch issues.
- CLAUDE.md claims this tool is a quality gate; that gate is currently open for half the dev community.
- On Linux contributors, nothing to notice → blind spot.

## Environment

- \`/bin/bash\` version: GNU bash 3.2.57 (macOS default)

## Related

- #116 (operational readiness) — same flavor of \"the happy-path install works for the author, breaks for anyone else\". Worth running all project shell tooling through \`shellcheck\` and a \`bash --posix\` smoke test in CI.

ylohnitram:
## Summary

- Collapse the daemon-flag-allowlist `case` label to a single line so stock macOS `/bin/bash` (3.2) parses it (the backslash-newline-inside-case-pattern pattern added in #71 made the whole script unparseable on any Mac without brew-bash earlier on PATH).
- Switch shebang to `/usr/bin/env bash` so brew-installed bash 4+ is picked up automatically.
- Document the bash-3.2-compat baseline in the header so future edits avoid `${var^^}`, `mapfile`, associative arrays, and the same backslash-newline trap.

No logic change — the allowlist still enforces exactly the same 11 flags as before.

## Test plan

- [x] `bash -n tools/colony-lint.sh` — clean
- [x] `bash tools/colony-lint.sh` on dev-apprenticeship — 45 passed, 0 failed, 5 skipped (identical to pre-change)
- [x] No other `|\\$` backslash-continuations elsewhere in `tools/`
- [ ] CI shellcheck + test-*.sh unit tests green

Closes #121

### DIFF
--- a/tools/colony-lint.sh
+++ b/tools/colony-lint.sh
@@ -1,10 +1,14 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # colony-lint.sh: validates colony structure, config, and scripts
 #
 # Autodiscovers federations and colonies in the repo.
-# Exit code 0 = all checks pass, 1 = one or more failures.
+# Exit code 0 = all checks pass, 1 = one or more failures, 2 = unmet prerequisites.
 #
 # Usage: ./tools/colony-lint.sh [path-to-repo-root]
+#
+# Runs on bash 3.2+ (stock macOS /bin/bash) and bash 4+. Avoid bash-4-only
+# constructs below (associative arrays, ${var^^}, mapfile, backslash-newline
+# inside case-pattern labels — see #121 for that last one).
 
 set -euo pipefail
 
@@ -181,11 +185,11 @@ if errors:
                     if ($0 !~ /\\[[:space:]]*$/) { in_cmd = 0 }
                 }
             ' "$start_script" | while read -r flag; do
+                # Pattern is deliberately on one line: bash 3.2 (stock macOS)
+                # miscompiles backslash-newline inside case-pattern labels
+                # (#121). Keep all alternatives on a single line.
                 case "$flag" in
-                    --tick-interval|--cb-per-tick|--colony|--deadline|--priority|\
-                    --enable-migration|--enable-replication|--allow-replica-replication|\
-                    --enable-exec|--enable-messaging|--deny-exec|\
-                    --help|-h) ;;
+                    --tick-interval|--cb-per-tick|--colony|--deadline|--priority|--enable-migration|--enable-replication|--allow-replica-replication|--enable-exec|--enable-messaging|--deny-exec|--help|-h) ;;
                     *) echo "$flag" ;;
                 esac
             done)


ylohnitram:
## Review: approve

### Allowlist drift check
Confirmed the collapsed single-line `case` pattern at line 192 enforces exactly the same 13 tokens as the old four-line version — 11 daemon flags + `--help` + `-h`. No drift from the collapse:

```
--tick-interval --cb-per-tick --colony --deadline --priority
--enable-migration --enable-replication --allow-replica-replication
--enable-exec --enable-messaging --deny-exec --help -h
```

### Bash-3.2 compat sweep on the rest of the script
Grepped for the usual bash-4-only offenders — none present:

- No `mapfile` / `readarray`
- No `declare -A` / associative arrays
- No `${var^^}` / `${var,,}` / `${var@U}`
- No `coproc`, `wait -n`, `|&`, `&>>`
- No `[[ =~ ]]`, `globstar`, `shopt`
- No `local -n` / `declare -n` nameref
- No `printf -v` (bash 3.1+ anyway, safe)

`&>/dev/null` is fine on 3.2 (in since 2.05b; it's `|&` and `&>>` that are 4.0+).

Process substitutions `< <(...)`, array `+=`, here-strings `<<<`, and `${#arr[@]}` guards are all bash 3.0+.

### Out-of-scope observations (not blockers for this PR)

1. `grep -oP '\[.*?\]\(\K[^)]+'` on lines 234 and 256 uses PCRE, which macOS BSD grep does not support. The call is `|| true` with stderr silenced, so on macOS the markdown-link checks silently produce an empty link list and `links_ok` stays `true` — the checks pass vacuously. Separate issue from the bash-3.2 parse fix; worth a follow-up.
2. The comment on line 165-166 says the allowlist "must match `DAEMON_FLAGS` in agentis-core/src/cli/daemon.rs". Core's current `DAEMON_FLAGS` omits `--deny-exec` (that one is only in `DAEMON_INNER_FLAGS`). The lint allowlist is strictly more permissive than core here — a start-colony.sh using `--deny-exec` will pass lint but fail at runtime. Pre-existing since #71; out of scope.

Neither of these is introduced or exacerbated by this PR. Ship it.

PR #126, opened 2026-04-14
fix(dev-apprenticeship): downselect GitLab JSON via --view (#119)

### LINKED ISSUE #119: dev-apprenticeship: 21 agents resend ~74 KB of GitLab payload to the LLM every tick — ~2 GB/day through Claude CLI, unacceptable cost/rate profile
## Problem

Every tick, every agent \`curl\`s the full GitLab endpoint it cares about, then passes the raw JSON response into a \`prompt(...)\` call. On my real project run:

\`\`\`
[labeler] input length: 74006 chars  (/issues?per_page=100)
[router]  input length: 69504 chars  (same endpoint, different prompt)
[prioritizer] input length: 69504 chars  (same endpoint, different prompt)
[style_reviewer] input length: 53357 chars  (/merge_requests?per_page=100)
[logic_reviewer] input length: 56581 chars  (same)
[security_reviewer] input length: 53357 chars  (same)
[test_reviewer] input length: 56581 chars  (same)
[scope_estimator] input length: 72456 chars  (/merge_requests?state=merged)
...
\`\`\`

**21 agents × ~70 KB average × 60 ticks/hour × 24h = ~2 GB/day of input tokens going to the LLM backend.** Per day. Per federation. And the second federation (X/Y2) doubles it. And nothing is cached or deduplicated.

### On per-token pricing (Anthropic API, OpenAI-compatible endpoints)

74 KB ≈ 18K input tokens × 21 agents × 1440 ticks/day ≈ **540 million input tokens/day**.

| Model | $/M input | Daily cost |
|---|---|---|
| Claude Sonnet 4.5 | ~$3 | ~$1,620/day |
| Claude Opus 4.5 | ~$15 | ~$8,100/day |
| Claude Haiku 4.5 | ~$1 | ~$540/day |

(Output tokens add more. Caching would help a lot but agentis-core has to support it, and even then redundant full payloads defeat the purpose.)

Nobody will run this on per-token for long.

### On flat-rate Claude CLI

No per-token bill, but the Claude subscription has an hourly / daily rate limit. 21 × 60 = 1260 prompts/hour is well past the usage limits on any individual Claude plan. Operators will hit throttling walls constantly, and from the agent's perspective those will look like transient errors → more retries → more throttling.

Plus: **the same 74 KB payload is sent three times per tick** (labeler, router, prioritizer all pull \`/issues\`), and four times for MRs (four code-review agents all call \`/merge_requests\`). Even without dedup across agents, simply **sharing the latest GitLab payload within a colony** would cut cost / rate usage ~70%.

## Root cause

\`.ag\` agents fetch and pass *raw* GitLab responses into \`prompt()\`:

\`\`\`ag
// labeler.ag
let raw = exec sh issues_cmd();
let analysis = prompt(
    \"Analyze this GitLab issues JSON. Count issues with labels...\",
    raw  // <-- 74 KB of JSON, most of which the prompt doesn't need
) -> LabelAnalysis;
\`\`\`

The JSON includes every field GitLab returns: descriptions, \`_links\`, timestamps down to microseconds, author PII, web URLs for every note, label objects (not just label names), milestone objects, project object. For \"count issues with labels\" — the actual prompt task — maybe 1 KB of signal is needed. The other 73 KB is padding the operator is paying for.

### Short-term (dev-apprenticeship): downselect payload before prompting

Each \`gitlab-api.sh\` endpoint should be parametrized by *which fields the caller needs*. Or \`.ag\` agents should project the JSON with \`jq\` before \`prompt()\`. For the \`labeler\` case:

\`\`\`ag
// Before prompt, reduce to just what the LLM needs to answer:
let trimmed = exec sh (\"$COLONY_DIR/scripts/gitlab-api.sh issues | jq '[.[] | {iid, title, labels: [.labels[]?]}]'\");
let analysis = prompt(\"Analyze ...\", trimmed);
\`\`\`

Typically 5–10x smaller payload. Actual numbers for labeler: 74 KB → ~6 KB.

### Short-term (dev-apprenticeship): dedupe fetches within a colony

Triage has three agents all pulling \`/issues\` three times per tick. A single \"poller\" agent could fetch once, broadcast the trimmed payload via \`emit()\`, and other agents \`listen()\` for it. Factor 3 saving for triage, factor 4 for code-review.

### Long-term (agentis-core): prompt caching

agentis-core should support whatever the LLM backend offers for prompt caching (Claude's prompt cache, OpenAI's prefix cache). A per-colony \"here is the current state of GitLab\" prefix that gets cached, with a small per-agent suffix \"and here's my specific question\" would dramatically cut cost. But this is orthogonal to dev-apprenticeship — dev-apprenticeship first needs to not waste a factor 5–10x on raw-JSON-as-prompt-input before caching helps.

## Why it matters now

Combined with:
- #109 (9 agents stuck in type-error loop, each loop is still a prompt call burning tokens)
- #110 (router/prioritizer stuck on adaptive engine, each retry burns tokens)
- #115 (agents that recover from transient errors eventually die anyway)

…the current dev-apprenticeship shipped state is \"burn a lot of LLM budget, learn almost nothing\". Fixing #109/#110/#115 without also fixing the payload size just means the waste converts from \"retries of broken prompts\" to \"successful prompts doing work, but with 90% padding\".

## Suggested priority

Not the same day as #107/109/110/115 fixes — those unblock functionality first. But it should be on the roadmap before anyone is asked to run dev-apprenticeship on a paid Anthropic account. On the flat-rate Claude CLI path, anything that reduces 1260 prompts/hour is operator-critical.

## Environment

- agentis: v1.1.8
- 2 federations running simultaneously on same Claude CLI

## Related

- #107 (exec timeout / PII — same data-path, different failure mode)
- #115 (curl timeout + permanent classification — fewer successful polls would also help rate-limit exposure)
- #117 (gitlab-api.sh error handling — 429 rate-limit specifically)

ylohnitram:
## Summary

- Add opt-in named `--view` flag to list endpoints in every colony's `scripts/gitlab-api.sh` (triage, code-review, planning, implementation, release). Each view is a `python3` projection (no `jq` dependency) defined inline in a `project_json` function at the top of the script. Default behavior unchanged — omitting `--view` returns today's full payload.
- Add 14 views across the 5 scripts, shrinking payloads ~10x for the agent's `prompt()` call (issue #119 goal). See commit body for the full list.
- Thread `--view <name>` through every matching `exec sh .../gitlab-api.sh` call site in `dev-apprenticeship/*/agents/*.ag`. Detail endpoints (`issue/<iid>`, `mr-changes`, `mr-commits`, `mr-notes`) are deliberately left unchanged — those agents genuinely need the full payload.
- `GITLAB_VIEW_MODE=raw` env-var escape hatch forces every view back to pass-through globally — one-var rollback for operators if a projection drops a field an agent silently depended on.
- Unknown view exits 2 via `emit_error`, matching the existing "unknown flag" contract.
- New `tools/test-gitlab-views.sh`: 45 assertions covering every (colony, view) pair against inline noisy GitLab fixtures. Verifies (a) valid JSON, (b) exact top-level key set per array element, (c) output <= 30% of input size. Picked up automatically by `colony-lint.sh`'s `test-*.sh` discovery.
- Bash 3.2 compatible throughout (issue #121 baseline): no `mapfile`, no associative arrays, no `${var^^}`, no backslash-newline inside case labels.

## Test plan

- [x] `bash tools/colony-lint.sh` → 46 passed / 0 failed (was 45 / 0)
- [x] `bash tools/test-gitlab-views.sh` → 45 passed / 0 failed
- [x] `bash -n` on all 5 modified `gitlab-api.sh` scripts
- [x] `bash -n` on new `tools/test-gitlab-views.sh`
- [x] All 21 `.ag` files still parse via `agentis commit` (lint-step in colony-lint)
- [x] `--view raw` and `GITLAB_VIEW_MODE=raw` both produce pass-through output
- [x] Unknown view (`--view bogus`) exits 2 with JSON error on stderr
- [ ] CI shellcheck clean on the modified scripts (runs in GH Actions; skipped locally)

Closes #119

### DIFF
--- a/dev-apprenticeship/code-review/agents/approval_decider.ag
+++ b/dev-apprenticeship/code-review/agents/approval_decider.ag
@@ -44,7 +44,7 @@ fn tick(reason: string) -> void {
     };
 
     // Learn from past human approval/rejection patterns
-    let mr_cmd = "$COLONY_DIR/scripts/gitlab-api.sh merge-requests --state merged";
+    let mr_cmd = "$COLONY_DIR/scripts/gitlab-api.sh merge-requests --state merged --view reviewer";
     let recent_merged = try { exec sh mr_cmd; } catch e { "[]"; };
 
     let approval_patterns = prompt(
--- a/dev-apprenticeship/code-review/agents/logic_reviewer.ag
+++ b/dev-apprenticeship/code-review/agents/logic_reviewer.ag
@@ -24,9 +24,9 @@ type LogicReview {
 fn mr_cmd() -> string {
     let ts = recall_latest("logic_reviewer:last_check");
     if len(ts) > 0 {
-        return "$COLONY_DIR/scripts/gitlab-api.sh merge-requests --since " + shell_escape(ts);
+        return "$COLONY_DIR/scripts/gitlab-api.sh merge-requests --since " + shell_escape(ts) + " --view reviewer";
     };
-    return "$COLONY_DIR/scripts/gitlab-api.sh merge-requests";
+    return "$COLONY_DIR/scripts/gitlab-api.sh merge-requests --view reviewer";
 }
 
 fn get_confidence() -> float {
--- a/dev-apprenticeship/code-review/agents/security_reviewer.ag
+++ b/dev-apprenticeship/code-review/agents/security_reviewer.ag
@@ -25,9 +25,9 @@ type SecurityReview {
 fn mr_cmd() -> string {
     let ts = recall_latest("security_reviewer:last_check");
     if len(ts) > 0 {
-        return "$COLONY_DIR/scripts/gitlab-api.sh merge-requests --since " + shell_escape(ts);
+        return "$COLONY_DIR/scripts/gitlab-api.sh merge-requests --since " + shell_escape(ts) + " --view reviewer";
     };
-    return "$COLONY_DIR/scripts/gitlab-api.sh merge-requests";
+    return "$COLONY_DIR/scripts/gitlab-api.sh merge-requests --view reviewer";
 }
 
 fn get_confidence() -> float {
--- a/dev-apprenticeship/code-review/agents/style_reviewer.ag
+++ b/dev-apprenticeship/code-review/agents/style_reviewer.ag
@@ -24,9 +24,9 @@ type StyleReview {
 fn mr_cmd() -> string {
     let ts = recall_latest("style_reviewer:last_check");
     if len(ts) > 0 {
-        return "$COLONY_DIR/scripts/gitlab-api.sh merge-requests --since " + shell_escape(ts);
+        return "$COLONY_DIR/scripts/gitlab-api.sh merge-requests --since " + shell_escape(ts) + " --view reviewer";
     };
-    return "$COLONY_DIR/scripts/gitlab-api.sh merge-requests";
+    return "$COLONY_DIR/scripts/gitlab-api.sh merge-requests --view reviewer";
 }
 
 fn get_confidence() -> float {
--- a/dev-apprenticeship/code-review/agents/test_reviewer.ag
+++ b/dev-apprenticeship/code-review/agents/test_reviewer.ag
@@ -24,9 +24,9 @@ type TestReview {
 fn mr_cmd() -> string {
     let ts = recall_latest("test_reviewer:last_check");
     if len(ts) > 0 {
-        return "$COLONY_DIR/scripts/gitlab-api.sh merge-requests --since " + shell_escape(ts);
+        return "$COLONY_DIR/scripts/gitlab-api.sh merge-requests --since " + shell_escape(ts) + " --view reviewer";
     };
-    return "$COLONY_DIR/scripts/gitlab-api.sh merge-requests";
+    return "$COLONY_DIR/scripts/gitlab-api.sh merge-requests --view reviewer";
 }
 
 fn get_confidence() -> float {
--- a/dev-apprenticeship/code-review/scripts/gitlab-api.sh
+++ b/dev-apprenticeship/code-review/scripts/gitlab-api.sh
@@ -8,13 +8,19 @@
 #   GITLAB_PROJECT - URL-encoded project path
 #
 # Usage:
-#   gitlab-api.sh merge-requests [--since ISO8601] [--state opened|merged|all]
+#   gitlab-api.sh merge-requests [--since ISO8601] [--state opened|merged|all] [--view <name>]
 #   gitlab-api.sh mr-changes <iid>
 #   gitlab-api.sh mr-notes <iid>
 #   gitlab-api.sh post-note <iid> --body <text>
 #   gitlab-api.sh approve <iid>
 #
-# Returns JSON to stdout. Exit code 0 on success, 1 on error, 2 on unknown flag.
+# Views (opt-in projection; default is full JSON):
+#   merge-requests --view reviewer  [{iid, state, title, labels,
+#                                     source_branch, target_branch, draft}]
+#   merge-requests --view raw       explicit pass-through (same as no flag)
+#   GITLAB_VIEW_MODE=raw            env-override that forces pass-through globally.
+#
+# Returns JSON to stdout. Exit code 0 on success, 1 on error, 2 on unknown flag/view.
 
 set -e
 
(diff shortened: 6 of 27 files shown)


ylohnitram:
## PR #126 Review

### Verdict: REQUEST CHANGES

| # | File | Line | Issue |
|---|------|------|-------|
| 1 | all 5 `dev-apprenticeship/*/scripts/gitlab-api.sh` | `gl_get(_q) ... \| project_json "$VIEW"` branches | **Error-propagation regression on `--view` paths.** `gl_call` returns 2/3/4/5 on auth/rate-limit/4xx/5xx without writing to stdout. Because the scripts don't `set -o pipefail`, the final exit code is whatever `project_json` returns. Reproduced with a mock that returns 2: `gl_get \| project_json "labeler"` → python `json.loads("")` throws `JSONDecodeError`, pipeline exits `1` (masking `2`) and adds a Python traceback to stderr on top of the actionable `auth failure (HTTP 401)` line from `gl_call`. With `--view raw` (or `GITLAB_VIEW_MODE=raw`) it's worse: `cat`/`printf` exits `0`, so the auth failure is silently swallowed and the caller's `try/catch` never fires. Before #126, these branches were bare `gl_get_q` calls with no pipe, so the exit code survived. Fix: add `set -o pipefail` (cheap), or only apply the projection after capturing success, e.g. `out=$(gl_get_q ...) && printf '%s' "$out" \| project_json "$VIEW"`. |

| 1 | `dev-apprenticeship/release/scripts/gitlab-api.sh` | 93-100 (`release-mr` view) | View drops MR `description`. `version_bumper.ag:115` and `changelog_writer.ag:111` ask the LLM to decide major/minor/patch and group changes by type — both signals (conventional-commits `BREAKING CHANGE:` footers, `release-note` prose) live in the description. Title + labels + target_branch cover the common cases but you lose the BREAKING-CHANGE signal. Consider adding `description` to `release-mr`, or documenting the loss. |
| 2 | `tools/test-gitlab-views.sh` | 175-198 | Test file exercises every declared view but does NOT cover: (a) `GITLAB_VIEW_MODE=raw` env override short-circuit, (b) unknown-view exit-2 path, (c) the `raw` arg pass-through. The PR checklist claims these were verified interactively; they'd be cheap to add as 3 more assertions so they don't regress silently. |
| 3 | `dev-apprenticeship/triage/scripts/gitlab-api.sh` | 415-419, 435-439 | The `members`/`labels --view <name>` dispatch falls through for unknown names (`*) INTERNAL_VIEW="$VIEW"`) which means `members --view labeler` silently produces `[{iid:null,title:null,labels:[]}]` with rc=0 instead of exit 2. Not reachable from current `.ag` sites (literal view names only), but inconsistent with the "unknown view exits 2" contract. Tightening the `*)` to reject unknown user-facing names would match the documented behaviour. |
| 4 | pre-existing | `release/agents/changelog_writer.ag:120`, `release/agents/ship_decider.ag:89` | Confirmed: `merge-requests --per-page 1` is passed to scripts that don't accept `--per-page` on that command. Exits 2 ("unknown flag: --per-page"), swallowed by `try/catch`, `mr_iid` ends up 0 so the downstream `post-note`/`ship_decider` note never fires. Pre-existing (not caused by #126), but file a tracking issue so the automated note-posting in those two agents isn't silently dead. |

### Positive Notes
- **View-field parity checked across all 19 `.ag` call sites.** Each site's `prompt()` context needs only fields that the corresponding view preserves. No hard field misses; the `release-mr` description drop (Suggestion #1) is the only semantic degradation worth flagging.
- **Unknown view exit-2 works in all 5 scripts** (verified with `project_json bogus_view_name` → `{"error":"unknown view: bogus_view_name"}` on stderr, rc=2). Same for `GITLAB_VIEW_MODE=raw` short-circuit — all 5 scripts honour it, including through the `members`/`labels` internal-dispatch wrapper.
- **`tools/test-gitlab-views.sh` is self-contained** — no network, fixtures are inline and realistically noisy (HTML/URLs/nested user objects) so the ≤30% size assertion reflects real GitLab payload shape. All 15 declared views are exercised with 3 assertions each (valid-JSON + exact-keys + size). `bash tools/colony-lint.sh` → 46 passed / 0 failed, as claimed.
- **Bash 3.2 compat clean** — no `declare -A`, no `mapfile`/`readarray`, no `${var^^}`, no backslash-newline inside case-pattern labels. `bash -n` passes on all 6 modified/new scripts.
- **python3 heredoc + `DATA="$(cat)"` stdin-capture pattern is consistent** across all 5 scripts and both `project_json` and the JSON-body builders (`create-issue`, `update-issue`, `commit-files`, `create-mr`, etc.).
- **Projection only runs on `gl_call` 2xx output** by construction (non-2xx writes to stderr, not stdout; `project_json` reads stdin), so retry/auth paths don't accidentally re-run through projection — they just have the exit-code propagation hole noted in Critical #1.

ylohnitram:
@ylohnitram Follow-up commit `7b91280` pushed addressing the review. Summary of what changed:

**Critical — error masking on the --view pipe (all 5 scripts):** applied option A (two-step `body="$(gl_get_q ...)" || exit $?` then `printf '%s' "$body" | project_json "$VIEW"`). Verified with a mocked `gl_get_q` returning 2 — exit code survives the pipe cleanly. No `set -o pipefail` added, so no blast radius on the other pipes in the scripts.

**release-mr view:** added `description` to the kept-fields set in `release/scripts/gitlab-api.sh` so `version_bumper.ag` / `changelog_writer.ag` can see `BREAKING CHANGE:` footers and release-note prose. Test expectations for `release/release-mr` updated accordingly.

**triage members/labels dispatch:** `*)` arm now emits `unknown view: $VIEW` and exits 2 instead of silently using the bogus name as an internal view. Consistent with the main `--view` handler.

**Test coverage gaps:** added three assertions in `tools/test-gitlab-views.sh`:
1. `GITLAB_VIEW_MODE=raw` short-circuit (output equals input byte-for-byte)
2. unknown view exits 2 with empty stdout
3. explicit `--view raw` arg pass-through

`bash tools/test-gitlab-views.sh` → 48 passed / 0 failed locally. `bash tools/colony-lint.sh` → 46 passed / 0 failed locally (count unchanged — `test-*.sh` is one PASS line).

**Suggestion #4 (pre-existing, not a #126 regression):** filed #127 tracking the `merge-requests --per-page 1` dead branch in `changelog_writer.ag:120` (and `ship_decider.ag:89`). Not fixed here per reviewer note.

Re-check please.

Issue #125, opened 2026-04-14
dev-apprenticeship: every agent uses prompt() for trivial data parsing — ~21 wasted LLM calls per tick just reading back stored numbers


ylohnitram:
## Problem

Separate from the \"big JSON payload every tick\" cost pattern in #119: **every agent also makes wasted LLM calls for trivial data extraction** where a native type coercion would do the job.

The most egregious: **every single one of the 21 agents parses its own \`<agent>:confidence\` memo through the LLM**.

\`\`\`ag
// triage/agents/labeler.ag (same pattern in all 21 agents)
fn get_confidence() -> float {
    let raw = recall_latest(\"labeler:confidence\");
    if len(raw) > 0 {
        return prompt(\"Return this number as a decimal. Nothing else.\", raw) -> float;  // <-- LLM call
    };
    return 0.0;
}
\`\`\`

\`raw\` is the string \`\"0.5\"\` or \`\"0.85\"\` — the value the operator set with \`agentis memo set\`. To turn it into a float, the agent asks Claude to do it. **Every tick. Every agent.**

\`\`\`bash
$ grep -rl \"Return this number as a decimal\" dev-apprenticeship/*/agents/*.ag | wc -l
21
\`\`\`

## Cost

- **1260 wasted prompts/hour** (21 × 60 ticks/hr) purely to re-parse floats.
- Each prompt is small — maybe 50–100 tokens in, 5 out — but on a paid Anthropic plan this still adds up (~4.5M tokens/day just on confidence parses, ~\$0.15/day at Sonnet rates, ~\$1/day at Opus rates — cheap per \*call\* but **per tick, unnecessary**).
- On flat-rate Claude CLI, it's 1260 requests/hour counted against the operator's usage cap. That's **21× the request volume** the actual learning work would generate if \`get_confidence()\` didn't need an LLM round-trip.

## Same pattern elsewhere

Other agents burn prompts for equally trivial extractions:

| Agent | Prompt | What it does |
|---|---|---|
| \`refactorer\`         | \"Extract the branch_name ... Return just the string, nothing else.\" | Should be \`draft.branch_name\` |
| \`plan_reviewer\`      | \"Return just the issue_id integer ... Nothing else.\" (×3 per tick) | Should be \`record.issue_id\` |
| \`changelog_writer\`   | \"Extract the iid integer of the first MR from this JSON array.\" | Should be \`json[0].iid\` |
[...]

PR #128, opened 2026-04-14
fix(dev-apprenticeship): accept --per-page on merge-requests (#127)

### LINKED ISSUE #127: release colony: changelog_writer calls merge-requests --per-page 1 which is silently unsupported
Surfaced by the PR #126 review (Suggestion #4 — pre-existing, not caused by #126).

### Summary

Two release-colony agents call `gitlab-api.sh merge-requests --per-page <N>`, but the `release/scripts/gitlab-api.sh merge-requests` subcommand does NOT accept `--per-page` (only `--state`, `--since`, and `--view`). The unknown flag triggers the default arm:

```sh
*) emit_error "unknown flag: $1"; exit 2 ;;
```

which makes the script exit 2. Both call sites are wrapped in `try { exec sh ... } catch e { "[]" }`, so the failure is swallowed. Downstream:

- `mr_iid` gets extracted from `"[]"` as `0`
- the `if mr_iid > 0` guard gates the note-posting branch
- therefore the automated `post-note` (changelog draft) / ship-decision note never fires

This is a dead branch that has looked healthy in logs (no error, no exception) but has never actually posted a note in production.

### Affected sites

- `dev-apprenticeship/release/agents/changelog_writer.ag:120`
  ```
  let mr_raw = try { exec sh \"\$COLONY_DIR/scripts/gitlab-api.sh merge-requests --per-page 1\"; } catch e { \"[]\"; };
  ```
- `dev-apprenticeship/release/agents/ship_decider.ag:89` (same pattern)

### Suggested fix

Either:
1. Add `--per-page <N>` as a supported flag on the `merge-requests` subcommand of `release/scripts/gitlab-api.sh` (symmetry with `releases`/`tags`/`pipelines` which already accept it). OR
2. Drop `--per-page 1` from the call sites and use `--view release-mr` + a downstream \"first element\" extraction (probably cleaner now that projections exist).

Option 1 is the smaller patch. Option 2 removes one LLM prompt round-trip (JSON array head can be done in a shell step), which aligns with the #119 cost-reduction push.

### Not a #126 regression

The `merge-requests --per-page` misuse predates #126. Filed as a follow-up so the silent-failure doesn't get forgotten after the #126 review comment is checked off.

ylohnitram:
## Summary

- **#127**: `merge-requests --per-page <N>` was silently unsupported across all four dev-apprenticeship `gitlab-api.sh` copies. Release colony's `changelog_writer.ag` and `ship_decider.ag` called it with `--per-page 1`, which tripped the `*) emit_error "unknown flag: $1"; exit 2` arm. The `try { exec sh ... } catch e { "[]" }` wrapper swallowed the failure, so `mr_iid` always extracted as `0` and the `if mr_iid > 0` note-posting branch never fired.
- Added `--per-page N` to the `merge-requests` subcommand in all four copies (code-review, implementation, planning, release), routed through a `PER_PAGE` variable.
- Preserved each file's pre-existing default: `20` for code-review/implementation/planning, `100` for release.
- Updated the `# Usage:` header in all four files.

## Test plan

- [x] `./tools/colony-lint.sh` — 46/46 pass
- [x] `bash -n` all four files
- [x] Smoke: `GITLAB_URL=http://example GITLAB_TOKEN=x GITLAB_PROJECT=x bash .../release/scripts/gitlab-api.sh merge-requests --per-page 1` — reaches the curl call (HTTP transport error) instead of failing at arg-parse with `unknown flag`.
- [ ] CI colony-lint

## Not migrated in this PR

The `prompt()`-based `mr_iid` extraction at `changelog_writer.ag:120` / `ship_decider.ag:89` is still a wasted LLM call. That belongs in the post-v1.2.0 migration to `json_get` (tracked under #125 and the pending core PR #506).

Closes #127

### DIFF
--- a/dev-apprenticeship/code-review/scripts/gitlab-api.sh
+++ b/dev-apprenticeship/code-review/scripts/gitlab-api.sh
@@ -8,7 +8,7 @@
 #   GITLAB_PROJECT - URL-encoded project path
 #
 # Usage:
-#   gitlab-api.sh merge-requests [--since ISO8601] [--state opened|merged|all] [--view <name>]
+#   gitlab-api.sh merge-requests [--since ISO8601] [--state opened|merged|all] [--per-page N] [--view <name>]
 #   gitlab-api.sh mr-changes <iid>
 #   gitlab-api.sh mr-notes <iid>
 #   gitlab-api.sh post-note <iid> --body <text>
@@ -220,17 +220,19 @@ case "$CMD" in
         SINCE=""
         STATE="opened"
         VIEW=""
+        PER_PAGE=20
         while [ $# -gt 0 ]; do
             case "$1" in
                 --since) SINCE="$2"; shift 2 ;;
                 --state) STATE="$2"; shift 2 ;;
                 --view) VIEW="$2"; shift 2 ;;
+                --per-page) PER_PAGE="$2"; shift 2 ;;
                 *) emit_error "unknown flag: $1"; exit 2 ;;
             esac
         done
         ARGS=(
             --data-urlencode "state=$STATE"
-            --data-urlencode "per_page=20"
+            --data-urlencode "per_page=$PER_PAGE"
             --data-urlencode "order_by=updated_at"
             --data-urlencode "sort=desc"
         )
--- a/dev-apprenticeship/code-review/scripts/gitlab-api.sh
+++ b/dev-apprenticeship/code-review/scripts/gitlab-api.sh
@@ -230,6 +230,9 @@ case "$CMD" in
                 *) emit_error "unknown flag: $1"; exit 2 ;;
             esac
         done
+        case "$PER_PAGE" in
+            ''|*[!0-9]*) emit_error "invalid --per-page: $PER_PAGE (integer required)"; exit 2 ;;
+        esac
         ARGS=(
             --data-urlencode "state=$STATE"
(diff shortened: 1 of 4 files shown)


ylohnitram:
## QA Report — PR #128

### CI
- colony-lint: SUCCESS (46 pass, 0 fail, 5 skipped on HEAD 8df127a)

### Review verdict: APPROVE
- All four gitlab-api.sh copies (code-review/implementation/planning/release) updated with:
  - `PER_PAGE=<file-specific default>` init (20/20/20/100 preserved)
  - `--per-page) PER_PAGE="$2"; shift 2 ;;` arg handler
  - `--data-urlencode "per_page=$PER_PAGE"` in ARGS
  - `# Usage:` header updated
- No regressions in sibling subcommands (`releases`/`tags`/`pipelines` have their own per-arm `PER_PAGE` reassignments).
- Input-validation guard added post-review: POSIX-safe `case "$PER_PAGE" in ''|*[!0-9]*)` rejects non-integer values before the HTTP call, preventing the exact silent-failure class (`try/catch -> "[]"`) that #127 was about.

### Smoke tests
- `--per-page abc` → `{"error": "invalid --per-page: abc (integer required)"}`, exit 2
- `--per-page 5` → reaches curl (network error, as expected with fake creds)
- `--per-page 1` (the changelog_writer / ship_decider pattern from #127) → now works as intended

Ready to merge.

PR #139, opened 2026-04-15
feat(dashboard): restart-required toast after confidence write (#137)

### LINKED ISSUE #137: federation-dashboard: promote writes memo file but running daemons keep stale in-memory value
## Summary

The federation-dashboard UI promote button (added per #105) correctly appends a new `{"generation":N,"timestamp":T,"value":"0.600"}` entry to `memo/<role>:confidence.jsonl`. However, the running agent daemon DOES NOT pick up the new value — it keeps using the confidence it loaded at spawn. The UI therefore gives a misleading "promoted" confirmation while the daemon tick behaviour is unchanged.

## Reproduction

1. Start dev-apprenticeship federation, let planning agents settle into `tick_ok > 0`.
2. In the dashboard UI, promote `risk_assessor`, `task_decomposer`, `scope_estimator` from 0.5 to 0.6.
3. Verify `tail memo/risk_assessor:confidence.jsonl` — new line with `value: "0.600"` is present. File write is correct.
4. Verify the running daemon: there is no CLI surface today to query its live confidence (see the companion agentis-core issue for the `daemon list --json` missing-field). But indirect evidence that the daemon did not reload:
   - No log line in `.agentis/logs/<agent>.log` mentioning confidence change.
   - No registry file mtime update that would indicate a reload.
   - Manual restart of the daemon is required before tick behaviour differs.

## Root cause

`agentis` itself has no live config reload mechanism (verified by grepping `--help` for `reload`/`refresh`/`watch`/`hot` — none found). Daemons load confidence once at spawn time. The dashboard write is therefore persistent-only, not runtime-effective.

## Expected behaviour

Option 1 — **warn the user** that the promote needs a restart to take effect. Minimal change: add a toast/banner after the promote action that reads _"Confidence updated in memo. Restart the daemon (`agentis daemon stop <id>` + respawn) for the new value to take effect in runtime."_

Option 2 — **auto-restart the affected daemon** from the dashboard as part of the promote action. Requires the dashboard to know how to respawn (currently the spawn logic lives in `start-colony.sh` per colony, with env vars + flags; dashboard would need to shell out).

Option 3 — **agentis-core adds live reload**. Filed separately; out of scope for this repo.

The minimum acceptable fix is Option 1 — today's UX silently misleads the operator into believing they have promoted an agent when in reality nothing about the agent's behaviour has changed until the next restart cycle.

## Related

- #105 — the per-agent confidence UI that introduced this behaviour.
- Companion agentis-core issue asking for `daemon list --json` to surface live confidence so the dashboard can compare memo-on-disk vs daemon-in-memory and flag the drift.

ylohnitram:
## Summary

Closes #137. After a successful `/confidence` write from the ▲/▼ buttons, show a CSS-animated DOM toast reminding the operator that the running daemon keeps using its in-flight value until its next tick (~60 s) and spawn-time decisions do not reload — so `agentis daemon stop --all && ./start-federation.sh` is required for immediate, guaranteed effect.

- **CSS** (`tools/federation-dashboard.sh` inside `<style>`): `.toast` + `.toast-head` + `.toast pre` + `.toast-foot` + `@keyframes toast-in`, using the dashboard's existing `--yellow` / `--cyan` vocabulary and `Share Tech Mono`.
- **JS** `setConfidence` / new `showRestartToast`: fires the toast on HTTP 200 regardless of direction (promote **and** demote — demote is safety-critical, operator trying to *stop* an autonomous agent), defers `location.reload()` by 12 s so the toast stays readable. Toast is built via `createElement`/`textContent` (no `innerHTML` from inputs), exposes `role="status"` + `aria-live="polite"` for screen readers, click-dismiss, 12 s auto-dismiss. `confirm()` preamble for `value >= 0.85` and the error path are unchanged.
- **Backend** `/confidence`: switched from `text/plain` to `application/json` with `memo_written` / `audit_logged` / `restart_required` fields. No external consumers of the response body exist today (only the inline browser JS), so this is safe.

Wording intentionally says "**requires restart to take immediate effect**" — strictly true given `recall_latest` is file-fresh each tick, unlike the stronger "never applies" claim in the original ticket. Rationale recorded in the planning comment on #137.

## Implementation notes

- No shell control flow altered — `colony-lint.sh` passes 46/0/5 (shellcheck skipped: not installed locally, unchanged from baseline).
- `bash -n tools/federation-dashboard.sh` clean.
- `dev-apprenticeship/dashboard.sh` wrapper unchanged per plan.
- Plan link: https://github.com/Replikanti/agentis-colonies/issues/137#issuecomment-4252129573

## Test plan

Backend curl regressions (can run automated):

- [ ] `curl -sS -X POST http://localhost:8420/confidence -d 'agent=router&value=0.6'` → HTTP 200, JSON body with `memo_written: true`, `restart_required: true`.
- [ ] `curl -sS -X POST http://localhost:8420/confidence -d 'agent=bogus&value=0.6'` → HTTP 400 `unknown agent: 'bogus'`.
- [ ] `curl -sS -X POST http://localhost:8420/confidence -d 'agent=router&value=1.5'` → HTTP 400 `value out of [0,1]: 1.5`.

Manual browser tests (§6 of the plan — **not executed by the implementation agent; reviewer / operator needed**):

- [ ] **M1** — Click ▲ on agent at 0.50 (→ 0.60). No `confirm()`. Toast appears bottom-right with yellow border. Reloads after ~12 s. Log row appended to `.dashboard/confidence-log.jsonl`.
- [ ] **M2** — Click ▲ on agent at 0.60 (→ 0.85). `confirm()` first; accept. Toast fires. Reloads after ~12 s.
- [ ] **M2b** — Same as M2 but cancel `confirm()`. No POST, no toast, no reload.
- [ ] **M3** — Click ▼ (demote 0.85 → 0.60). No `confirm()`. Toast fires (symmetric rule — demoting an autonomous agent is safety-critical).
- [ ] **M4** — Click ▲ on two agents within a second. Two toasts stack; each dismisses independently.
- [ ] **M5** — Click toast body. Disappears immediately; reload still fires at t=12 s.
- [ ] **M6** — Simulate backend error (e.g. temporarily block `agentis` on PATH) and click ▲. `alert('Set failed: ...')` appears; no toast; no reload.
- [ ] **M7** — Keyboard-only navigation with a screen reader. Toast is announced via `aria-live="polite"` without stealing focus from the buttons.
- [ ] **M8** — After toast fires, run `agentis daemon stop --all && ./start-federation.sh`, then `agentis memo get <agent>:confidence`. Value matches what the toast announced — proves the recommended command works end-to-end.

### DIFF
--- a/tools/federation-dashboard.sh
+++ b/tools/federation-dashboard.sh
@@ -693,13 +693,23 @@ function setConfidence(agent, value) {
     headers: {'Content-Type': 'application/x-www-form-urlencoded'},
     body: body,
   })
-    .then(r => r.text().then(t => ({ok: r.ok, text: t})))
-    .then(({ok, text}) => {
-      if (!ok) {
-        alert('Set failed: ' + text);
+    .then(r => {
+      // 200 returns application/json; 4xx/5xx return text/plain. Parse per branch.
+      if (r.ok) return r.json().then(data => ({ok: true, data}));
+      return r.text().then(text => ({ok: false, text}));
+    })
+    .then(result => {
+      if (!result.ok) {
+        alert('Set failed: ' + result.text);
         return;
       }
-      showRestartToast(agent, value);
+      // The page has <meta http-equiv="refresh" content="60"> in <head>. Without
+      // removing it, a pending browser-level refresh can fire inside the 12 s
+      // toast window and destroy the toast before the operator finishes reading
+      // — same reason the kill-switch handler removes it on success.
+      const m = document.querySelector('meta[http-equiv="refresh"]');
+      if (m) m.remove();
+      showRestartToast(agent, value, result.data);
       // Defer the static-HTML regen + reload so the toast stays readable.
       // Without the delay location.reload() destroys the toast before the
       // operator can see the restart command.
@@ -710,12 +720,16 @@ function setConfidence(agent, value) {
     .catch(e => alert('Set failed: ' + e));
 }
 
-function showRestartToast(agent, value) {
+function showRestartToast(agent, value, data) {
   const el = document.createElement('div');
   el.className = 'toast';
   el.setAttribute('role', 'status');
   el.setAttribute('aria-live', 'polite');
-  const v = (typeof value === 'number') ? value.toFixed(2) : String(value);
+  // Prefer the server's authoritative value (it's what actually landed on
+  // disk via `agentis memo set`) and fall back to the click-site value.
+  const serverValue = data && typeof data.value === 'string' ? parseFloat(data.value) : NaN;
+  const shown = Number.isFinite(serverValue) ? serverValue : value;
+  const v = (typeof shown === 'number') ? shown.toFixed(2) : String(shown);
   const head = document.createElement('div');
   head.className = 'toast-head';
   head.textContent = 'Memo written — daemon restart required';
@@ -731,7 +745,12 @@ function showRestartToast(agent, value) {
   pre.textContent = 'agentis daemon stop --all && ./start-federation.sh';
   const foot = document.createElement('div');
   foot.className = 'toast-foot';
-  foot.textContent = 'Audit: appended to .dashboard/confidence-log.jsonl — click to dismiss';
+  // audit_logged comes from the backend; missing/undefined means legacy
+  // response — fall back to the optimistic wording rather than a false claim.
+  const auditOk = !data || data.audit_logged !== false;
+  foot.textContent = auditOk
+    ? 'Audit: appended to .dashboard/confidence-log.jsonl — click to dismiss'
+    : 'Audit write failed — restart still required. Click to dismiss.';
   el.appendChild(head);
   el.appendChild(line1);
   el.appendChild(line2);
--- a/tools/federation-dashboard.sh
+++ b/tools/federation-dashboard.sh
@@ -461,6 +461,32 @@ HEADEOF
     0%, 100% { box-shadow: 0 0 10px rgba(255,0,255,0.3); }
     50% { box-shadow: 0 0 25px rgba(255,0,255,0.6); }
   }
+  .toast {
+    position: fixed; right: 16px; bottom: 16px; z-index: 9999;
+    max-width: 420px; padding: 12px 14px 12px 16px;
+    background: var(--surface); border: 1px solid var(--yellow);
+    border-radius: 4px; color: var(--text);
+    font-family: 'Share Tech Mono', 'Courier New', monospace;
+    font-size: 12px; line-height: 1.6;
+    box-shadow: 0 0 20px rgba(255,255,0,0.35);
+    cursor: pointer;
+    animation: toast-in 0.25s ease-out;
+  }
+  .toast .toast-head {
+    color: var(--yellow); font-size: 11px; letter-spacing: 2px;
+    text-transform: uppercase; margin-bottom: 6px;
+    text-shadow: 0 0 6px rgba(255,255,0,0.4);
+  }
+  .toast pre {
+    margin: 6px 0; padding: 6px 8px; background: rgba(0,255,255,0.05);
+    border-left: 2px solid var(--cyan); color: var(--cyan);
+    font-size: 11px; white-space: pre-wrap; word-break: break-all;
+  }
+  .toast .toast-foot { color: var(--muted); font-size: 10px; margin-top: 6px; }
+  @keyframes toast-in {
+    from { opacity: 0; transform: translateY(8px); }
+    to   { opacity: 1; transform: translateY(0); }
+  }
   ::-webkit-scrollbar { width: 4px; }
   ::-webkit-scrollbar-track { background: transparent; }
   ::-webkit-scrollbar-thumb { background: var(--border); border-radius: 2px; }
@@ -673,11 +699,47 @@ function setConfidence(agent, value) {
         alert('Set failed: ' + text);
         return;
       }
-      // Regenerate the static HTML so the reloaded page reflects the new
-      // memo value immediately instead of waiting for the 60 s background loop.
-      fetch('/refresh', { method: 'POST' }).catch(() => {}).finally(() => location.reload());
+      showRestartToast(agent, value);
+      // Defer the static-HTML regen + reload so the toast stays readable.
+      // Without the delay location.reload() destroys the toast before the
+      // operator can see the restart command.
+      setTimeout(() => {
+        fetch('/refresh', { method: 'POST' }).catch(() => {}).finally(() => location.reload());
+      }, 12000);
     })
     .catch(e => alert('Set failed: ' + e));
+}
+
+function showRestartToast(agent, value) {
+  const el = document.createElement('div');
+  el.className = 'toast';
+  el.setAttribute('role', 'status');
+  el.setAttribute('aria-live', 'polite');
+  const v = (typeof value === 'number') ? value.toFixed(2) : String(value);
+  const head = document.createElement('div');
+  head.className = 'toast-head';
+  head.textContent = 'Memo written — daemon restart required';
+  const line1 = document.createElement('div');
+  line1.textContent = agent + ':confidence set to ';
+  const b = document.createElement('b');
+  b.textContent = v;
+  line1.appendChild(b);
+  line1.appendChild(document.createTextNode(' on disk.'));
+  const line2 = document.createElement('div');
+  line2.textContent = 'The running daemon will keep using its in-flight value until its next tick (up to ~60 s) and spawn-time decisions will not reload. For immediate, guaranteed effect run:';
+  const pre = document.createElement('pre');
+  pre.textContent = 'agentis daemon stop --all && ./start-federation.sh';
+  const foot = document.createElement('div');
+  foot.className = 'toast-foot';
+  foot.textContent = 'Audit: appended to .dashboard/confidence-log.jsonl — click to dismiss';
+  el.appendChild(head);
+  el.appendChild(line1);
+  el.appendChild(line2);
+  el.appendChild(pre);
+  el.appendChild(foot);
+  el.addEventListener('click', () => el.remove());
+  document.body.appendChild(el);
+  setTimeout(() => { if (el.parentNode) el.remove(); }, 12000);
 }
 
 (function() {
@@ -1016,6 +1078,7 @@ class Handler(SimpleHTTPRequestHandler):
             # Audit log: append a JSONL row per change. .dashboard/ is
             # already operator-visible (index.html, history.json live here)
             # so this keeps all per-federation dashboard state colocated.
+            audit_ok = False
             try:
                 with open(confidence_log, 'a') as f:
                     f.write(json.dumps({
@@ -1024,13 +1087,20 @@ class Handler(SimpleHTTPRequestHandler):
                         'value': value,
                         'remote': self.client_address[0],
                     }) + '\n')
+                audit_ok = True
             except OSError:
                 # Audit write failure must not mask the successful memo set.
                 pass
             self.send_response(200)
-            self.send_header('Content-Type', 'text/plain')
+            self.send_header('Content-Type', 'application/json')
             self.end_headers()
-            self.wfile.write(f'{agent}:confidence set to {value:.3f}'.encode())
+            self.wfile.write(json.dumps({
+                'agent': agent,
+                'value': f'{value:.3f}',
+                'memo_written': True,
+                'audit_logged': audit_ok,
+                'restart_required': True,
+            }).encode())
             return
         if self.path == '/kill':
             # Run from fed_dir, not .dashboard. The actual fix for #91 is


ylohnitram:
## Code review

### Verdict
Request changes — one real interaction bug with the existing meta-refresh; the rest are nits.

### Findings

**Blocker**

- **Meta-refresh at line 313 will nuke the toast mid-countdown.** `<meta http-equiv="refresh" content="60">` is unconditional page-level auto-reload. If the operator clicks ▲ at t=55s after the page loads, the toast appears, and at t=60s the browser reloads — the restart command is on screen for ~5s, not the 12s the deferred `location.reload()` budget promises. Worst case (click at t=59s): ~1s visibility. The `killFederation` handler already demonstrates the fix at `tools/federation-dashboard.sh:909-910`:
  ```js
  const m = document.querySelector('meta[http-equiv="refresh"]');
  if (m) m.remove();
  ```
  `setConfidence` should do the same on HTTP 200 (before `showRestartToast`), otherwise the plan's §4.C "defer reload by 12s" guarantee is violated by the page-level refresh the JS timer doesn't know about. This is the central UX promise of the ticket — operators must see the restart command — so I'd flag it as a blocker, not a nit.

**Should fix**

- **Backend JSON fields are sent but never read.** `/confidence` now returns `{memo_written, audit_logged, restart_required}` (line 1097-1103) but the front end still does `r.text()` (line 696) and ignores everything except `r.ok`. Plan §4.B justified the JSON switch specifically so the toast could "report truthfully whether the audit line landed" — but the footer text hardcodes `Audit: appended to .dashboard/confidence-log.jsonl` regardless of `audit_logged`. Either parse the response and conditionally render the audit footer (the plan's actual intent), or revert to text and avoid the unused payload. As-is, the backend change is dead weight that expands the surface area of the contract for no behavioural gain.

- **Batch-promote / rapid-click creates N pending reload timers.** On 21 clicks in 5s, 21 `setTimeout(..., 12000)` timers are outstanding. Each fires a `/refresh` POST independently, then the first `location.reload()` wins and unloads the page (cancelling the rest). Not a correctness bug — the page does reload, and the allowlist+500 timeout on the Python side handle the duplicate regen fine — but worth debouncing: on entry to the `then` handler, `clearTimeout` any previously-stored reload handle and store the new one in a module-scoped variable. Avoids 21 concurrent bash subprocess regens pounding the server for nothing. Borderline between "should fix" and "nit"; calling it a should-fix because fixing the blocker above forces you into this code anyway.

**Nit / preference**

- `createElement` + `textContent` instead of plan's `innerHTML` — this is *better* than the plan, not worse. Plan assumed `agent` was allowlist-safe (true per line 1036) but defence-in-depth against future refactors is cheap. Good call. No change needed.

- `aria-live="polite"` vs `"assertive"`: the plan picked polite and the implementation followed. Polite queues behind current speech; a screen-reader user who just heard the `confirm()` dialog may have the toast delayed, and a 12s auto-dismiss does not wait for the SR queue. Given the toast says "daemon won't pick up change until next tick and spawn-time decisions won't reload" — which is closer to a warning than a status — `assertive` is defensible. Not blocking; just flagging that this is a judgement call the plan didn't interrogate.

- Symmetric promote+demote toast: I buy the argument. Demote 0.85→0.6 is an operator pulling an autonomous agent back — they need to know the daemon won't honour the rollback until the next tick (up to 60s of continued autonomous GitLab writes). Symmetry is correct. Counter-case of "operator doesn't want to see it on rollback" is weak — the toast is click-dismissable and auto-dismisses in 12s; the cost of showing it when unwanted is minimal, the cost of *not* showing it on a demote-to-safety is an agent still posting to GitLab after operator thought they'd stopped it.

- Shell-quoting of the heredoc: `<<'JSEOF'` at line 570 is single-quoted, so `$`, backticks, and `\` in the new JS are literal. The new code has no unprotected substitutions. Safe. Plan didn't mention this; good that the implementer picked the right block.

- Style: the `(typeof value === 'number') ? value.toFixed(2) : String(value)` branch at line 718 — `setConfidence` is only ever called from generated onclick handlers with a numeric literal (`nextStep()` returns `number|null` and `null` is filtered by the `disabled` path at line 759), so `value` is always a number here. Harmless defence but slightly misleading. Not worth changing.

- Audit-log silent failure (line 1091-1093, pre-existing — not in this diff): `except OSError: pass` with no logging. Outside the PR scope but worth a follow-up ticket: if disk fills and audit silently drops, the "Audit: appended to ..." footer becomes a lie. The new `audit_logged` JSON field was the right hook for this; it's just unused (see should-fix #1).

Verdict recap: the meta-refresh interaction is the only thing that prevents the PR from delivering on its own ticket, and the fix is two lines copied from the existing kill handler. Everything else is polish.

ylohnitram:
## QA report

Scope: commit `5cf5764` on `feat/137-dashboard-restart-toast`; single file `tools/federation-dashboard.sh`. Checks performed from the worktree against this exact SHA.

### Automated checks

- [x] **`bash -n tools/federation-dashboard.sh`** — clean (no syntax errors).
- [x] **`./tools/colony-lint.sh`** — `46 passed, 0 failed, 5 skipped` (shellcheck not installed locally; identical to baseline `origin/main` which also reports `46 passed, 0 failed, 5 skipped`). No regression introduced. CI `colony-lint` on the PR is also green.
- [x] **Python AST parse of the inline `PYSERVER` heredoc** — extracted lines 964–1148 into a standalone file (185 lines) and ran `python3 -c "import ast; ast.parse(...)"` → parses cleanly, proving the rewritten `/confidence` JSON branch is valid Python.
- [x] **Backend curl regressions (§6 of the plan)** — spun up a minimal federation (`/tmp/dash-test/minifed` with one colony and one agent, `agentis init` in the federation parent) and ran the dashboard on port 18421. All three cases behave as specified:
  - `agent=router&value=0.6` → `HTTP 200`, `Content-Type: application/json`, body `{"agent":"router","value":"0.600","memo_written":true,"audit_logged":true,"restart_required":true}`. Audit row appended to `.dashboard/confidence-log.jsonl`.
  - `agent=bogus&value=0.6` → `HTTP 400`, `Content-Type: text/plain`, body `unknown agent: 'bogus'` (unchanged from baseline — error path still plain text by design).
  - `agent=router&value=1.5` → `HTTP 400`, `Content-Type: text/plain`, body `value out of [0,1]: 1.5`.
- [x] **Git state** — worktree clean, branch is exactly one commit ahead of `origin/main` (`5cf5764`), pushed, `mergeable: MERGEABLE`, no conflicts.

### Static review of the new JS/CSS

- [x] **Zero `innerHTML` in the new block.** Grepped lines 682–743: 0 hits. The new `showRestartToast` builds the entire toast via `document.createElement(...)` + `textContent` / `appendChild` — exactly as the plan required (belt-and-braces XSS hygiene even though `agent` comes from an allowlist). Pre-existing `innerHTML` usage elsewhere in the file is untouched and out of scope.
- [x] **Error path intact.** `alert(` appears twice inside `setConfidence` (lines 699 and 710) — the non-2xx branch and the `fetch` rejection branch both still surface errors via `alert()`. No toast fires on failure. Confirmed against plan §5.
- [x] **`setTimeout` delay = 12000 ms** in both places: line 706 (deferred `location.reload()` inside `setConfidence`) and line 742 (toast auto-dismiss inside `showRestartToast`). Matches plan §4.C and §3 UX rationale.
- [x] **`confirm()` preamble unchanged.** The `value >= 0.85` promotion gate at lines 685–689 is byte-for-byte the pre-existing guard; only the body after the gate was rewritten.
- [x] **Toast fires on both promote and demote** — no directional branch in `setConfidence`; single unconditional `showRestartToast(agent, value)` call after the 2xx check. Matches plan §5 "symmetry rule" (demote from 0.85 is safety-critical).
- [x] **A11y attributes present.** `el.setAttribute('role', 'status')` and `el.setAttribute('aria-live', 'polite')` both set on the toast root before insertion.
- [x] **CSS lives in the right place** — `.toast` / `.toast .toast-head` / `.toast pre` / `.toast .toast-foot` / `@keyframes toast-in` appended immediately after `@keyframes pulse-kill` at lines 464–489, inside the existing `<style>` block (closed at line 493). Uses `var(--yellow)`, `var(--cyan)`, `var(--surface)`, `Share Tech Mono` — consistent with the dashboard's existing vocabulary.
- [x] **Backend JSON shape matches plan §4.B.** Keys returned: `agent`, `value`, `memo_written`, `audit_logged`, `restart_required`. `audit_logged` correctly reflects actual disk write success (wrapped in `try/except OSError`, `audit_ok` default `False`). Memo-set failure still returns 500 text (unchanged) — no false-positive "memo written" on failure.

Minor note (non-blocking): `CLAUDE.md` still says "Current baseline: 45 passed" but the actual lint count has been 46 for at least as long as this worktree; not introduced by this PR.

### Deferred to operator / reviewer (M1–M8)

The browser-side matrix enumerated in the PR body requires a live dashboard in a real browser and a running federation with `GITLAB_URL`/`GITLAB_TOKEN` configured. I verified only the HTTP-level contract those tests depend on (the three curl cases above); the visible/interactive behaviour still needs a human:

- [ ] **M1** — ▲ at 0.50 → 0.60, no `confirm()`, toast appears bottom-right with yellow border, reload fires ~12 s later, log row appears in `.dashboard/confidence-log.jsonl`.
- [ ] **M2** — ▲ at 0.60 → 0.85, `confirm()` fires first, accept, toast fires, reload after ~12 s.
- [ ] **M2b** — Same as M2 but cancel `confirm()` — no POST, no toast, no reload.
- [ ] **M3** — ▼ demote 0.85 → 0.60, no `confirm()`, toast fires (symmetric rule).
- [ ] **M4** — ▲ on two agents within ~1 s — toasts stack, dismiss independently.
- [ ] **M5** — Click toast — disappears immediately; reload still fires at t = 12 s.
- [ ] **M6** — Simulate backend error (e.g. shadow `agentis` on `PATH` with a failing stub) — `alert('Set failed: ...')` appears, no toast, no reload.
- [ ] **M7** — Keyboard-only navigation with a screen reader — toast announced via `aria-live="polite"` without stealing focus from the ▲/▼ buttons.
- [ ] **M8** — After toast fires: run `agentis daemon stop --all && ./start-federation.sh`, then `agentis memo get <agent>:confidence`. Value matches the toast. Proves the recommended restart command is correct end-to-end (this one needs GitLab creds to bring the full federation up).

### Verdict

**Ready for reviewer.** Every automated and curl-level check passes; the static review finds no deviations from the plan (`innerHTML`-free, `alert()` error path preserved, 12 000 ms delays, symmetric toast, JSON contract as specified); lint count unchanged from baseline. M1–M8 still need a human at a browser — but those are intentionally manual and are already checkboxed in the PR body.

ylohnitram:
## Review fixes — commit d32a194

Both findings from the code review addressed:

**Blocker — meta-refresh race:** on HTTP 200, `setConfidence` now removes the `<meta http-equiv="refresh" content="60">` tag before showing the toast, mirroring the same pattern the kill-switch handler uses on successful stop (line ~909). The deferred 12 s `location.reload()` is now the only pending reload, so the toast is guaranteed to stay on screen for the full 12 s regardless of when inside the refresh cycle the operator clicked.

**Should-fix — dead JSON payload:** front-end switched from `r.text()` to `r.json()` for 200 responses. The error path still uses `r.text()` because 4xx/5xx return `text/plain`. `showRestartToast` now takes the parsed payload as a third argument and uses it for two things:

1. The headline number is drawn from the server's `data.value` (string, parsed back to float) — that's what actually landed on disk via `agentis memo set`, not just what was clicked. Falls back to the click-site value if the payload is missing a field.
2. The footer switches between "Audit: appended to `.dashboard/confidence-log.jsonl`" and "Audit write failed — restart still required" based on `data.audit_logged`. Stops the toast from claiming audit success when the audit write actually failed.

Verification:

- `bash -n tools/federation-dashboard.sh` — clean.
- `./tools/colony-lint.sh` — 46 passed / 0 failed / 5 skipped (unchanged baseline).
- Manual browser tests (M1–M8 in the original PR body) still deferred to operator.

ylohnitram:
## Re-review — commit d32a194

### Verdict
Approve with nits — both blockers are resolved correctly and no new issues of consequence.

### Findings

**Blockers (from previous review) — resolved**

- *Meta-refresh race (previous blocker):* fixed. Removal happens on the `r.ok` branch, after the error guard, and strictly before `showRestartToast(...)` — no window where a pending 60 s auto-refresh can fire inside the 12 s toast window. Error branch leaves the tag intact, so 400/500 responses still get the page's normal auto-refresh back to a sane state.
- *Dead JSON payload (previous should-fix):* fixed. 200 path runs `r.json()`, 4xx/5xx path runs `r.text()`. `alert('Set failed: ' + result.text)` still fires on errors identically to before. The toast now renders the server's authoritative `data.value` (with a `parseFloat` + `Number.isFinite` fallback to the click-site value — defensive and correct) and branches the footer on `data.audit_logged`.

**Nits (non-blocking)**

1. *JSON parse failure on a 200 response.* If the backend ever returned 200 with a non-JSON body, `r.json()` rejects and the error bubbles to the trailing `.catch(e => alert('Set failed: ' + e))`. Acceptable — but the wording is slightly misleading (the memo *did* land on disk; only the parse failed). Shouldn't happen in practice given the backend always writes `application/json` on the 200 path; flagging only because the alert text would confuse an operator on that theoretical path.

2. *Value precision — off-step values.* The allowlist is agent-name only; `value` is validated as `0.0 <= value <= 1.0` but not snapped to `CONF_STEPS`. In practice the UI only ever POSTs `[0.5, 0.6, 0.85]` so this is moot, but if some other caller POSTed e.g. `0.555`, backend stores `"0.555"`, toast shows `0.56` (`toFixed(2)` rounds). Disk is still authoritative, so it's cosmetic. Not worth fixing unless/until another caller appears.

3. *Audit-failed wording vs visual treatment.* "Audit write failed — restart still required." is truthful, but it's the same visual treatment as the happy path — an operator scanning quickly could miss it. A red/amber border on the toast when `auditOk === false` would make it pop. Tradeoff: adds CSS + conditional class for a rare edge case. I'd lean "not worth it" unless audit failures show up non-trivially in practice — the text alone is fine for now.

4. *Post-removal auto-refresh fallback.* Once `m.remove()` runs, the only remaining reload path is the 12 s `setTimeout` → `fetch('/refresh').catch(() => {}).finally(() => location.reload())`. `.finally` runs even if `.catch` swallowed an error, and `location.reload()` is effectively infallible, so the deferred reload is robust. The only theoretical hole is a heavily-throttled background tab delaying `setTimeout` indefinitely with no 60 s meta-refresh left to save it — browsers do eventually fire queued timers, and this is a foreground operator dashboard, so I'm not asking for a fallback.

### Summary
Both original blockers are fixed cleanly, the parse-per-branch pattern is the right shape, and the fallback logic in `showRestartToast` (parse-then-NaN-check, `audit_logged !== false` triple-check) handles legacy/malformed payloads gracefully. Ship it.

Issue #149, opened 2026-04-15
federation-dashboard: full redesign for operational visibility (learning progress, consistent naming, tooltips, decision aid, federation-down handling)


ylohnitram:
## Summary

The current `federation-dashboard.sh` (the `tools/` web UI for federation observability) is operationally inadequate. Half the data on screen is not actionable, the half that IS actionable is presented inconsistently or buried, and the dashboard provides no signal on the most important operational question: **which agent should be promoted, and why.** This issue captures the operator's full critique and lays out a concrete redesign.

## Operator critique (what's wrong today)

1. **No per-agent learning progress visible.** I cannot see who is learning, how fast, and at what quality. This is the single most important signal for "should I promote this agent". Today I have to ssh into the box, `wc -l .agentis/experience/*.jsonl`, parse JSONL by hand to get `delta`, `outcome`, `cb` aggregates, then map agent_id back to role manually. The dashboard knows where the experience files live; surfacing this is table-stakes.

2. **Inconsistent agent identification across panels.** The health / running-agent panel shows agents by **hash (agent_id)**. The confidence panel shows agents by **role name**. So when I see `risk_assessor: 0.6` and want to ask "is that one healthy?", I have to mentally translate name → hash → look up in the other panel. This is not a UX, this is a cipher.

3. **No tooltips / per-agent detail.** I want to hover an agent and see at minimum:
   - one-line description of what the agent does (currently lives in the `.ag` source comment header — mine that into a tooltip)
   - last 5 experience entries (action, outcome, delta)
   - tick stats (ok / err / success rate / avg cb per tick)
   - confidence + history (when it last changed, by whom, generation)
   - last 3 log lines

   Without this I have to ssh + grep just to know what an agent is supposed to be doing.

4. **Suggestion feeds at the bottom are uninformative.** I look at them and don't understand what's happening. Truncated text, no agent attribution, no timestamps, no priority — just rows. Either fix the formatting and add the missing context, or remove the feed entirely.

5. **No actions other than promote.** I can promote (which is itself broken — see #137 for the auto-restart-after-promote bug). I cannot:
   - demote (revert a bad promote)
   - quarantine an agent (stop it spamming)
   - trigger evolve from the UI
   - acknowledge / dismiss a critical health alert
   - pause the federation
   - configure tick-interval per agent
   These should be one-click actions, not "ssh and `agentis daemon stop <id>`".

[...]

Issue #148, opened 2026-04-15
Auto-promote / auto-evolve: external cron script (layer 1) with roadmap to federation self-governance


ylohnitram:
## Summary

Build a self-governance system that decides when individual agents should be promoted (confidence level up) or evolved (`.ag` source mutation) based on observed fitness signals from `experience/*.jsonl`. Start as an external cron-driven script, evolve into a federation-native supervisor, ultimately into a federation-decides-itself loop.

## Roadmap (3 layers, in order)

The three layers represent increasing levels of agentis-native self-governance. Land them in sequence; each layer is a stepping stone to the next.

### Layer 1 — External cron + script (this issue's primary scope)

A standalone tool living in `agentis-colonies/tools/auto-promote.sh` (or `tools/conductor.py`), invoked by a periodic cron entry. It runs **outside** the federation, reads experience + memo + `agentis daemon list --json`, and decides per-agent whether to promote / evolve.

**Decision rules** (configurable via a YAML/TOML alongside the script):

```yaml
promote:
  prerequisites:
    min_entries: 200             # need enough data
    min_runtime_hours: 48        # don't promote in the first day
    reject_rate_threshold: 0.05  # reject_rate < 5%
    delta_slope_window: 100      # last N entries
    delta_slope_min: 0           # avg delta non-negative over window
  steps:
    - from: 0.5
      to: 0.6
    - from: 0.6
      to: 0.85
evolve:
  trigger:
    delta_slope_negative_for: 1000  # entries
    OR_reject_rate_above: 0.20
  config:
    generations: 3
    population: 4
    weights: cb,val,exp
```

**Actions the script takes** (each gated by per-agent dry-run flag default-on):

- `dashboard promote <agent>` — append to `memo/<role>:confidence.jsonl` AND restart the daemon (using the same per-agent stop+respawn pattern documented in #137 fix).
- `agentis evolve <agent>.ag` — spawn an evolve subprocess, capture results, log to a journal file.

**Output** — append-only journal `tools/auto-promote-journal.jsonl`:

```json
{"ts": 1776280000, "agent": "risk_assessor", "decision": "promote", "from": 0.5, "to": 0.6, "evidence": {"entries": 287, "reject_rate": 0.02, "delta_slope": 0.01}}
```

So you can audit "why was Tom promoted at 03:14" days later.

### Layer 2 — Federation-native supervisor agent (follow-up, separate ticket)

[...]

PR #152, opened 2026-04-16
feat(dashboard): auto-restart daemon on promote (#137 Option 2)

### LINKED ISSUE #137: federation-dashboard: promote writes memo file but running daemons keep stale in-memory value

ylohnitram:
## Summary

Implements **Option 2** from #137 — auto-restart of the target daemon when the operator bumps an agent's confidence in the dashboard, so the new value takes effect on the next click rather than requiring a manual `agentis daemon stop --all && ./start-federation.sh` cycle. Per the user-feedback priority-bump comment on the issue, Option 1 (toast) was "a downgrade in 2026, not a target"; Option 1 wording now survives only as the fallback when the auto-restart request fails outright.

After `agentis memo set` lands on disk, `/confidence` runs:

1. Resolves the agent's daemon from `daemon list --json` by matching `basename(source) == <role>.ag` (#137 criterion 2).
2. `agentis daemon stop <agent_id>` + `kill -0` poll (5 s) → SIGTERM (5 s) → SIGKILL (2 s) fallback — workaround for agentis-core #523's unreliable per-agent stop (criterion 3).
3. Sweeps sidecars under `.agentis/daemon/<agent_id>.{pid,watchdog.pid,colony,heartbeat,status,stop}` and `<agent_id>.inbox/` (criterion 4). Stale `.heartbeat` would otherwise cause the next spawn's watchdog to kill its fresh child on first tick — same root cause `start-federation.sh` wipes after laptop sleep.
4. Respawns detached with the same flags `start-colony.sh` uses (`--colony`, `--enable-exec`, `--enable-messaging`, `--tick-interval 60000`) and env vars (`GITLAB_URL`, `GITLAB_TOKEN`, `GITLAB_PROJECT`, `GITLAB_ME`, `COLONY_DIR`) read from the colony's `colony.toml` (criterion 5).
5. Verifies the new daemon registers with a different pid and fresher `started_at` than before, polling for up to 15 s (criterion 8).

The frontend toast progresses through **Memo written → Stopping daemon → Respawning → Verifying new daemon** with per-step ✓/✗ marks. Terminal state goes green ("Loaded 0.60 ✓ router") on success; red with a paste-ready per-agent respawn command plus the whole-federation fallback on failure (criterion 7). Dismiss timer is armed only at terminal state so a worst-case ~30 s restart never outlasts the toast.

## Changes

- **Single file**: `tools/federation-dashboard.sh` (+599 / −47)
  - New helpers in PYSERVER block: `parse_toml_section`, `list_daemons`, `find_agent_daemon`, `pid_alive`, `wait_for_exit`, `cleanup_sidecars`, `build_manual_command`, `shlex_quote`, `restart_daemon`.
  - `/confidence` handler now returns the structured `restart` object alongside the existing `memo_written` / `audit_logged` fields; `restart_required` now reflects the real outcome.
  - JS: `createProgressToast` replaces the old `showRestartToast`; stepper UI with spinner rows, `applyResult`/`renderHttpError` terminals, `scheduleReload` switches window (12 s success / 30 s failure).
  - CSS: `.toast.toast-ok` (green), `.toast.toast-err` (red), `.toast-step` row styles, `.toast-spin` indicator.

## Test plan

- [x] `bash -n tools/federation-dashboard.sh` — syntax OK
- [x] `./tools/colony-lint.sh` — 46 passed / 0 failed / 5 skipped
- [x] Embedded PYSERVER + PY heredocs parse via `ast.parse`
- [x] Rendered inline JS passes `node --check`
- [x] End-to-end: start full dev-apprenticeship federation, POST `/confidence agent=router&value=0.6` — restart completes in ~2 s, new pid ≠ old pid, `started_at` advances, daemon reads `confidence = 0.6`. Response `restart.succeeded = true` with full event trail.
- [x] Zombie path: router daemon already stopped → `resolve` records `running: false`, stop/wait_exit steps skipped, cleanup + spawn + verify still succeed.
- [x] Error paths unchanged: unknown agent, out-of-range value, non-float value → HTTP 400 with plain-text body.

## Follow-ups

- Once agentis-core #523 lands, the SIGTERM/SIGKILL fallback in `restart_daemon` can be dropped — `agentis daemon stop <id>` will be reliable on its own.
- Once agentis-core #524 lands, the verify step can also re-read `confidence` from `daemon list --json` and assert it matches the disk value, not just that the process came back up (criterion 8's in-memory verification caveat noted in the issue).

Closes #137 (Option 2). Foundational for #148 and #149.

### DIFF
--- a/tools/federation-dashboard.sh
+++ b/tools/federation-dashboard.sh
@@ -1163,7 +1163,7 @@ trap 'kill $REGEN_PID 2>/dev/null; exit 0' INT TERM
 # with 400 (never pass user strings to subprocess unchecked, #105).
 ALL_AGENTS_CSV="$(IFS=,; echo "${ALL_AGENTS[*]}")"
 python3 - "$DASH_DIR" "$PORT" "$SCRIPT_PATH" "$FED_DIR" "$ALL_AGENTS_CSV" "$AGENT_COLONY_MAP" <<'PYSERVER'
-import sys, os, subprocess, json, time, signal, shutil, shlex, urllib.parse
+import sys, os, subprocess, json, time, signal, shutil, shlex, threading, urllib.parse
 from http.server import HTTPServer, SimpleHTTPRequestHandler
 
 serve_dir, port = sys.argv[1], int(sys.argv[2])
@@ -1370,13 +1370,24 @@ def restart_daemon(agent):
             'events': events,
         }
 
-    # Read gitlab config up front so a malformed colony.toml fails fast
-    # before we stop the running daemon.
+    # Read gitlab config before touching the daemon — a missing or
+    # incomplete colony.toml must bail out before we stop anything,
+    # otherwise the operator ends up with a killed daemon and no
+    # respawn (parse_toml_section returns {} on missing file).
     config_path = os.path.join(colony_dir, 'config', 'colony.toml')
     gitlab = parse_toml_section(config_path, 'gitlab')
     gl_url = gitlab.get('url', '')
     gl_token = gitlab.get('token', '')
     gl_project_raw = gitlab.get('project', '')
+    if not gl_url or not gl_token or not gl_project_raw:
+        rec('config', 'error',
+            message=f'incomplete [gitlab] in {config_path}: '
+                    f'url={bool(gl_url)} token={bool(gl_token)} project={bool(gl_project_raw)}')
+        return {
+            'attempted': False, 'succeeded': False,
+            'error': f'incomplete [gitlab] section in {config_path}',
+            'events': events,
+        }
     gl_me = gitlab.get('me', '')
     gl_project = gl_project_raw.replace('/', '%2F')
     env_overrides = {
@@ -1506,6 +1517,7 @@ def restart_daemon(agent):
 
     if not new_daemon:
         code = proc.poll()
+        threading.Thread(target=proc.wait, daemon=True).start()
         rec('verify', 'error',
             message=f'new daemon not registered within 15s (launcher exit={code})')
         return {
@@ -1518,10 +1530,11 @@ def restart_daemon(agent):
     # Reap the launcher wrapper so it does not linger as a zombie in the
     # dashboard's process table. The actual daemon child is already
     # detached (start_new_session=True) and outlives the launcher.
-    try:
-        proc.wait(timeout=2)
-    except subprocess.TimeoutExpired:
-        pass
+    # Fire-and-forget thread: if the launcher is still running (rare —
+    # it normally exits once the watchdog forks), the thread blocks
+    # until it does, keeping the process table clean without stalling
+    # the HTTP response.
+    threading.Thread(target=proc.wait, daemon=True).start()
 
     rec('verify', 'ok',
         new_pid=new_daemon.get('pid'),
--- a/tools/federation-dashboard.sh
+++ b/tools/federation-dashboard.sh
@@ -715,9 +715,13 @@ function setConfidence(agent, value) {
   // single-threaded and SSE adds more moving parts than this operator
   // tool warrants). If the real response lands before a given timer, the
   // real terminal state wins — see applyResult() below.
+  // Timers spread to match the server worst case: stop can take up to
+  // 12 s (5 s poll + 5 s SIGTERM + 2 s SIGKILL), spawn ~1 s, verify up
+  // to 15 s. We advance optimistically; the real terminal state wins on
+  // server response regardless.
   const t1 = setTimeout(() => toast.setStep('stop', 'active'), 400);
-  const t2 = setTimeout(() => { toast.setStep('stop', 'done'); toast.setStep('spawn', 'active'); }, 3500);
-  const t3 = setTimeout(() => { toast.setStep('spawn', 'done'); toast.setStep('verify', 'active'); }, 8000);
+  const t2 = setTimeout(() => { toast.setStep('stop', 'done'); toast.setStep('spawn', 'active'); }, 12000);
+  const t3 = setTimeout(() => { toast.setStep('spawn', 'done'); toast.setStep('verify', 'active'); }, 15000);
   const clearFakes = () => { clearTimeout(t1); clearTimeout(t2); clearTimeout(t3); };
 
   const body = 'agent=' + encodeURIComponent(agent) + '&value=' + encodeURIComponent(value);
@@ -1159,7 +1163,7 @@ trap 'kill $REGEN_PID 2>/dev/null; exit 0' INT TERM
 # with 400 (never pass user strings to subprocess unchecked, #105).
 ALL_AGENTS_CSV="$(IFS=,; echo "${ALL_AGENTS[*]}")"
 python3 - "$DASH_DIR" "$PORT" "$SCRIPT_PATH" "$FED_DIR" "$ALL_AGENTS_CSV" "$AGENT_COLONY_MAP" <<'PYSERVER'
-import sys, os, subprocess, json, time, signal, shutil, urllib.parse
+import sys, os, subprocess, json, time, signal, shutil, shlex, urllib.parse
 from http.server import HTTPServer, SimpleHTTPRequestHandler
 
 serve_dir, port = sys.argv[1], int(sys.argv[2])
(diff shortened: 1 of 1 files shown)


ylohnitram:
## QA Review — REQUEST CHANGES

**Reviewer:** coordinator session (agentis-pr-reviewer agent)

### CI Status
Pending verification (review ran before CI completed).

| # | File | Lines | Issue |
|---|------|-------|-------|
| 1 | `tools/federation-dashboard.sh` | 1316-1328, 1532 | **GITLAB_TOKEN exposed to browser.** `build_manual_command()` embeds `GITLAB_TOKEN=<actual_token>` into `manual_cmd` string returned in JSON response. On failure path, token is rendered into DOM via `pre.textContent` — visible on screen, DevTools, browser cache. **Fix:** Replace `GITLAB_TOKEN=<value>` with `GITLAB_TOKEN=\$GITLAB_TOKEN` in `env_overrides` when building `manual_cmd` (operator's shell already has it exported via `start-colony.sh`). Alternatively, redact to `GITLAB_TOKEN=glpat-****` or omit `manual_cmd` from success responses entirely. |

### Suggestions (non-blocking)

| # | File | Lines | Suggestion |
|---|------|-------|------------|
| 1 | `tools/federation-dashboard.sh` | 1332-1339 | Use `shlex.quote` from stdlib instead of custom reimplementation. `shlex` is already available, the custom impl adds unnecessary quoting edge-case surface. |
| 2 | `tools/federation-dashboard.sh` | 1455-1472 | Zombie `Popen` object on success path — never `wait()`ed. Consider `proc.poll()` or context manager to avoid zombie process table entries. |
| 3 | `tools/federation-dashboard.sh` | 719-721 | Fake progress timers (3.5s stop, 4.5s spawn) don't match server worst-case (12s stop via SIGTERM+SIGKILL). Consider adjusting to ~12s/~20s. |

### Highlights
- Five-phase sequence (resolve → stop/wait/SIGTERM/SIGKILL → cleanup → spawn → verify) is correct
- #523 workaround escalation thorough (daemon stop → 5s poll → SIGTERM → 5s poll → SIGKILL → 2s poll)
- Sidecar cleanup covers all 6 extensions + `.inbox/`
- Spawn flags exactly match all five `start-colony.sh` scripts
- Agent names validated against `allowed_agents` allowlist, values validated as float in [0,1]
- `subprocess.Popen` uses list args (no shell=True), DOM uses `textContent` (no innerHTML)
- `start_new_session=True` correctly detaches daemon from dashboard process
- Verify step checks `new_pid != old_pid` AND `started_at >= spawn_ts - 1`
- Config read from `colony.toml` before stopping (fail-fast)
- Error toast includes both per-agent manual command AND whole-federation fallback

**Verdict: REQUEST CHANGES — fix #1 (token leak) before merge. Rest is non-blocking.**

ylohnitram:
## QA Review (post-fix) — APPROVE

**Reviewer:** coordinator session (agentis-pr-reviewer agent, second pass)

### CI Status
- **colony-lint**: PASS (46 passed, 0 failed, 5 skipped)

### Critical Issues

None found. The GITLAB_TOKEN leak flagged in the first review is fully resolved.

### Suggestions (non-blocking)

| # | File | Lines | Suggestion |
|---|------|-------|------------|
| 1 | `tools/federation-dashboard.sh` | ~1373-1376 | **Misleading "fails fast" comment.** `parse_toml_section` silently returns `{}` on a missing file. If `colony.toml` is absent, the code proceeds to stop the old daemon and spawn a new one with empty `GITLAB_URL`/`GITLAB_TOKEN`/etc — the new daemon would register (verify succeeds) but fail on its first tick. Consider an early-return check (`if not gl_url or not gl_token: rec('config', 'error', ...); return`) before the stop block, or at minimum update the comment. |
| 2 | `tools/federation-dashboard.sh` | ~1518-1524 | **Unreachable zombie on launcher timeout.** If `proc.wait(timeout=2)` raises `TimeoutExpired`, the Popen object is dropped without ever reaping the child. For an operator tool that typically runs one or two restarts per session this is negligible; a `threading.Thread(target=proc.wait).start()` would eliminate it entirely. |
| 3 | `tools/federation-dashboard.sh` | ~68 | **AGENT_COLONY_MAP built via bash string concat.** If a future colony or agent name contains a double-quote or backslash, the resulting JSON will be malformed and the Python `json.loads` fallback (empty map) will silently disable all restarts. Current names are safe. The existing `try/except` on the parse handles the failure gracefully, so this is defense-in-depth. |
| 4 | `tools/federation-dashboard.sh` | ~1408-1410 | **`stdout`/`stderr` from `agentis daemon stop` included in events trail.** These are returned in the JSON response body (visible in DevTools). Currently safe — `daemon stop` only prints "stop signal sent: \<id\>" to stderr — but if a future agentis version ever prints config details in stop output, they would leak. Worth a note. |

### Verification of first-review fixes

| Fix | Status |
|-----|--------|
| **GITLAB_TOKEN leak** | ✅ Fixed. `build_manual_command()` emits `$GITLAB_TOKEN` (shell var reference), not the expanded value. Real token lives only in `env_overrides` passed to `Popen(env=...)`, never serialized to JSON. Verified across success and failure response paths. |
| **stdlib `shlex.quote`** | ✅ Custom `shlex_quote` removed, all call sites use `shlex.quote`. |
| **Zombie Popen** | ✅ `proc.wait(timeout=2)` added after verify succeeds. Minor residual in timeout path (suggestion #2 above). |
| **Fake-progress timers** | ✅ Timers spread to 12 s / 15 s matching worst-case SIGTERM+SIGKILL + verify poll. |

### Highlights

- **No XSS surface.** All new DOM writes use `textContent` and `createElement`/`appendChild`. No `innerHTML` from server-supplied data anywhere in new code.
- **Robust stop escalation.** `daemon stop` → 5 s `kill -0` poll → SIGTERM → 5 s poll → SIGKILL → 2 s poll. Each `os.kill` wrapped in try/except. "Survived SIGKILL" bail-out with manual command fallback.
- **Timer race handled.** `clearFakes()` called on both success and error paths before applying terminal state. Dismiss timer armed only at terminal.
- **Clean Popen lifecycle.** `start_new_session=True` detaches daemon, `stdin/stdout/stderr=DEVNULL` prevents fd leaks.
- **Config read before stop.** Colony TOML is parsed before the daemon is stopped — a bad config path doesn't leave the operator with a killed daemon and no respawn.
- **Good event trail.** `rec()` helper provides step+status+kwargs diagnostic log for both the stepper UI and raw JSON debugging.

**Verdict: APPROVE — no critical issues remain. Suggestions are non-blocking quality improvements for a follow-up.**

PR #190, opened 2026-04-18
fix(planning): rate-limit observe step to 30 min to cut LLM waste (#187)

### LINKED ISSUE #187: planning agents burn ~4k LLM calls/day on observe-step with act branch that almost never fires on low-activity projects
## Summary

The three planning agents (`risk_assessor`, `scope_estimator`, `task_decomposer`) run an **unconditional `prompt()` observe step every tick** while their **tier-gated acting branch almost never fires** on low-activity projects. On dev-apprenticeship targeting X/Y1 that works out to roughly **4 000+ wasted LLM calls per day** — pattern-learning that no downstream branch ever consumes.

## Anatomy of the waste (using `risk_assessor.ag` as the reference)

Every tick the agent does two independent GitLab polls with very different behaviour:

| Step | Where | Input | Guard | Effect |
|---|---|---|---|---|
| Observe | `risk_assessor.ag:44-65` | `gitlab-api.sh issues --view planning` (entire recent feed) | `len(recent_raw) > 10` (~always true) | `prompt(...)` → learn `"observed"` |
| Act | `risk_assessor.ag:69-140` | `gitlab-api.sh issues --needs-planning --since <last_check>` (delta) | `len(issues_raw) >= 3` (rarely satisfied) | `prompt(...)` → tier-branch → learn `"emitted"` / `"review-gated"` / `"acted"` |

The observe step runs **every single tick**. The act step runs **only when a new unplanned issue has appeared since the previous tick**. On a quiet project like Y1 (few new unplanned issues per day) the act step effectively never runs.

`scope_estimator` and `task_decomposer` follow the same shape.

## Evidence from experience store

As of today (dev-apprenticeship on v1.4.0, federation running 10+ minutes since most recent promote):

| Agent | Total rows | Tagged `observed` | Tagged `review-gated` / `acted` / `emitted` |
|---|---:|---:|---:|
| risk_assessor    | 425 | 425 | **0** |
| scope_estimator  | 288 | 288 | **0** |
| task_decomposer  | 426 | 426 | **0** |

Every single row is from the observe step. Zero rows from any acting branch, in any tier the agents have ever been at. The acting branch has not fired once in the entire history of the federation.

## Is the observed pattern library used?

Barely. Each agent's `recommend()` call feeds the acting prompt context (`risk_assessor.ag:81`). `recommend()` reads from the observed-pattern library. But the call site is **inside the acting branch** — so the pattern library is only consulted when the acting branch fires, which it almost never does. The observe step is learning patterns that feed a consumer that never runs.

## Scale of the waste

Tick interval (planning colony, per `start-colony.sh`): `60000ms` = 60 seconds.

Per agent per tick: 1 `prompt()` call (the observe step) is virtually guaranteed to fire; the second `prompt()` (act step) almost never fires on low-activity projects.

Per agent per day: 60 min × 24 h = 1 440 observe prompts.
Three agents: **~4 320 LLM calls/day** on the observe step alone.

On any project where new unplanned issues do not arrive every ~60 seconds — which is essentially every real project smaller than X/Y2 — this is pure waste.

## Why the observe step is shaped this way

The observe step was presumably intended to let the agent build a pattern library continuously, so that when the acting branch eventually fires it has rich context to condition on. That intent is sound. The problem is the shape:

- **Observe is unconditional** (any recent issue feed triggers it) while **act is strict-delta-gated** (only new unplanned issues since last tick). The two should be on comparable activity gates so they fire together.
- **Observe consumes LLM budget** proportional to tick frequency, not to project activity. On idle federations observe behaves as a constant-rate heartbeat LLM call.
- **The patterns observe produces are never read except by act**, which rarely runs.

### A. Rate-limit observe relative to act-branch activity
Only run the observe step every Nth tick (e.g. 1 tick in 10), OR only when the act branch has fired at least once in the past M ticks. On a project that does not exercise the act branch, the observe step gradually backs off instead of burning a fixed LLM budget forever.

### B. Couple observe to the same delta gate as act
Run observe only when the act branch is also going to run (same `len(issues_raw) >= 3` precondition, or a superset of it). This makes observe a prelude to act rather than an independent heartbeat.

### C. Make observe conditional on recommend-library staleness
Skip observe when the most recent `observed` row is newer than X minutes (configurable). Idempotent library refresh instead of fixed-rate relearn.

### D. Cache the observe output in memo and re-read across ticks
The current shape re-invokes LLM each tick even if the recent feed has not changed. Hash the input, cache the `IncidentPatterns` struct keyed on that hash, and skip the `prompt()` when the cache is fresh. This is the most mechanical fix — no semantic change, pure de-duplication.

### E. Tie observe to external triggers
Observe only when `triage:new_issue` or a new issue comment hits the bus, not on a wall-clock tick. Reactive rather than polling.

(A) and (D) are cheapest; (E) is most structural.

## Scenario-shape consideration

This issue affects the planning trio specifically because they combine "unconditional observe" with "strict-delta act." Other agents in the federation that I have read (`security_reviewer`, `approval_decider`, `ship_decider`, `release_checker`) front-load their `tick()` with a single early-return guard and hit `prompt()` only when there is real work — they do not burn LLM on observe during idle periods. So the waste is concentrated in the planning colony. A fix applied only to those three `.ag` files is already high-value.

## Impact

On a federation targeting an internal, low-activity GitLab project, the current shape means ~4 000 LLM calls/day with no visible output and no downstream consumer. Across a week that is ~30 000 calls, which at current Opus pricing is real money for no return. Beyond the direct cost it also muddies auto-promote signal (see #186 — heuristic already can't tell observe rows apart from acting rows; reducing observe-row volume makes the signal cleaner even if auto-promote is fixed independently).

ylohnitram:
## Summary

- The three planning agents (`risk_assessor`, `scope_estimator`, `task_decomposer`) used to run an **unconditional `prompt()` observe step every tick** — ~1 440 LLM calls per agent per day at the default 60 s tick interval. The pattern library they built only feeds `recommend()` inside the act branch, which rarely fires on low-activity projects, so the work went almost entirely unused (see issue evidence: 425/288/426 observed rows with 0 emitted/review-gated/acted rows).
- Adds a **30-min staleness gate** on the observe step in each of the three `.ag` files. The observe prompt runs only when `now - last_observe_ts >= 1800 s`; otherwise the tick proceeds straight to the act branch. Fresh observes stamp `memo_write("<agent>:observe_last_ts", now_epoch)`.
- **Fail-open** by design: if `date -u +%s` fails, `last_obs` is empty, or the observe produces zero patterns, the next tick re-runs observe — no starvation of the pattern library.

## Expected impact

| Metric                                        | Before                 | After                 |
|-----------------------------------------------|------------------------|-----------------------|
| Observe LLM calls / agent / day               | 1 440                  | ~48                   |
| Across the planning trio / day                | ~4 320                 | ~144                  |
| Reduction                                     | —                      | **~97 %**             |

Act-branch cadence is unchanged — still gated on `len(issues_raw) >= 3`. The fix is purely about stopping heartbeat-rate observe prompts on idle projects.

Fixes #187.

## Test plan

- [x] `./tools/colony-lint.sh` — 71 passed, 0 failed, 5 skipped.
- [x] `agentis commit <agent>.ag` — all three agents parse cleanly (syntax validated individually).
- [ ] Live verification on dev-apprenticeship next time the federation is running: confirm the `:observe_last_ts` memo lands on the first observed tick and that subsequent ticks within 30 min skip the observe prompt.

## Design notes

- `let observe_stale = if len(last_obs) > 0 { ... } else { 1; };` — `.ag` has immutable `let` bindings, so the gate is an if-expression rather than a mutable flag. Each branch terminates its last expression with `;` per `.ag` block syntax.
- Interval is hardcoded to 1800 s rather than read from colony.toml. The issue lists configurability as "(configurable)" but a hardcoded sensible default is enough to resolve the waste; a future PR can promote it to a knob if operators want to tune it.
- No change to the act-branch shape. No change to tier gating. No change to event wiring.

### DIFF
--- a/dev-apprenticeship/planning/agents/risk_assessor.ag
+++ b/dev-apprenticeship/planning/agents/risk_assessor.ag
@@ -50,8 +50,14 @@ fn tick(reason: string) -> void {
     let last_obs = recall_latest("risk_assessor:observe_last_ts");
     let observe_stale = if len(last_obs) > 0 {
         if len(now_epoch) > 0 {
+            // Negative delta (operator ran `date -s <past>` or similar
+            // wall-clock regression) — treat as stale and re-observe
+            // rather than skipping forever on a clock that never catches
+            // up.
             let delta = parse_int(now_epoch) - parse_int(last_obs);
-            if delta < 1800 { 0; } else { 1; };
+            if delta >= 0 {
+                if delta < 1800 { 0; } else { 1; };
+            } else { 1; };
         } else { 1; };
     } else { 1; };
 
--- a/dev-apprenticeship/planning/agents/risk_assessor.ag
+++ b/dev-apprenticeship/planning/agents/risk_assessor.ag
@@ -39,29 +39,50 @@ fn get_confidence() -> float {
 }
 
 fn tick(reason: string) -> void {
+    // Staleness gate for the observe step (#187): the pattern library
+    // only feeds `recommend()` inside the act branch below, which rarely
+    // fires on low-activity projects. Re-running the observe prompt on
+    // every 60 s tick burns ~1 440 LLM calls/day for a library that may
+    // never be read. Skip the observe prompt unless the last successful
+    // observe was >= OBSERVE_MIN_INTERVAL_SEC ago. 1 800 s (30 min) keeps
+    // the library fresh without the heartbeat-LLM pathology.
+    let now_epoch = try { exec sh "date -u +%s"; } catch e { ""; };
+    let last_obs = recall_latest("risk_assessor:observe_last_ts");
+    let observe_stale = if len(last_obs) > 0 {
+        if len(now_epoch) > 0 {
+            let delta = parse_int(now_epoch) - parse_int(last_obs);
+            if delta < 1800 { 0; } else { 1; };
+        } else { 1; };
+    } else { 1; };
+
     // 1. Observe: pull recent opened issues and skim for incident-style patterns
-    let recent_raw = try {
-        exec sh "$COLONY_DIR/scripts/gitlab-api.sh issues --view planning";
-    } catch e {
-        print("[risk_assessor] GitLab issues poll failed:", e);
-        "";
-    };
+    if observe_stale == 1 {
+        let recent_raw = try {
+            exec sh "$COLONY_DIR/scripts/gitlab-api.sh issues --view planning";
+        } catch e {
+            print("[risk_assessor] GitLab issues poll failed:", e);
+            "";
+        };
 
-    if len(recent_raw) > 10 {
-        let patterns = prompt(
-            "Analyze this recent GitLab issues JSON. Focus on issues labeled 'incident', 'bug', or 'blocker'. What recurring themes appear (dependency surprises, integration gotchas, flaky services)? Return JSON: observed_count (int, how many incident-style issues you saw), recurring_themes (string, one short paragraph).",
-            recent_raw
-        ) -> IncidentPatterns;
+        if len(recent_raw) > 10 {
+            let patterns = prompt(
+                "Analyze this recent GitLab issues JSON. Focus on issues labeled 'incident', 'bug', or 'blocker'. What recurring themes appear (dependency surprises, integration gotchas, flaky services)? Return JSON: observed_count (int, how many incident-style issues you saw), recurring_themes (string, one short paragraph).",
+                recent_raw
+            ) -> IncidentPatterns;
 
-        if patterns.observed_count > 0 {
-            learn(
-                "risk_assessment",
-                "batch of " + to_string(patterns.observed_count) + " incident-style issues",
-                patterns.recurring_themes,
-                "success",
-                ["observed", "planning"]
-            );
-            print("[risk_assessor] Observed", patterns.observed_count, "incident-style issues");
+            if patterns.observed_count > 0 {
+                learn(
+                    "risk_assessment",
+                    "batch of " + to_string(patterns.observed_count) + " incident-style issues",
+                    patterns.recurring_themes,
+                    "success",
+                    ["observed", "planning"]
+                );
+                if len(now_epoch) > 0 {
+                    memo_write("risk_assessor:observe_last_ts", now_epoch);
+                };
+                print("[risk_assessor] Observed", patterns.observed_count, "incident-style issues");
+            };
         };
     };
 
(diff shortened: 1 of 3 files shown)


ylohnitram:
## QA Report (agentis-qa-agent)

Ship-it verdict with 14/14 checks passing, 0 defects.

### Summary
- `colony-lint.sh`: **71 passed, 0 failed, 5 skipped**
- `agentis commit` syntax: all three files pass (verified via colony-lint's syntax stage)
- Adjacent tests: `test-dashboard-gates` 22/22, `test-gitlab-views` 48/48
- Private path leaks: **none**

### Verification checklist

| # | Check | Result |
|---|---|---|
| 1 | `colony-lint.sh` ends with 0 failures | PASS (71/0/5) |
| 2 | `risk_assessor`: observe block gated on `if observe_stale == 1` | PASS |
| 3 | `risk_assessor`: `memo_write(...observe_last_ts)` only inside `observed_count > 0` | PASS |
| 4 | `scope_estimator`: observe block gated on `if observe_stale == 1` | PASS |
| 5 | `scope_estimator`: `memo_write(...observe_last_ts)` only inside `observed_pairs > 0` | PASS |
| 6 | `task_decomposer`: observe block gated on `if observe_stale == 1` | PASS |
| 7 | `task_decomposer`: `memo_write(...observe_last_ts)` only inside `epics_seen > 0` | PASS |
| 8 | Act branch (`if len(issues_raw) < 3 { return; }` and after) unchanged, outside observe gate | PASS |
| 9 | `memo_write("<agent>:last_check", now)` still fires after the act path | PASS |
| 10 | Three-way `tier()` calls (autonomous / review-gated / propose / else) untouched | PASS |
| 11 | Fail-open: if `now_epoch == ""`, `observe_stale == 1` via the `else { 1; }` branch | PASS |
| 12 | 1800-second threshold consistent across all three agents | PASS |
| 13 | No private absolute paths in the diff | PASS |
| 14 | `observe_last_ts` memo key appears only in the three touched agents (no collision) | PASS |

### Defects

None.

### Nits (informational, non-blocking)

- **LOW**: Fail-open semantics depend on a subtle detail — when `last_obs == ""` staleness is `1` regardless of `now_epoch`; when `last_obs` is set but `now_epoch == ""` the outer `else { 1; }` also resolves to `1`. Edge: if `last_obs` is set and wall-clock jumps backwards so `now_epoch < last_obs`, `delta < 1800` stays true and observe is skipped. Symmetric with forward skew; tiny concern.
- **LOW**: The staleness gate (~7 lines) is duplicated across the three agents. `.ag` has no cross-file import, so sharing via a helper is not ergonomic today; acceptable copy-paste.

### Verdict

Ship it. All structural and semantic checks match the PR description; fail-open, threshold consistency, memo-key uniqueness, tier-gating preservation, and `last_check` cursor continuity are all intact.

Issue #227, opened 2026-04-20
planning peers (risk_assessor/scope_estimator/task_decomposer) missing add-note idempotency guard


ylohnitram:
## Problem

Follow-up from #223, which identified the missing idempotency guard in `plan_reviewer.ag`. The same shape applies to the three peer agents that also post GitLab comments from their autonomous branch:

- `dev-apprenticeship/planning/agents/risk_assessor.ag:124` — `gitlab-api.sh add-note $issue_id --body ...`
- `dev-apprenticeship/planning/agents/scope_estimator.ag:121` — same
- `dev-apprenticeship/planning/agents/task_decomposer.ag:121` — same

Each of these three agents triggers on the planning colony's label filter (currently hardcoded to `needs-planning`, see #223 for the trigger decoupling). Once the trigger label is fixed to match a real-world label convention (e.g. `DEV::not started` in X), the long-lived nature of real workflow labels creates the same spam risk described in #223: the agent ticks every 60s, sees the same issue still in scope, calls `prompt()` to re-generate the assessment, and posts another comment.

## Partial insulation, not a full fix

The three peers have `observe_last_ts` memo that the `plan_reviewer` doesn't. But `observe_last_ts` is a *tick gate* on "have I observed since the last issue-update timestamp", not a *post gate* on "have I already posted this exact assessment to this issue". So on a project with actively-discussed issues (where `updated_at` advances on every comment), the guard re-opens within one tick cycle.

The concrete symptom on a persistent label would be: N comments from each of the three peers, then N more on the next tick when the issue's `updated_at` advances because *those comments themselves* advanced it. Recursive storm.

## Proposed fix

Same pattern as the `plan_reviewer` fix in #223:

```
// in each peer, before prompt() / add-note path:
let iid_str = to_string(target_iid);
let posted_key = "risk_assessor:" + iid_str + ":posted";  // / scope_estimator / task_decomposer
if len(recall_latest(posted_key)) > 0 {
    return;
};
// ...existing prompt + add-note + learn...
memo_write(posted_key, "1");
```

Three files, one-memo-key-per-agent. Memo key should include the agent role to avoid collisions if two peers ever assess the same issue independently.

Optional: revisit whether `observe_last_ts` should be retained as well. If the `:posted` marker is authoritative, `observe_last_ts` becomes redundant. Probably worth a secondary pass.

## Why not fold this into #223

#223 scopes to `plan_reviewer` + trigger label. Splitting this out because:
[...]

PR #228, opened 2026-04-20
fix(planning): configurable trigger label + plan_reviewer idempotency guard (#223)

### LINKED ISSUE #223: planning colony: hardcoded 'needs-planning' trigger + missing idempotency guard in plan_reviewer
## Problem

The planning colony uses a single hardcoded GitLab label as the trigger for all four agents' work: `needs-planning`.

`dev-apprenticeship/planning/scripts/gitlab-api.sh`:

```bash
if [ "$NEEDS_PLANNING" -eq 1 ]; then
    ARGS+=(--data-urlencode "labels=needs-planning")
fi
```

Every planning agent (`risk_assessor`, `scope_estimator`, `task_decomposer`, `plan_reviewer`) calls `gitlab-api.sh issues --needs-planning --view planning` and relies on the returned JSON being non-empty to proceed. If the target GitLab project uses a different label convention, the whole colony silently does nothing:

```
$ gitlab-api.sh issues --needs-planning --view planning
[]
```

This is what happens on a real-world project I'm running against. The project uses GitLab scoped labels like `DEV::not started`, `DEV::in progress`, `DEV::blocked`, `DEV::merged` to track the implementation lifecycle. There is no `needs-planning` label, and flat kebab-case doesn't match the project's `X::Y` convention, so adding one would be a foreign body in the label taxonomy.

In a 45-minute session all four planning agents were healthy and ticking (~30 `tick_ok`, 0 errors), but `plan_reviewer` produced zero log entries since startup, accumulated zero acting entries, and has no path forward under the current auto-promote prereqs (`min_acting_entries: 60` for 0.6 → 0.8). Meanwhile the three peers got promoted to `0.95` autonomous because their promotion path uses observation-only experience, not acting experience — so the planning colony drifts into a split state where the emitters graduate but the assembler can't.

### 1. Make the trigger label operator-configurable

`gitlab-api.sh` should read the trigger label from a flag and/or `colony.toml`, with `needs-planning` as the default for backward compatibility. Something like:

```toml
# planning/config/colony.toml
[planning]
trigger_label = "DEV::not started"   # default: "needs-planning"
```

Scoped labels (`::`) are valid in GitLab queries and `curl --data-urlencode` handles the encoding already, so no new encoding logic is needed. Agents would either keep calling `--needs-planning` (default stays as-is) or be updated to pass the configured label explicitly.

### 2. Add an idempotency guard to `plan_reviewer.ag`

Even with the right label, there is a design gap: **what if the trigger label is long-lived?**

The original `needs-planning` is presumably short-lived (plan gets posted → someone removes the label → colony moves on). But for real-world workflows where the trigger label persists until a human action (e.g. `DEV::not started` stays until someone starts coding, which can be days), `plan_reviewer` will:

- Tick every 60s
- Find the same issue still labeled
- Call `prompt()` to assemble a fresh plan (LLM cost per tick)
- Post a fresh comment via `gitlab-api.sh add-note` (`plan_reviewer.ag:143`)
- …forever

Inspecting `plan_reviewer.ag` shows no guard. Memo keys in use:

```
plan_reviewer:{issue_id}:scope       # peer output stash
plan_reviewer:{issue_id}:risks       # peer output stash
plan_reviewer:{issue_id}:breakdown   # peer output stash
plan_reviewer:last_check             # liveness
plan_reviewer:confidence             # tier memo
```

No `plan_reviewer:{issue_id}:posted` marker. Adding one is small:

```
let iid_str = to_string(target_iid);
if len(recall_latest("plan_reviewer:" + iid_str + ":posted")) > 0 {
    return;
};
// ...existing assembly / prompt / add-note / learn...
memo_write("plan_reviewer:" + iid_str + ":posted", "1");
```

The same concern likely applies to the three peers' `add-note` paths (`risk_assessor.ag:124`, `scope_estimator.ag:121`, `task_decomposer.ag:121`), though they may be partially insulated by their `observe_last_ts` memo. Worth a secondary pass.

## Why both changes together

A label-only fix without the idempotency guard is worse than the current status quo: it turns silent dormancy into a per-minute comment spam stream. A guard-only fix doesn't help because the trigger label still never matches in the target project.

## Context

Reproduced on `dev-apprenticeship v0.1.0` + agentis `v1.4.3`, macOS (arm64).

- 21/21 agents spawn cleanly
- `risk_assessor`, `scope_estimator`, `task_decomposer` auto-promoted to `0.95` (autonomous) after ~14 minutes
- `plan_reviewer` stuck at `0.6` (propose) indefinitely
- Zero actual LLM calls wasted (early-exit before `prompt()`), but zero useful work either

ylohnitram:
## Summary

Closes #223. Ships the two coupled changes the issue proposes:

1. **Configurable planning trigger label** — new `[planning] trigger_label` key in `planning/config/colony.example.toml`. Default preserves pre-#223 behavior; operators on projects using scoped labels (e.g. `DEV::not started`) can point the planning colony at the local label without touching agent code.
2. **`plan_reviewer` idempotency guard** — new `plan_reviewer:<iid>:posted` memo marker short-circuits `prompt()` + `add-note` once the plan is posted for an issue.

Peer-agent idempotency (`risk_assessor`/`scope_estimator`/`task_decomposer`) is **out of scope** here — that's the #227 follow-up, shipped separately per the issue's "why not fold" rationale.

## Design notes

- **Script-layer, not agent-layer**: the trigger-label substitution lives in `gitlab-api.sh` (`labels=${PLANNING_TRIGGER_LABEL:-needs-planning}`). All 4 planning agents keep calling `--needs-planning` verbatim, so zero `.ag` diff for the label decoupling. `--data-urlencode` already handles scoped labels and spaces.
- **Marker only in posting branches**: the idempotency guard writes the marker in the `autonomous` and `review-gated` branches (both post a GitLab note). `propose` emits on the bus and `shadow` only `learn()`s — those don't post, so marking them would prevent legitimate future posts if the agent is promoted.
- **Satisfies `check-prompt-gate`** — the new `recall_latest(posted_key)` is a memo gate directly covering the `prompt()` on its hot path.

## Backward compatibility

- Configs without `[planning] trigger_label`: `parse_toml` returns empty → `${:-needs-planning}` fallback → byte-identical to today.
- Agents keep the literal `--needs-planning` flag. No migration for callers.
- New memo key (`plan_reviewer:<iid>:posted`) — no collision with existing keys.

## Semver

**MINOR** — new optional config key per CLAUDE.md classification.

## Test plan

- [x] `./tools/colony-lint.sh` — 80 passed, 0 failed, 5 skipped (`shellcheck` / `agentis` unavailable on this host).
- [x] `bash -n` on both modified scripts.
- [ ] Live smoke test (post-merge): point a staging planning colony at a GitLab project with a scoped trigger label and verify (a) issues flow through, (b) second tick short-circuits before `prompt()` after the plan is posted.

## Files touched

- `dev-apprenticeship/planning/config/colony.example.toml` — new `[planning]` section with `trigger_label` key.
- `dev-apprenticeship/planning/scripts/start-colony.sh` — parse + export `PLANNING_TRIGGER_LABEL`.
- `dev-apprenticeship/planning/scripts/gitlab-api.sh` — env-var resolved default in `--needs-planning` branch; header doc updated.
- `dev-apprenticeship/planning/agents/plan_reviewer.ag` — idempotency guard before prompt; marker write in both posting branches.
- `dev-apprenticeship/CHANGELOG.md` — `### Added` entry for the config key, `### Fixed` entry for the guard.

Plan: https://github.com/Replikanti/agentis-colonies/issues/223#issuecomment-4279905868

### DIFF
--- a/dev-apprenticeship/CHANGELOG.md
+++ b/dev-apprenticeship/CHANGELOG.md
@@ -17,6 +17,15 @@ is asserted until multi-version CI is in place.
 
 ### Added
 
+- **Configurable planning trigger label** — new `[planning] trigger_label`
+  key in `planning/config/colony.example.toml`. Operators on projects
+  that don't use a flat `needs-planning` label (e.g. scoped-label
+  taxonomies like `DEV::not started`) can point the planning colony at
+  the local label without edits to the 4 agent files. Default preserves
+  pre-#223 behavior. `--data-urlencode` handles scoped labels and spaces
+  at the API layer, no new encoding logic required.
+  ([#223](https://github.com/Replikanti/agentis-colonies/issues/223))
+
 ### Changed
 
 ### Deprecated
@@ -24,6 +33,15 @@ is asserted until multi-version CI is in place.
 ### Removed
 
 ### Fixed
+
+- **`plan_reviewer` idempotency guard** — new `plan_reviewer:<iid>:posted`
+  memo marker short-circuits the `prompt()` + `add-note` path once a
+  plan has been posted for an issue. Long-lived workflow labels (e.g.
+  `DEV::not started` that persists for days until a human starts work)
+  previously drove re-assembly and re-posting every 60s for the same
+  still-labeled issue. Marker is written in the `autonomous` and
+  `review-gated` branches only; `propose`/`shadow` don't post.
+  ([#223](https://github.com/Replikanti/agentis-colonies/issues/223))
 
 ### Security
 
--- a/dev-apprenticeship/CHANGELOG.md
+++ b/dev-apprenticeship/CHANGELOG.md
@@ -36,11 +36,15 @@ is asserted until multi-version CI is in place.
 
 - **`plan_reviewer` idempotency guard** — new `plan_reviewer:<iid>:posted`
   memo marker short-circuits the `prompt()` + `add-note` path once a
-  plan has been posted for an issue. Long-lived workflow labels (e.g.
-  `DEV::not started` that persists for days until a human starts work)
-  previously drove re-assembly and re-posting every 60s for the same
-  still-labeled issue. Marker is written in the `autonomous` and
-  `review-gated` branches only; `propose`/`shadow` don't post.
+  plan has been successfully posted for an issue at `autonomous` or
+  `review-gated` tier. Long-lived workflow labels (e.g. `DEV::not started`
+  that persists for days until a human starts work) previously drove
+  per-tick re-posting on the same still-labeled issue. The marker is
+  written **only** when the GitLab call returns a non-empty body, so
+  failed posts (auth/rate-limit/5xx/transport) are retried on the next
+  tick instead of silently consumed. `propose`/`shadow` tiers do no
+  external write and remain unmarked by design (so a future tier
+  promotion isn't blocked by a stale marker).
   ([#223](https://github.com/Replikanti/agentis-colonies/issues/223))
 
 ### Security
--- a/dev-apprenticeship/planning/README.md
+++ b/dev-apprenticeship/planning/README.md
@@ -57,7 +57,9 @@ When a new issue needs planning, the Scope Estimator, Risk Assessor, and Task De
 
 2. Configure your GitLab connection in `colony.toml`.
 
-3. Start the colony:
+3. (Optional) Override `[planning] trigger_label` if your project uses a label other than `needs-planning` to signal an issue is ready for planning. Scoped labels (`DEV::not started`) and labels with spaces are supported — `--data-urlencode` handles the encoding (#223).
+
+4. Start the colony:
    ```bash
    ./scripts/start-colony.sh
    ```
--- a/dev-apprenticeship/planning/agents/plan_reviewer.ag
+++ b/dev-apprenticeship/planning/agents/plan_reviewer.ag
@@ -104,6 +104,23 @@ fn tick(reason: string) -> void {
         return;
     };
 
+    // #223: idempotency guard. Long-lived workflow labels (e.g.
+    // "DEV::not started" on projects using scoped-label taxonomies)
+    // persist across human-gated transitions, so without this marker
+    // the reviewer would re-prompt() and re-add-note() every tick for
+    // days on end. Marker is written only in the branches that actually
+    // post a GitLab note (autonomous + review-gated); propose/shadow do
+    // no external write, so leaving them unmarked is correct.
+    let iid_str = to_string(target_iid);
+    let posted_key = "plan_reviewer:" + iid_str + ":posted";
+    if len(recall_latest(posted_key)) > 0 {
+        let now_e = try { exec sh "date -u +%Y-%m-%dT%H:%M:%SZ"; } catch e { ""; };
+        if len(now_e) > 0 {
+            memo_write("plan_reviewer:last_check", now_e);
+        };
+        return;
+    };
+
     let s = recall_latest("plan_reviewer:" + to_string(target_iid) + ":scope");
     let r = recall_latest("plan_reviewer:" + to_string(target_iid) + ":risks");
     let b = recall_latest("plan_reviewer:" + to_string(target_iid) + ":breakdown");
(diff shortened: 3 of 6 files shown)


ylohnitram:
## QA Report — PR #228 (agentis-qa-agent)

Branch: `fix/223-planning-trigger-label` @ 0d68672 (pre-fix) and 658384b (post-fix).

### Summary

- `tools/colony-lint.sh`: **80 passed / 0 failed / 5 skipped** (shellcheck/agentis not on this host).
- `tools/test-parse-toml.sh`: **25/25 passed** against the post-#223 example TOML.
- `tools/check-prompt-gate.sh`: **passed** — the new `recall_latest(posted_key)` is accepted as the gate for the sole `prompt()` in `plan_reviewer.ag`.
- `bash -n` on both modified scripts: **clean**.
- End-to-end wire trace: parse → export → `--data-urlencode` correctly produces `labels=DEV%3A%3Anot+started` for scoped + spaced labels; pre-#223 config (no `[planning]` section) produces byte-identical `labels=needs-planning`.
- CHANGELOG: `### Added` + `### Fixed` bullets reference #223. `VERSION` unchanged (correct — release PR territory).
- Cross-file `PLANNING_TRIGGER_LABEL` name symmetry verified (start-colony.sh, gitlab-api.sh, header docs). All 4 `.ag` agents still call `--needs-planning` verbatim — zero `.ag` diff for the trigger half.

### Findings

| # | Severity | Finding | Status |
| 1 | HIGH | `plan_reviewer` wrote `posted_key` marker even when `gitlab-api.sh add-note` failed — silent post-loss + false-success `learn()` row | Fixed in 658384b |
| 2 | LOW | `planning/README.md` didn't mention the new `[planning] trigger_label` key | Fixed in 658384b |
| 3 | LOW | CHANGELOG "re-assembly every 60s" claim slightly broader than the fix delivers (only re-posting is prevented; re-assembly still runs at propose/shadow tiers) | Fixed in 658384b |

### Finding 1 (HIGH) — silent failure: marker written on failed post

**Pre-fix shape** at `plan_reviewer.ag` L158–166 (autonomous) and L176–187 (review-gated):

```ag
try { exec sh post_cmd; } catch e { print(...); };
memo_write(posted_key, "1");  // ran regardless of post outcome
learn(..., "success", ["acted"]);
```

A non-zero exit from `exec sh` throws `ExecFailed (code 2010)`. `try/catch` catches it and control continues past the block. Verified against agentis v1.4.3 via a minimal daemon repro:

```
CAUGHT: Error { code: 2010, message: ExecFailed: exec sh failed (exit 1) ... }
after try block
```

Consequence: `gl_call` in `gitlab-api.sh` distinguishes failure modes (2 auth / 3 rate-limit / 4 client / 5 transport-or-5xx) and exits with those codes. Under `set -e`, a 401/429/5xx/transport failure silently consumed the issue — next tick short-circuits on `recall_latest(posted_key)`. The accompanying `learn("success", ["acted"])` also polluted the acting-row fitness that auto-promote reads.

**Post-fix** (commit 658384b): capture `exec sh` output into a variable; on catch, return empty string. Gate `memo_write(posted_key, ...)` and the `"success"` learn() on `len(post_result) > 0`. On failure, emit a `fail` learn() row so the failure is observable in experience telemetry. Same shape as `release/agents/version_bumper.ag:135–158` (tag/release creation).

### Finding 2 (LOW) — planning README silent on `trigger_label`

`dev-apprenticeship/planning/README.md` §Setup previously only referenced the default label by name. Added a one-liner pointing operators at the new knob when their project uses non-default labels.

### Finding 3 (LOW) — CHANGELOG wording

The original bullet said "previously drove re-assembly and re-posting every 60s". `propose`/`shadow` tiers still run the assembly `prompt()` — only the `add-note` call is skipped at those tiers. Narrowed the bullet to match what the guard actually prevents.

### Verification artifacts (2026-04-20 @ 658384b)

- `./tools/colony-lint.sh` → `Results: 80 passed, 0 failed, 5 skipped`.
- `./tools/test-parse-toml.sh` → `25 passed, 0 failed`.
- `./tools/check-prompt-gate.sh` → exit 0, no findings.
- `bash -n dev-apprenticeship/planning/scripts/{gitlab-api,start-colony}.sh` → both OK.
- Pre-#223 config (no `[planning]` section) → `PLANNING_TRIGGER_LABEL=""` → `labels=needs-planning` on wire. Byte-identical.
- `trigger_label = "DEV::not started"` → `labels=DEV%3A%3Anot+started` on wire.
- `memo_write(posted_key, ...)` grep confirms present in exactly the two posting branches and absent from propose / shadow / dormant.
- New failure-mode learn() present only in the two posting branches.

### Recommendation

All three findings addressed in 658384b. Ready for human review.

---

*Report produced by `agentis-qa-agent`. Full local report at `/tmp/colonies-plans/qa-223.md`.*

PR #240, opened 2026-04-20
fix(dashboard): prefer federation-local .agentis/ over parent-level (#238)

### LINKED ISSUE #238: federation-dashboard: /../.agentis lookup cross-reads between federations that share a parent
## Problem

`tools/federation-dashboard.sh` resolves the experience + log directories by walking **up** one level from the federation root:

```bash
# federation-dashboard.sh:109-115
local EXPERIENCE_DIR="${FED_DIR}/../.agentis/experience"
if [ ! -d "$EXPERIENCE_DIR" ]; then
    EXPERIENCE_DIR=".agentis/experience"
fi
local LOG_DIR="${FED_DIR}/../.agentis/logs"
if [ ! -d "$LOG_DIR" ]; then
    LOG_DIR=".agentis/logs"
fi
```

The fallback to local `.agentis/` only fires when the parent-level path doesn't exist. When two federations live as siblings under the same parent (the stock layout of this repo, with `dev-apprenticeship/` and `dev-apprenticeship-app/` both under `agentis-colonies/`), the parent-level lookup resolves to the **same directory** for both:

```
dev-apprenticeship/../.agentis          → agentis-colonies/.agentis   (Y1's real dir)
dev-apprenticeship-app/../.agentis      → agentis-colonies/.agentis   (same!)
```

The second federation's own `.agentis/` (a real directory at `dev-apprenticeship-app/.agentis/` on my setup) is never consulted, because the parent-level path exists and takes precedence.

## Symptoms on a running two-federation setup

I was running `dev-apprenticeship` against `X/Y1` on port 8420 and `dev-apprenticeship-app` against `X/Y2` on port 8421. The app federation had its own 16 experience files (500 rows of post-v1.4.1 `delta` data, all for X/Y2 MRs) in `dev-apprenticeship-app/.agentis/experience/`. Symptoms:

1. **App dashboard (port 8421) showed `MR 162` in the event timeline.** That's a Y1 MR number; X/Y2 is in the 5000+ range. 16 such entries, all attributed to `agent: 'ca4a1926'` (= `ca4a19269327`, the Y1 logic_reviewer). Expected app's own `logic_reviewer.ag` agent_id, got Y1's.
2. **App dashboard showed `experience_count: 0` for every agent.** The collector loads experience rows from the resolved `EXPERIENCE_DIR`, then matches them by agent_id. The agent_ids passed from the app federation don't match any Y1 experience file, so the match set is empty and every agent renders as "no data."
3. **Claude flat-rate usage was burning** while the app dashboard said nothing was happening. Real app agents were ticking, calling `prompt()`, writing to `dev-apprenticeship-app/.agentis/experience/`. The dashboard was reading from the wrong place, not the runtime.

## Reproduction

```
# From a fresh clone of agentis-colonies/ with both federations installed.
cd agentis-colonies

# Start Y1 federation + its dashboard.
(cd dev-apprenticeship && ./start-federation.sh &)
./tools/federation-dashboard.sh dev-apprenticeship 8420 &

# Let it run long enough to produce experience entries, then start app federation.
(cd dev-apprenticeship-app && ./start-federation.sh &)
./tools/federation-dashboard.sh dev-apprenticeship-app 8421 &

# Visit http://localhost:8421 — it will render Y1 data
# (MR numbers, agent timelines) attributed to app agent slots.

ls -l dev-apprenticeship/.agentis
# lrwxr-xr-x → ../.agentis            (symlink to the shared top-level .agentis)

ls -ld dev-apprenticeship-app/.agentis
# drwxr-xr-x                          (real, standalone directory, 500 rows of data inside)

# Dashboard resolves EXPERIENCE_DIR=../.agentis/experience for BOTH federations.
```

## Why this layout is plausible

`dev-apprenticeship/` is symlinked (`.agentis → ../.agentis`) so it naturally shares the parent-level `.agentis` with anything else in the repo — that's the intended path for the single-federation install. `dev-apprenticeship-app/` was added later (in-tree) with its own standalone `.agentis/` to keep its runtime state isolated. The dashboard's parent-walk heuristic was never extended to know about the second case.

## Proposed fix

Flip the precedence: prefer the federation-local `.agentis/` if it exists, fall back to the parent-level one only when local is absent.

```bash
# Prefer federation-local .agentis/ (isolated federation layout).
local EXPERIENCE_DIR="${FED_DIR}/.agentis/experience"
if [ ! -d "$EXPERIENCE_DIR" ]; then
    # Fall back to shared parent-level .agentis/ (symlinked single-federation layout).
    EXPERIENCE_DIR="${FED_DIR}/../.agentis/experience"
fi
# Same for LOG_DIR.
```

This matches what the symlinked case already does (`dev-apprenticeship/.agentis` → `../.agentis` resolves to the parent via the local-first check), and gives standalone-federation layouts their own state.

## Scope

- `tools/federation-dashboard.sh` — two blocks to flip (`EXPERIENCE_DIR`, `LOG_DIR`).
- Verify `federation-dashboard-collector.py` callers pass the right paths through.
- Smoke test: run both federations in parallel, confirm port 8420 shows Y1 MRs only and port 8421 shows app MRs only.

## Related

- Surfaced while running portability-audit federation on `X/Y2` for the first time — same X corporate GitLab, sibling federations under one `agentis-colonies/` checkout.
- Discovered alongside a separate concern about the `memo.max_size = 50` default being tight for 21-agent federations (multiple `:confidence` + `:last_check` + per-agent extras sit right on the ceiling, any stray key trips router into critical). Filing that separately if worth it.

ylohnitram:
## Summary

- Flip path-resolution precedence in `tools/federation-dashboard.sh` so federation-local `.agentis/experience` and `.agentis/logs` win over the parent-level walk (`${FED_DIR}/../.agentis/...`).
- Sibling federations under a shared parent (the `dev-apprenticeship` + `dev-apprenticeship-app` layout in this repo) were cross-reading each other's experience/logs because both resolve `${FED_DIR}/../.agentis` to the same directory.
- Symlinked single-federation layout (`<fed>/.agentis -> ../.agentis`) continues to work — the local-first check resolves through the symlink just like the old parent-walk did.

## Test plan

- [x] `tools/test-dashboard-fedpath.sh` (new): standalone layout reads federation-local and does NOT cross-read parent; symlinked layout resolves through the symlink.
- [x] `tools/test-timeline-rendering.sh` (existing): still passes (logs at `$TMP/.agentis/logs` one level above FED_DIR — now exercises the parent-level fallback branch).
- [x] `./tools/colony-lint.sh` — 81 passed, 0 failed, 5 skipped. Includes auto-picked new test.

Fixes #238

### DIFF
--- a/dev-apprenticeship/watch-suggestions.sh
+++ b/dev-apprenticeship/watch-suggestions.sh
@@ -12,11 +12,24 @@ set -e
 
 SCRIPT_PATH="$(python3 -c 'import os, sys; print(os.path.realpath(sys.argv[1]))' "$0")"
 SCRIPT_DIR="$(dirname "$SCRIPT_PATH")"
-LOG_DIR="${1:-${SCRIPT_DIR}/../.agentis/logs}"
 
-if [ ! -d "$LOG_DIR" ]; then
-    # Try current working directory
-    LOG_DIR=".agentis/logs"
+# Resolve log directory with fed-local-first precedence (mirrors
+# tools/federation-dashboard.sh after #238). The previous default
+# "${SCRIPT_DIR}/../.agentis/logs" resolved to the repo-root shared
+# .agentis/ when two federations live as siblings under one checkout,
+# cross-reading the other federation's logs. Daemons write to
+# <federation>/.agentis/logs, so prefer that first and only fall
+# through to the parent-level or cwd path when the local one is absent.
+if [ -n "${1:-}" ]; then
+    LOG_DIR="$1"
+else
+    LOG_DIR="${SCRIPT_DIR}/.agentis/logs"
+    if [ ! -d "$LOG_DIR" ]; then
+        LOG_DIR="${SCRIPT_DIR}/../.agentis/logs"
+    fi
+    if [ ! -d "$LOG_DIR" ]; then
+        LOG_DIR=".agentis/logs"
+    fi
 fi
 
 if [ ! -d "$LOG_DIR" ]; then
--- a/tools/federation-dashboard.sh
+++ b/tools/federation-dashboard.sh
@@ -106,11 +106,26 @@ generate() {
     done
     COLONY_LIST_PY="[${COLONY_LIST_PY%,}]"
 
-    local EXPERIENCE_DIR="${FED_DIR}/../.agentis/experience"
+    # #238: prefer federation-local .agentis/ if present (standalone-federation
+    # layout with a real .agentis/ directory inside the federation root). Fall
+    # back to the parent-level .agentis/ only when local is absent — that is
+    # the symlinked single-federation layout where `<fed>/.agentis -> ../.agentis`
+    # already resolves through the local-first check. Without this ordering,
+    # sibling federations under a shared parent (e.g. dev-apprenticeship +
+    # dev-apprenticeship-app both under agentis-colonies/) cross-read each
+    # other's experience/logs because `<fed>/../.agentis` resolves to the same
+    # shared directory for both.
+    local EXPERIENCE_DIR="${FED_DIR}/.agentis/experience"
+    if [ ! -d "$EXPERIENCE_DIR" ]; then
+        EXPERIENCE_DIR="${FED_DIR}/../.agentis/experience"
+    fi
     if [ ! -d "$EXPERIENCE_DIR" ]; then
         EXPERIENCE_DIR=".agentis/experience"
     fi
-    local LOG_DIR="${FED_DIR}/../.agentis/logs"
+    local LOG_DIR="${FED_DIR}/.agentis/logs"
+    if [ ! -d "$LOG_DIR" ]; then
+        LOG_DIR="${FED_DIR}/../.agentis/logs"
+    fi
     if [ ! -d "$LOG_DIR" ]; then
         LOG_DIR=".agentis/logs"
     fi
--- a/tools/test-dashboard-fedpath.sh
+++ b/tools/test-dashboard-fedpath.sh
@@ -53,8 +53,12 @@ free_port() {
 }
 
 boot_dashboard() {
-    local fed_dir="$1" port="$2" log_file="$3"
-    setsid bash "$DASHBOARD_SH" "$fed_dir" "$port" >"$log_file" 2>&1 &
+    local fed_dir="$1" port="$2" log_file="$3" cwd="${4:-}"
+    if [ -n "$cwd" ]; then
+        ( cd "$cwd" && setsid bash "$DASHBOARD_SH" "$fed_dir" "$port" >"$log_file" 2>&1 ) &
+    else
+        setsid bash "$DASHBOARD_SH" "$fed_dir" "$port" >"$log_file" 2>&1 &
+    fi
     local pid=$!
     DASH_PIDS="$DASH_PIDS $pid"
     for _ in 1 2 3 4 5 6 7 8 9 10; do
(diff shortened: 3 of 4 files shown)


ylohnitram:
## QA Report — PR #240 (#238 dashboard-fedpath)

### Verdict: ship

The precedence flip does what the PR claims. All three documented deploy layouts resolve correctly, no pre-existing consumer regresses, new regression test locks in the fix, and colony-lint passes 81/0. Findings below are non-blocking follow-ups or scope observations.

### Summary
- `tools/test-dashboard-fedpath.sh`: 3/3 pass locally
- `tools/test-timeline-rendering.sh`: 19/19 pass (now exercises the parent-level fallback branch as the PR description acknowledges)
- `./tools/colony-lint.sh`: 81 passed, 0 failed, 5 skipped (shellcheck skips — CI runs it)
- GitHub Actions `colony-lint` check: green

### Findings

| 1 | INFO | correctness | Precedence `fed-local -> parent -> cwd` is correct for all three layouts (standalone, symlinked-to-parent, legacy cwd) | - [x] verified |
| 2 | INFO | correctness | Legacy "federation w/o local `.agentis/`, intentionally reading parent" case still works — fallback to `${FED_DIR}/../.agentis` is preserved | - [x] verified |
| 3 | INFO | correctness | `_` loop-variable rename in `boot_dashboard()` has no semantic effect — loop body does not reference `i`/`_` | - [x] verified |
| 4 | INFO | correctness | Regression test verifies the collector-read source (asserts `local:only_in_federation_local` present AND `shared:event` absent), not just HTTP 200 | - [x] verified |
| 5 | LOW | scope | `dev-apprenticeship/watch-suggestions.sh` has the same bug class as #238 — its default `LOG_DIR="${SCRIPT_DIR}/../.agentis/logs"` points at the repo-root `.agentis/`, which is exactly the wrong directory under a sibling-federation layout. Pre-existing on `main` — not introduced by this PR, but in scope for #238 if the intent was to eliminate all cross-read paths for dev-apprenticeship | - [x] addressed in e2fd20e |
| 6 | LOW | test coverage | `test-dashboard-fedpath.sh` does not exercise the final cwd fallback (layout c — no `.agentis` under fed or parent). The fallback is unchanged from main so no regression risk, but the test matrix is incomplete | - [x] addressed in e2fd20e |
| 7 | LOW | docs | Stale comment at `tools/test-timeline-rendering.sh:47`: `# The dashboard reads logs from "$FED_DIR/../.agentis/logs"` — now only one of two possible paths. Fixture still works (only `$FED_DIR/.agentis/daemon` is created, so `-d "$FED_DIR/.agentis/logs"` is false and the fallback fires), but the comment is misleading post-#238 | - [x] addressed in e2fd20e |

### Details

#### Finding 1 — Precedence verified across three layouts
- **Case A (standalone):** fixture creates `$FED/.agentis/logs/local_agent.log` and `$TMP/.agentis/logs/shared_agent.log`. Served HTML contains `local:only_in_federation_local` and does NOT contain `shared:event`. Confirms fed-local wins over parent. PASS.
- **Case B (symlinked):** `$FED_B/.agentis -> ../.agentis`. Served HTML contains `shared:event` — the `-d` check follows the symlink so the local-first branch transparently reads the parent. PASS.
- **Case C (legacy-parent, not in new test but covered by `test-timeline-rendering.sh`):** fixture creates `$FED_DIR/.agentis/daemon` (no logs subdir) and `$TMP/.agentis/logs/stub_agent.log`. Dashboard falls back to parent-level and ingests the log. PASS.

#### Finding 5 — Sibling cross-read in `watch-suggestions.sh`
```
dev-apprenticeship/watch-suggestions.sh:15
LOG_DIR="${1:-${SCRIPT_DIR}/../.agentis/logs}"
```
`$SCRIPT_DIR` is `dev-apprenticeship/`, so the default resolves to `agentis-colonies/.agentis/logs` — i.e. the shared parent, which is precisely what #238 identifies as cross-reading sibling federation data. Federation `.ag` daemons actually write to `dev-apprenticeship/.agentis/logs` (see `*/scripts/start-colony.sh` which uses `AGENTIS_LOGS="$REPO_ROOT/dev-apprenticeship/.agentis/logs"`). So the default `watch-suggestions.sh` path is already wrong on `main` and reports "no logs" unless the user passes an explicit argument. Pre-existing bug, fix belongs in a follow-up PR (or an in-scope amendment if you want #238 to close the federation-data-isolation story completely).

#### Finding 7 — Stale comment in `test-timeline-rendering.sh`
Line 47 now says "The dashboard reads logs from `$FED_DIR/../.agentis/logs`" but post-#238 it reads from `$FED_DIR/.agentis/logs` first. Fixture still exercises the right branch (parent-level) because `$FED_DIR/.agentis/logs` isn't created. Trivial comment touch-up.

### Notes on earlier-run test failures

During the initial colony-lint run I saw `test-kill-endpoint.sh` fail 4/5. Verified the failures were environmental — leftover dashboard processes from an earlier manual run collided on port allocation / `.dashboard/` state. A clean rerun (after verifying no `federation-dashboard` procs) passes 5/5 on both this branch and `main`. Not a PR-caused regression.

### Files reviewed
- `tools/federation-dashboard.sh` (changed)
- `tools/test-dashboard-fedpath.sh` (new)
- `tools/test-timeline-rendering.sh` (adjacent — verified still passes)
- `tools/test-kill-endpoint.sh` (adjacent — verified not regressed)
- `tools/kill-federation.sh` (other `.agentis` consumer — already uses fed-local-first precedence, no action needed)
- `tools/auto-promote.sh` (other `.agentis` consumer — uses `cd "$FED_DIR"` so CLI resolves to fed-local, no action needed)
- `tools/federation-dashboard-collector.py` (defensive `os.path.isdir` guards on log_dir — graceful on missing path)
- `dev-apprenticeship/watch-suggestions.sh` (pre-existing same-class bug; see finding 5)

---

### Addendum — all three findings addressed inline (e2fd20e)

- **#5**: `watch-suggestions.sh` now uses fed-local-first → parent → cwd precedence matching `tools/federation-dashboard.sh`. Same three-arm resolver; closes the cross-read surface for the live-suggestions feed under sibling-federation layouts.
- **#6**: Added Case C to `test-dashboard-fedpath.sh` — FED without fed-local or parent-level `.agentis/`, booted with an explicit cwd where `.agentis/logs/` exists. Asserts the cwd log is read AND that the unrelated `$TMPDIR/.agentis/` from Cases A/B does NOT leak in (regression guard against a resolver that would silently walk further up the tree).
- **#7**: Replaced the stale one-path comment at `test-timeline-rendering.sh:47` with a description of the three-step precedence and which arm this fixture exercises (parent-level fallback).

Re-ran locally post-fix: `test-dashboard-fedpath.sh` 5/0 (A.1, A.2, B.1, C.1, C.2), `test-timeline-rendering.sh` 19/0, `colony-lint.sh` 81/0/5.

Also confirmed by grep scan that no other `.agentis/` consumer in the tree still uses the bare `../.agentis` idiom: `tools/kill-federation.sh` and `tools/federation-dashboard.sh` already use fed-local-first precedence; `tools/auto-promote.sh` works via `cd "$FED_DIR"` so the CLI resolves fed-local; the Python collectors guard on `os.path.isdir`. `watch-suggestions.sh` was the only remaining consumer with the cross-read default.

PR #241, opened 2026-04-20
dashboard: calibrate evolve flat-slope threshold from production data (#163)

### LINKED ISSUE #163: federation-dashboard: calibrate evolve flat-slope threshold from production data
## Context

Follow-up to #160. The `actionGate('evolve')` rule in `tools/federation-dashboard.sh` uses a hardcoded magnitude threshold to decide when fitness slope is "flat enough to disable evolve":

```js
if (Math.abs(f.slope) < 1e-6) {
  // disabled, "Fitness slope flat over the last N entries..."
}
```

Where `f.slope` is the linear-regression slope of `delta` values over the last 100 experience entries (server-side computation in the same file).

## Problem

`1e-6` is a guess, not measured. Typical `delta` magnitudes are ~`0.01`, so a slope of `1e-4` over 100 entries is also effectively noise — but it sits well above `1e-6` and would trip the "evolve enabled" branch. In practice the `flat-slope` disabled state may almost never fire, defeating the gate's purpose: operators get the evolve button on agents whose fitness is fluctuating around zero.

PR reviewer flagged this in the review of #160 as cosmetic but real: choosing the right magnitude needs empirical data from running daemons, which we don't have yet.

## Proposed plan

1. Once dev-apprenticeship federation has run long enough to accumulate ≥100 experience entries per agent in at least one colony, run `tools/auto-promote.sh --dry-run` (or directly inspect `experience/<id>.jsonl`) and record the per-agent slope distribution.
2. Pick a threshold empirically, e.g. percentile of observed `|slope|` for agents where evolve would be a no-op (Opus plateau). Likely candidates: `1e-4` or `1e-3`.
3. Update the threshold in the same gate. Optionally surface it as a constant near `auto-promote-config.yaml`'s `delta_slope_min` so dashboard + cron stay aligned.

## Where to drop the data

When production data is available, attach the per-agent slope numbers as a comment on this issue and the threshold can be tuned in a 1-line PR.

## Related

- #160 — capability + fitness-aware action gating (introduces the gate)
- `tools/federation-dashboard.sh` — `actionGate('evolve')` `flat-slope` branch
- `tools/auto-promote-config.yaml` — `delta_slope_window: 100`, `delta_slope_min: 0`

ylohnitram:
## Summary

- Bump `SLOPE_FLAT_THRESHOLD` in `tools/federation-dashboard.html.template` from the pre-calibration `1e-6` guess to the production-calibrated `1e-4`.
- Promote the magic number to a named const so future re-calibrations are a one-line tweak and the value is discoverable.
- The evolve gate's `flat-slope` reason string now surfaces the threshold so operators can read the calibration from the UI.

## Calibration source

Data snapshot posted on #163: 23 agents across two federations (`dev-apprenticeship` on `X/Y1`, `dev-apprenticeship-app` on `X/Y2`) on `agentis v1.4.3` with `fitness_delta` populated per outcome (post-core-PR-#542).

Per-agent `|slope|` distribution over the last 100 experience entries splits cleanly into:

| cluster | range | agents |
|---|---|---|
| true plateau | `|slope| <= 1e-5` | `code_writer`, `commit_composer`, `router`, `prioritizer` |
| oscillation | `|slope|` in `[1e-4, 3e-3]` | most everything else |
| gap | `(1e-5, 1e-4)` | **empty** |

`1e-4` sits inside the gap — it catches every cluster-1 plateau without false positives from cluster 2. `1e-3` would also lump in oscillating agents like `logic_reviewer` and `test_reviewer`, so it was rejected as too aggressive.

Full tables (including Y1 / app breakdowns and threshold-sensitivity analysis) are in the #163 comment.

## Test plan

- [x] `tools/test-timeline-rendering.sh` — new test 20 guards against reverting to `1e-6`, orphaning the constant, or the flat-slope branch bypassing it. 20/20 pass.
- [x] `./tools/colony-lint.sh` — 80 passed, 0 failed, 5 skipped.

## Out of scope

- `auto-promote-config.yaml`'s `delta_slope_min` / `delta_slope_negative_for` are semantically orthogonal (promote prereq / evolve trigger, not the evolve flat-slope gate). No change proposed there — the recommendation on the issue was explicit about only the dashboard threshold.
- Re-snapshot in 2–4 weeks once the pre-v1.4.1 zero-prefix ages out of the `last100` window for the seven agents still carrying it (see caveats on the issue). If the plateau cluster shifts, the one-line `SLOPE_FLAT_THRESHOLD = …` can be retuned without touching the gate logic.

Fixes #163

### DIFF
--- a/tools/federation-dashboard.html.template
+++ b/tools/federation-dashboard.html.template
@@ -647,6 +647,17 @@ const TIMELINE_AUTO_HIDE_STALE_KEY = 'dashboard.timeline.autoHideStale.' + FED_N
 const TIMELINE_STALE_WINDOW_MS = 3600 * 1000;
 // #167: every class the timeline emits, in render order.
 const TIMELINE_CLASSES = ['emit', 'recv', 'suggest', 'error', 'action', 'finding'];
+// #163: magnitude below which the fitness slope over the last 100 experience
+// entries is treated as flat and evolve is disabled as an Opus plateau. Raised
+// from the original 1e-6 guess to 1e-4 after production calibration against
+// dev-apprenticeship on v1.4.3 with populated fitness_delta values: the
+// observed per-agent |slope| distribution splits cleanly into a true-plateau
+// cluster at |slope| <= 1e-5 (code_writer, commit_composer, router,
+// prioritizer — the canonical "evolve is pointless" cases the gate is meant
+// to catch) and an oscillation cluster at |slope| in [1e-4, 3e-3], with no
+// data in the gap between. 1e-4 is the natural separator — full snapshot
+// tables are on the issue.
+const SLOPE_FLAT_THRESHOLD = 1e-4;
 
 function tlReadJSON(key, fallback) {
   try {
@@ -1423,10 +1434,10 @@ function actionGate(a, type, target) {
         evidence: f
       };
     }
-    if (Math.abs(f.slope) < 1e-6) {
+    if (Math.abs(f.slope) < SLOPE_FLAT_THRESHOLD) {
       return {
         enabled: false, ruleType: 'flat-slope',
-        reason: 'Fitness slope flat over the last ' + f.window + ' entries (likely Opus plateau) \u2014 evolve unlikely to find improvement.',
+        reason: 'Fitness slope flat over the last ' + f.window + ' entries (|slope| < ' + SLOPE_FLAT_THRESHOLD.toExponential(0) + ', likely Opus plateau) \u2014 evolve unlikely to find improvement.',
         evidence: f
       };
     }
--- a/tools/test-timeline-rendering.sh
+++ b/tools/test-timeline-rendering.sh
@@ -429,6 +429,33 @@ else
     fail "19: heredoc detected — extract content to a standalone file (see #172)"
 fi
 
+# --- #163: evolve flat-slope threshold is named, reachable, and calibrated.
+#     Guards against reverting to the pre-calibration 1e-6 guess or silently
+#     orphaning the constant (declared but never used). The production
+#     calibration in #163 picked 1e-4 as the natural separator between the
+#     true-plateau cluster (|slope| <= 1e-5) and the oscillation cluster
+#     (|slope| >= 1e-4). ---
+if python3 - "$TEMPLATE_HTML" <<'PY' 2>/dev/null
+import sys, re
+with open(sys.argv[1]) as f:
+    src = f.read()
+m = re.search(r'const\s+SLOPE_FLAT_THRESHOLD\s*=\s*([0-9eE.+\-]+)\s*;', src)
+if not m: sys.exit(1)
+value = float(m.group(1))
+# Must equal the calibrated value (not the pre-calibration 1e-6).
+if abs(value - 1e-4) > 1e-12: sys.exit(2)
+# Must actually be referenced (no orphaned constant).
+if 'SLOPE_FLAT_THRESHOLD' not in src[m.end():]: sys.exit(3)
+# Legacy literal must no longer guard the flat-slope branch.
+if re.search(r'Math\.abs\(\s*f\.slope\s*\)\s*<\s*1e-6\b', src): sys.exit(4)
+sys.exit(0)
+PY
+then
+    pass "20: SLOPE_FLAT_THRESHOLD defined as 1e-4, referenced, no leftover 1e-6 guard (#163)"
+else
+    fail "20: SLOPE_FLAT_THRESHOLD missing, wrong value, orphaned, or 1e-6 guard survived (#163)"
+fi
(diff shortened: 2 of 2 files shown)


ylohnitram:
## QA Report — PR #241 (issue #163, evolve flat-slope threshold calibration)

### Summary

- CI status: `colony-lint` PASS on clean runner (GitHub Actions).
- Local `./tools/test-timeline-rendering.sh`: **20/20 pass** (including new test 20 for #163).
- Diff scope: exactly two files, 40 additions / 2 deletions. No data-path or server-side changes.

### Findings

| # | Severity | Category | Finding | Blocker | Status |
|---|----------|----------|---------|---------|--------|
| 1 | INFO | calibration | `1e-4` sits inside a clean 2-order gap in the observed `|slope|` distribution on v1.4.3 data; 23 agents surveyed across two federations, reasoning aligns with issue comment #6's snapshot tables. | no | [x] verified |
| 2 | INFO | placement | Constant at template global scope (line 660), outside any per-agent block. Renderer is a 10-sentinel `str.replace` that runs once; constant appears once in output HTML regardless of agent count. | no | [x] verified |
| 3 | INFO | ui | `SLOPE_FLAT_THRESHOLD.toExponential(0)` produces exactly `"1e-4"` (confirmed in Node). Reason string reads naturally: `(|slope| < 1e-4, likely Opus plateau)`. | no | [x] verified |
| 4 | INFO | wire-up | Collector emits `experience_slope_recent` → template `computeFitness` reads `a.experience_slope_recent` → `Math.abs(f.slope) < SLOPE_FLAT_THRESHOLD`. No unit mismatch — both sides use the same server-side-computed linreg slope over the last 100 entries. | no | [x] verified |
| 5 | INFO | cross-config | `auto-promote-config.yaml`'s `delta_slope_min` (promote prereq: average delta non-negative) and `delta_slope_negative_for` (evolve trigger: sustained negative slope over 1000 rows) are **semantically orthogonal** to the dashboard's flat-slope gate — they answer different questions. No coupling needed; PR correctly leaves them alone. | no | [x] verified |
| 6 | INFO | impact | On the calibration federation, the change correctly flips `router` (`8.6e-5`) and `prioritizer` (`2.1e-5`) from "evolve enabled" to "evolve disabled" (matching the snapshot's plateau cluster) without creating new false positives in the oscillation cluster. No cluster-2 agents get newly-disabled. | no | [x] verified |
| 7 | LOW | test | Test 20's third check (`'SLOPE_FLAT_THRESHOLD' not in src[m.end():]`) enforces "referenced somewhere after the definition," not "referenced *as the guard* in the flat-slope branch." A future regression that wires the constant only into the reason string while reverting the comparison to a different literal (e.g. `1e-5`) could pass. The legacy-guard check catches only `1e-6` verbatim. The PR itself is not affected — this is a minor future-proofing gap. | no | [x] noted, not blocking |
| 8 | LOW | re-snapshot | PR's own caveat: some agents' `last100` windows still contain pre-v1.4.1 zero rows, biasing slopes toward zero. Once the zero prefix ages out, the cluster-1 ceiling may climb and `1e-4` could become too tight. PR description already commits to a re-snapshot in 2–4 weeks; the one-line tweak pathway is established. | no | [x] noted, operator follow-up |

### Verdict: **ship it**

The threshold bump from `1e-6` to `1e-4` is **correctly calibrated** against real v1.4.3 production data (23 agents, two federations). The empirical `|slope|` distribution shows a clean 2-order-of-magnitude gap in `(1e-5, 1e-4)` with no data inside it, so `1e-4` is a genuinely defensible separator — not a guess. The wire-up is correct (named constant at global scope, referenced in the actual comparison, reason string surfaces the value). The new test 20 guards against the two most likely regressions (reverting the literal, orphaning the constant). Post-merge operator impact is minimal — two cluster-1 agents correctly flip to evolve-disabled, no false positives in cluster 2.

**Non-blocking follow-ups** (operator / future PR):
- Finding #7: strengthen test 20 to assert the constant guards the flat-slope *comparison*, not just that it appears somewhere after the definition.
- Finding #8: re-snapshot per the PR's own plan once the pre-v1.4.1 zero prefix ages out of the last-100 windows; retune if the cluster-1 ceiling climbs above `1e-4`.

### Files reviewed

- `tools/federation-dashboard.html.template` (constant declaration at line 660, gate comparison + reason string at lines 1437–1440)
- `tools/test-timeline-rendering.sh` (new test 20 at lines 432–457)
- `tools/federation-dashboard-collector.py` (server-side slope computation at lines 205–223 feeding the threshold)
- `tools/federation-dashboard-renderer.py` (confirms template → index.html is a single-pass sentinel substitution — constant is not per-agent)
- `tools/auto-promote-config.yaml` (confirmed semantically orthogonal — no change needed)

Issue #200, opened 2026-04-18
implementation colony: sticky assigned-issue re-burns LLM prompts every tick


ylohnitram:
## Problem

The three `implementation/` agents (`code_writer`, `test_writer`, `refactorer`) re-analyze the same artefact on every tick as long as the work is not advanced by an autonomous-tier action. Quantified: one stuck assigned issue at propose / review-gated / shadow tier burns ≈ 180 `prompt()` calls / hour across the colony (60 s tick × 3 agents). Idle ticks are already 0 after #147 — this is the single largest residual LLM waste in the federation.

### `code_writer.ag`

- `implementation/agents/code_writer.ag:79` polls `assigned-issues` **without a `--since` filter** — every tick gets the full list.
- Line 87 gate is `len(issues_raw) > 10`, which catches `[]` but is satisfied as long as any assignment exists.
- Line 91 `draft = prompt(...)` fires every tick regardless of whether the *same* issue was drafted the previous tick. There is no `last_drafted_iid` memo.

### `test_writer.ag` and `refactorer.ag`

- `implementation/agents/test_writer.ag:81` and `implementation/agents/refactorer.ag:82` both `listen("implementation:code_draft", 100)` — the bus primitive drains, so they would be idle in isolation.
- But `code_writer.ag` re-emits `implementation:code_draft` on every tick where the draft prompt fires (lines 127 / 141 / 145 depending on tier). So `test_writer` and `refactorer` get a fresh emit per tick and re-prompt accordingly.
- Result: upstream waste propagates downstream — one stuck assigned issue triggers three LLM prompts per tick.

## Impact

- 60 s tick × 60 min = 60 ticks / hour. Three agents × 60 ticks = **180 `prompt()` calls / hour per stuck issue**.
- Since the operator-calibration phase deliberately pins agents at propose tier until auto-promote's gates allow promotion (#186 / #192), and the #195 feedback loop is still under construction, most production ticks are currently in the "stuck" regime.

## Proposed fix

Two symmetric layers:

1. **Upstream gate in `code_writer.ag`.** Before the `draft = prompt(...)` at line 91:
   - Extract `first_iid = parse_int(to_string(json_get(issues_raw, \"[0].iid\")))` and `upd = to_string(json_get(issues_raw, \"[0].updated_at\"))`.
[...]

Issue #201, opened 2026-04-18
reviewers + planning observers: pattern-learning prompt has no staleness gate


ylohnitram:
## Problem

Four code-review agents (`style_reviewer`, `logic_reviewer`, `security_reviewer`, `test_reviewer`) and three planning observers (`scope_estimator`, `risk_assessor`, `task_decomposer`) call a "learn from recent human patterns" `prompt()` on every work tick. The delta-gate in front of them (`len(raw) < 3` from the `--since` filter) catches *no-activity* ticks but not *same-activity* ticks — when a stuck MR or unplanned issue stays open for hours, the same window of comments / issues is re-summarised tick after tick, even though the summary is identical.

Smaller waste than #200, but non-zero and easy to close.

## Evidence

| Agent | Line | Prompt |
|---|---|---|
| `style_reviewer.ag` | 103 | \"Analyze these MR review comments. Find comments about style issues...\" |
| `logic_reviewer.ag` | 77 | analogous human_patterns extraction |
| `security_reviewer.ag` | 78 | analogous |
| `test_reviewer.ag` | 69 | analogous |
| `scope_estimator.ag` | 72 | \"Summarize recently merged MRs...\" (gated by `len(merged_raw) > 10`) |
| `risk_assessor.ag` | 74 | analogous |
| `task_decomposer.ag` | 72 | analogous |

In every case the prompt's *input* is "recent history since `last_check`". After the learn completes, `last_check` advances — so the *next* tick's input is a tighter window. But on a busy project the window often overlaps substantially, and on a slow project the same small window is re-analysed until new activity arrives.

## Impact

- On a quiet project (1 MR / day, 2 issues / week): ≈ 5-10 extra `prompt()` calls / day.
- On an active project during a sprint push: tens per day — still small against draft / review prompts, but strictly avoidable.
- The prompt's output is *pattern memory* — it is naturally stable over short windows, so re-running it without new data is pure waste.

## Proposed fix

Add a per-agent staleness memo:

1. Before firing the pattern prompt, `let last_learn = recall_latest(\"<agent>:last_pattern_learn_ts\");`
2. Compare against `now` (from the existing `date +%s` shell call). If the gap is less than a threshold (default 3600 s = 1 h; planning observers already use a 30 min gate conceptually — align with that), skip the prompt.
3. After the prompt completes successfully, write `memo_write(\"<agent>:last_pattern_learn_ts\", to_string(now_epoch));`.

[...]

PR #242, opened 2026-04-20
fix(implementation): gate MR-learning loop per iid (#239)

### LINKED ISSUE #239: implementation colony agents die silently after hours of MR-stuck-loop — no tick_err, no graceful stop, no lifecycle event
## Symptom

Running `dev-apprenticeship` federation against a real GitLab project (X/Y1) for ~6 hours. Three of four implementation-colony agents — `code_writer`, `commit_composer`, `refactorer` — have dropped to `state=stopped health=stopped` one-by-one without any of the usual exit signatures:

- `tick_err=0` (no ticks ever errored — counter frozen at the value it held while running)
- `restarts=0` (watchdog never replaced them)
- No `daemon.stopped` entry in `.agentis/lifecycle/events.jsonl` for the current session
- No `tick:error` block in the agent log
- The stored pid in `.agentis/daemon/<id>.pid` no longer corresponds to a live process
- `.agentis/daemon/<id>.watchdog.pid` is missing
- `.agentis/daemon/<id>.status` is missing

A previous session of the same federation shows a clearer tell: `commit_composer` had `{"event":"daemon.stop_timed_out","stage":"sigkill","grace_ms":10000,...}` — i.e. something took long enough to enter watchdog SIGKILL escalation. In the current session no such event was written, the process just vanished.

## What the agents were doing before death

Same pattern in all three logs: 300+ tick_ok entries, each a one-liner of the form `[<agent>] Learned from ... MR <N>`, all referencing the same MR. For example, `code_writer`'s tail is 316 consecutive `Learned from MR 180` entries spanning ~5 hours at one tick per ~75 s. No other log content — no errors, no variation, no terminal marker.

The fourth implementation agent, `test_writer`, is still running but shows the exact same tight loop (`Learned from 2 test files in MR 180` × many hundreds) and will almost certainly follow the others. Agents in other colonies (triage / code-review / planning / release) have no such loops and remain healthy.

## Hypothesis

Two specific things about the implementation colony are suspicious:

1. **Highest `cb_budget`** of any colony in the federation: 2000 / 1500 / 1000 / 600 vs 400–600 elsewhere. If the .ag runtime leaks memory per tick in proportion to budget (or if the learn/memo path retains per-tick state that grows with cb usage), implementation agents would accumulate faster and hit OS limits sooner.
2. **Tight, never-advancing observation loop**: `code_writer`'s logic at `implementation/agents/code_writer.ag:43-46` advertises a #200 fix to early-exit when the top assigned issue hasn't changed since last tick. The log shape suggests the early-exit doesn't fire on this path — every tick still reaches the `Learned from MR <same N>` emit, which I expected to only happen on a fresh assignment. The combination (tight loop × high cb) maps cleanly to the observed failure mode.

The other colonies don't have this shape — their agents have cb_budget in the low hundreds and their observed paths are either reactive (wait for a bus emit) or gated by `--since` against timestamped queries.

## Reproduction

Any dev-apprenticeship install with at least one GitLab issue that stays labeled with the implementation trigger for hours (a common real-world shape: a single `DEV::in progress` (or equivalent) issue that the human hasn't finished yet). Watch `.agentis/logs/<code_writer_id>.log` for 2–3 hours. Expected: the log accumulates identical `Learned from MR <N>` entries until the OS kills the process, then silence with no error trace anywhere.

## Impact

- The federation silently loses a colony's worth of learning throughput after a few hours of uptime against any project with a long-lived trigger-labeled issue. Most real GitLab projects have such issues.
- No watchdog signal means no auto-restart and no dashboard visibility beyond the `state=stopped` on the registry row — which reads identically to a graceful stop.
- Operators with the dashboard open might never notice their implementation colony has been dead for hours.

## What would help

- A terminal `daemon.stopped` event on any exit path — including SIGKILL-by-OS — so post-mortems have at least one timestamp + reason to correlate against.
- Instrumented memory per-agent (RSS, heap allocations) exposed via `agentis daemon list --json` so operators can spot drift before the OS intervenes.
- Either an idempotency guard in `code_writer.ag` (memo `code_writer:{iid}:learned`) so repeat-observations don't re-enter the learning path, or a cap on per-iid learn entries that short-circuits once memo shows `N` prior observations of the same target.

Happy to file the agent-level idempotency concern separately (same class as #227 for planning peers) if that's a cleaner scope than landing it inside this runtime investigation.

## Environment

- agentis `v1.4.3` (runtime)
- dev-apprenticeship `v0.3.0` (federation)
- macOS arm64
- Long-running dev federation against an X GitLab project with `IMPLEMENTATION_TRIGGER_LABEL=DEV::in progress` (3 matching issues, one of which is MR 180's parent issue — hence the repeat)

ylohnitram:
## Summary

- Adds `should_learn_from_mr(mr_iid)` to each of the four implementation-colony agents (`code_writer`, `test_writer`, `refactorer`, `commit_composer`). Single-key memo per agent (`<agent>:last_learned_mr_iid`), no TTL — each distinct merged MR is learned from at most once per daemon lifetime.
- Adds `tools/test-learn-idempotency.sh` as an auto-discovered colony-lint regression test (20 grep-based checks across the 4 agents).
- CHANGELOG `### Fixed` entry for #239 under `[Unreleased]`.

Plan was [posted on the issue](https://github.com/Replikanti/agentis-colonies/issues/239#issuecomment-4284791110) and approved before coding.

## Root cause (short version)

`merge-requests --since <last_check>` with GitLab's `updated_after` filter keeps returning the same MR at index `[0]` as long as that MR's `updated_at` keeps getting bumped (new comment, pipeline event, label change). Every tick re-entered the learn-path prompt at `cb_budget` 600–2000 per agent, producing hundreds of duplicate `Learned from MR <N>` log entries per hour on a real long-lived trigger-labeled issue. That load is what precedes the silent agent-daemon deaths described in the issue.

Pattern follows the existing `should_draft_code` (#200) idiom: `recall_latest` → compare iid → return bool, with `memo_write` stashed inside the gate body before any downstream subprocess / prompt / commit so failures don't re-burn the LLM budget.

## Out of scope (upstream in agentis-core)

Runtime-layer concerns from the issue's *What would help* section stay upstream and will be filed separately:

1. Terminal `daemon.stopped` event on SIGKILL-by-OS.
2. Per-agent RSS / heap instrumentation surfaced via `agentis daemon list --json`.
3. Watchdog auto-restart when pidfile goes stale without a graceful stop.

This colony-side fix removes the load that triggers the class of runtime failure. The upstream items make future load problems visible and recoverable.

## Verification

Run locally in the worktree:

```
./tools/test-learn-idempotency.sh   # 20/0
./tools/colony-lint.sh              # 81/0 (was 80, +1 new auto-discovered test)
```

## Test plan

- [ ] CI green (colony-lint passes, new test picked up by its tools/ test-*.sh loop)
- [ ] Spawn `agentis-qa-agent` once CI is green
- [ ] On approval: merge, then file the three upstream concerns as separate agentis-core issues

Closes #239.

### DIFF
--- a/dev-apprenticeship/CHANGELOG.md
+++ b/dev-apprenticeship/CHANGELOG.md
@@ -25,6 +25,26 @@ is asserted until multi-version CI is in place.
 
 ### Fixed
 
+- **Implementation agents: MR-level idempotency gate on the learning path**
+  ([#239](https://github.com/Replikanti/agentis-colonies/issues/239)).
+  `code_writer`, `test_writer`, `refactorer`, and `commit_composer` each
+  gain a `should_learn_from_mr(mr_iid)` gate that memoizes the last MR
+  iid learned from and short-circuits before the `mr-changes` /
+  `mr-commits` subprocess, the LLM `prompt()`, and the `learn()` call
+  when the same MR is seen again. Without the gate, `merge-requests
+  --since <last_check>` kept returning the same MR at index `[0]` as
+  long as its `updated_at` kept getting bumped (new comment, pipeline
+  event, label change) — at the implementation colony's `cb_budget`
+  (600 – 2000 per agent), that produces hundreds of duplicate `Learned
+  from MR <N>` entries per hour on a long-lived issue and the memory
+  load that precedes the silent agent-daemon deaths described in the
+  issue. Memo keys: `<agent>:last_learned_mr_iid`. Single-key, no TTL —
+  we want at-most-once per distinct MR iid per daemon lifetime.
+  Upstream concerns (terminal `daemon.stopped` on SIGKILL-by-OS,
+  per-agent RSS instrumentation, watchdog auto-restart on silent death)
+  remain open in `agentis-core`; this colony-side fix removes the load
+  that triggers the class of runtime failure.
+
 ### Security
 
 ## [0.3.0] — 2026-04-20
--- a/dev-apprenticeship/implementation/agents/code_writer.ag
+++ b/dev-apprenticeship/implementation/agents/code_writer.ag
@@ -71,6 +71,29 @@ fn should_draft_code(iid_str: string, upd: string) -> bool {
     return false;
 }
 
+// #239: learn from each merged MR at most once per daemon lifetime.
+// Without this gate, the `merge-requests --since <last_check>` query
+// keeps returning the same MR at index [0] as long as its `updated_at`
+// keeps getting bumped (new comments, pipeline events, label changes),
+// so the learning-path prompt re-fires every tick at cb_budget 2000 —
+// hundreds of duplicate `Learned from MR <N>` log entries and an
+// OOM-class load on long-lived MRs. Single-key memo, no TTL: we want
+// at-most-once per distinct MR iid.
+fn should_learn_from_mr(mr_iid: int) -> bool {
+    if mr_iid <= 0 {
+        return false;
+    };
+    let iid_str = to_string(mr_iid);
+    let last_iid = recall_latest("code_writer:last_learned_mr_iid");
+    if len(last_iid) == 0 {
+        return true;
+    };
+    if iid_str != last_iid {
+        return true;
+    };
+    return false;
+}
+
 fn tick(reason: string) -> void {
     // 1. Learn from recently merged MRs
     let cmd = merged_mr_cmd();
@@ -84,7 +107,8 @@ fn tick(reason: string) -> void {
     if len(raw) > 10 {
         let mr_iid = parse_int(to_string(json_get(raw, "[0].iid")));
 
-        if mr_iid > 0 {
+        if should_learn_from_mr(mr_iid) {
+            memo_write("code_writer:last_learned_mr_iid", to_string(mr_iid));
             let changes_cmd = "$COLONY_DIR/scripts/gitlab-api.sh mr-changes " + to_string(mr_iid);
             let changes_raw = try { exec sh changes_cmd; } catch e { ""; };
 
--- a/dev-apprenticeship/implementation/agents/commit_composer.ag
+++ b/dev-apprenticeship/implementation/agents/commit_composer.ag
@@ -42,6 +42,27 @@ fn get_confidence() -> float {
     return 0.0;
 }
 
+// #239: learn from each merged MR at most once per daemon lifetime.
+// Without this gate, the `merge-requests --since <last_check>` query
+// keeps returning the same MR at index [0] as long as its `updated_at`
+// keeps getting bumped, so the learning-path prompt re-fires every
+// tick at cb_budget 600. Single-key memo, no TTL: at-most-once per
+// distinct MR iid.
+fn should_learn_from_mr(mr_iid: int) -> bool {
+    if mr_iid <= 0 {
+        return false;
+    };
+    let iid_str = to_string(mr_iid);
+    let last_iid = recall_latest("commit_composer:last_learned_mr_iid");
+    if len(last_iid) == 0 {
+        return true;
+    };
+    if iid_str != last_iid {
+        return true;
+    };
+    return false;
+}
+
 fn tick(reason: string) -> void {
     // 1. Learn commit conventions from recently merged MRs
     let cmd = merged_mr_cmd();
@@ -55,7 +76,8 @@ fn tick(reason: string) -> void {
     if len(raw) > 10 {
         let mr_iid = parse_int(to_string(json_get(raw, "[0].iid")));
 
-        if mr_iid > 0 {
+        if should_learn_from_mr(mr_iid) {
+            memo_write("commit_composer:last_learned_mr_iid", to_string(mr_iid));
             let commits_cmd = "$COLONY_DIR/scripts/gitlab-api.sh mr-commits " + to_string(mr_iid);
             let commits_raw = try { exec sh commits_cmd; } catch e { "[]"; };
 
--- a/dev-apprenticeship/implementation/agents/refactorer.ag
+++ b/dev-apprenticeship/implementation/agents/refactorer.ag
@@ -77,6 +77,27 @@ fn should_handle_refactor(iid_str: string) -> bool {
     return false;
 }
 
+// #239: learn from each merged MR at most once per daemon lifetime.
+// Without this gate, the `merge-requests --since <last_check>` query
+// keeps returning the same MR at index [0] as long as its `updated_at`
+// keeps getting bumped, so the learning-path prompt re-fires every
+// tick at cb_budget 1000. Single-key memo, no TTL: at-most-once per
+// distinct MR iid.
+fn should_learn_from_mr(mr_iid: int) -> bool {
+    if mr_iid <= 0 {
+        return false;
+    };
+    let iid_str = to_string(mr_iid);
+    let last_iid = recall_latest("refactorer:last_learned_mr_iid");
+    if len(last_iid) == 0 {
+        return true;
+    };
+    if iid_str != last_iid {
+        return true;
+    };
+    return false;
+}
+
 fn tick(reason: string) -> void {
     // 1. Learn from refactoring patterns in merged MRs
     let cmd = merged_mr_cmd();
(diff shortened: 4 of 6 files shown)


ylohnitram:
## QA Report — PR #242 (issue #239, MR-learning idempotency gate)

### Verification summary

| Item | Result |
|---|---|
| `./tools/colony-lint.sh` | 81 passed, 0 failed, 5 skipped (shellcheck not installed locally) |
| `./tools/test-learn-idempotency.sh` | 20/20 passed |
| `./tools/check-changelog.sh` (with `GITHUB_BASE_REF=main`) | CHANGELOG consistent with diff |
| `./tools/check-prompt-gate.sh` | exit 0 (all memo-gated) |
| Implementation agents audited (MR-learning path) | 4/4 (`code_writer`, `test_writer`, `refactorer`, `commit_composer`) — exhaustive |
| Gate memo keys, copy-paste consistency | All four correctly namespaced as `<agent>:last_learned_mr_iid` |
| Tick N+1 walkthrough (MR 180 unchanged) | Gate returns false → no subprocess, no `prompt()`, no `learn()` (correct) |
| Tick N+2 walkthrough (new MR 181 merged) | Gate returns true → memo updated to `"181"`, subprocess + prompt fire once (correct) |
| `memo_write` before `prompt()` / subprocess (no re-burn on failure) | Yes — line-exact in all four agents |

### Findings

|---|---|---|---|---|
| 1 | none | correctness | Gate logic correct for the stated bug. At-most-once per distinct MR iid per daemon lifetime holds — walkthrough: next tick on unchanged top MR returns `iid_str == last_iid` → gate returns false and all three cost drivers (`mr-changes` subprocess, `prompt()` at `cb 600–2000`, `learn()` write) are skipped. | - [x] verified |
| 2 | none | correctness | Memo-write placement semantically safe. `memo_write(...:last_learned_mr_iid, ...)` is the **first** statement inside the gate body, before the subprocess and before `prompt()`. A `mr-changes` subprocess failure, a `prompt()` exception, or the inner `len(changes_raw) > 10` / `patterns.test_files_seen > 0` / `patterns.refactor_mrs > 0` / `style.commits_seen > 0` negatives all leave the memo persisted — next tick on the same MR short-circuits. | - [x] verified |
| 3 | none | scope completeness | Scope is exhaustive relative to the bug. Only implementation has a ticking MR-learning path without an existing gate. code-review's 4 reviewers already have the `last_pattern_learn_ts` TTL gate from issue #201, and release agents don't run a learn prompt on the top MR per tick. Four agents fixed matches four agents at risk. | - [x] verified |
| 4 | none | test coverage | Test script is auto-discovered by `colony-lint.sh` (via `tools/test-*.sh` loop) and lands under `[PASS] tools: test-learn-idempotency.sh unit tests`. 5 checks × 4 agents = 20 assertions, all pass. | - [x] verified |
| 5 | LOW | defense-in-depth (non-blocking, not a regression) | **At-most-once-per-distinct-iid is strict.** If `prompt()` on MR 180 throws transiently (OAuth hiccup, upstream 5xx), the memo is already stamped and the daemon will never re-attempt learning from MR 180 during this lifetime. That's the intentional tradeoff documented in the CHANGELOG and the commit body — "we want the load off now, don't re-burn on failure." It's the right tradeoff for the class of failure this PR targets (thousands of duplicate learns/hr is objectively worse than losing one learning per failed MR). Not a blocker; could be revisited later with a bounded retry counter if data shows too many learn-misses. | - [ ] noted, not required |
| 6 | LOW | test strictness (non-blocking) | `test-learn-idempotency.sh` does not assert the *ordering* invariant "`memo_write` before `prompt()`" — only that both exist. A future refactor could move the memo-write after a failure-prone call and still pass. Stronger check (acceptable follow-up, not required here): grep for `memo_write("<agent>:last_learned_mr_iid"` appearing on a lower line number than the first `prompt(` within the gate body. | - [ ] follow-up, optional |
| 7 | LOW | test strictness (non-blocking) | The gate signature check `fn should_learn_from_mr(mr_iid: int) -> bool {` is signature-locked. A harmless parameter rename (e.g. `iid`) would false-fail. Fine for an idiom-enforcement check; flagging it only because the `check-prompt-gate.sh` style of "gate by idiom, not signature" is the prior art in this repo. | - [ ] follow-up, optional |
| 8 | none | re-learning from same MR across runtime concerns | First-principles: should the agent ever re-learn from the same MR if its diff genuinely changes? **No** — a merged MR's diff is immutable (force-push post-merge is rare and disruptive enough that treating it as a new MR would not be the right behaviour). The only `updated_at` bumps post-merge are comments, pipeline events, label changes — none of which change what the learner observes. So the at-most-once-per-iid rule is semantically correct, not just a load-reduction hack. | - [x] verified, no change needed |
| 9 | none | CHANGELOG scoping | `### Fixed` entry correctly scopes this as colony-side load reduction, names the three upstream runtime concerns (terminal `daemon.stopped`, RSS instrumentation, watchdog auto-restart), and avoids overclaiming that this single-handedly fixes the silent-death class of runtime failure. Idiomatic continuation of the #200 / #201 memo-gate series. | - [x] verified |

### Verdict

**Ship it.** The fix does what the PR body claims. The load reduction is real and the semantics are sound from first principles (not just load-chasing — merged-MR diffs are immutable, so at-most-once is genuinely the right rule). The colony-side scope is exactly right: the runtime concerns listed in the PR body belong upstream and are intentionally out of scope. Colony-lint clean, new test suite passes 20/20, CHANGELOG entry idiomatic.

Three LOW-severity notes in the table above are follow-up opportunities, not blockers — they describe possible defense-in-depth refinements to the test script and a theoretical behaviour to reconsider later if retry-loss-per-failure becomes a measurable problem.

### Files reviewed

- `dev-apprenticeship/implementation/agents/code_writer.ag`
- `dev-apprenticeship/implementation/agents/test_writer.ag`
- `dev-apprenticeship/implementation/agents/refactorer.ag`
- `dev-apprenticeship/implementation/agents/commit_composer.ag`
- `tools/test-learn-idempotency.sh`
- `dev-apprenticeship/CHANGELOG.md`
- `dev-apprenticeship/code-review/agents/logic_reviewer.ag` (cross-check — confirmed out of scope, already gated via #201)

ylohnitram:
## Addressed QA findings

### #6 — ordering assertion (addressed in 6eaf6a7)

`test-learn-idempotency.sh` now asserts the line-number invariant `gate_line < memo_line < prompt_line` for each agent. Finds the first `prompt(` after the `if should_learn_from_mr(` line and checks that the matching `memo_write("<agent>:last_learned_mr_iid", ...)` falls between them. A future refactor that moves the memo-write below prompt() will fail the test.

Output on current HEAD (all four agents):

```
ok — memo_write before first prompt() inside gate body (lines 110 < 111 < 116)
ok — memo_write before first prompt() inside gate body (lines 116 < 117 < 122)
ok — memo_write before first prompt() inside gate body (lines 114 < 115 < 123)
ok — memo_write before first prompt() inside gate body (lines 79 < 80 < 85)
```

### #7 — loose signature check (addressed in 6eaf6a7)

The gate-function existence check is now `grep -Eq "^fn should_learn_from_mr\("` instead of the full `"fn should_learn_from_mr(mr_iid: int) -> bool {"` signature. Anchors to a definition at column 0 (catches "is this actually a function def, not a call"), locks the function name, and ignores the parameter name and annotations. Matches the `check-prompt-gate.sh` prior-art idiom of gating by shape rather than exact signature.

Test count: 20 → 24 (one new ordering check per agent). Colony-lint remains green.

### #5 — retry semantics, kept as intentional

Leaving as-is. The argument:

1. **Merged-MR diffs are immutable.** Once an MR is merged, its diff will not change — the only things that bump `updated_at` post-merge are comments, pipeline events, label changes. So "miss one learn on a transient prompt() failure" literally means losing one opportunity to ingest a pattern from that specific MR; the diff itself is not retryable information.
2. **Retry introduces the bug it prevents.** Any bounded retry mechanism (a per-iid attempt counter memo) has to decide when the counter resets. Reset on any `updated_at` change and we're back to re-entering the prompt every tick on `updated_at`-bumping MRs. Reset only on new-iid and the retry buys us at most N attempts *across the daemon's entire lifetime*, which isn't meaningfully different from zero.
3. **Data-driven decision.** If the loss rate turns out to be measurable in production (a signal we currently don't have), a bounded retry *with a cold start* after N days is straightforward to add on top of the current gate. Right now there is no evidence it's a problem, and the complexity cost of retry infrastructure outweighs the hypothetical benefit.

Happy to file as a follow-up issue for explicit tracking if you'd prefer a place to accumulate production signal against this decision.

Issue #195, opened 2026-04-18
feedback loop: .ag agents need reality-check pattern to produce honest fitness signal


ylohnitram:
## Problem

Every `learn()` call in every agent across all 21 `.ag` files currently emits `outcome="success"` or `"partial"` on the happy path. No agent ever emits `outcome="fail"`, `outcome="reject"`, `verdict="reject"`, or `rejected=true`. This is not a coding oversight — it is a **structural** consequence of missing feedback: the agent has no way to learn that its past action was wrong, so it defaults to optimism.

Concrete example — `triage/labeler`:
1. Tick N: agent sees new issue, suggests `bug` label, emits on bus + writes `learn(..., outcome="success", tags=["emitted", "triage"])`.
2. Human (or another agent) sees the suggestion, decides it's wrong, removes the `bug` label or replaces it.
3. Tick N+1: labeler processes the next issue. It has no awareness that its last suggestion was rejected. Every experience row says "success" regardless of reality.

## Impact on #186's auto-promote heuristic

[#186](https://github.com/Replikanti/agentis-colonies/issues/186) / [#192](https://github.com/Replikanti/agentis-colonies/pull/192) fixed the *classification* of experience rows (acting vs. observe), so the heuristic now correctly ignores observe-step noise. But:

- `reject_rate_acting` is **structurally ≡ 0** because no agent produces reject markers → gate never fires, never brakes bad promotions.
- `delta_slope_acting` is always non-negative because `fitness_delta ∈ {+0.15, +0.02}` for every acting row → slope is flat-positive forever → gate always passes.
- `evolve_reject_above = 0.20` and `delta_slope_negative_for = 1000` triggers likewise never fire.

Net effect: the only gates that actually brake promotion today are `entries_total` / `entries_acting` / `runtime_hours`. An agent that makes catastrophically wrong suggestions is indistinguishable from one making perfect ones, as long as it ticks long enough and emits in acting mode often enough.

## Proposed pattern: reality-check tick

Add a periodic \"reality check\" branch to each agent's `tick()`. The branch:
1. Reads back a past suggestion memo (what the agent proposed at tick N-K for some artefact X).
2. Queries the current GitLab state of X (label / MR status / comment thread / whatever is relevant).
3. Compares: does reality match the suggestion?
[...]

Issue #186, opened 2026-04-17
auto-promote heuristic counts observe-step rows as tier evidence — recommends promotion without proof acting branch ever fired


ylohnitram:
## Summary

`tools/auto-promote.sh` (layer-1 auto-promote) recommends promoting an agent up a tier using a fitness signal that is dominated by **observe-step** `learn()` rows — rows that fire regardless of tier and always succeed. The agent can therefore collect hundreds of "success" rows without ever exercising the acting branch of its **current** tier, yet still receive a ✓ promote recommendation for the **next** tier.

Result: the dashboard surfaces high-confidence promote suggestions for agents that have never demonstrated the behaviour the target tier permits.

## Concrete example (dev-apprenticeship federation, today)

All three planning agents are on confidence `0.8` (review-gated tier). Dashboard recommendations:

```
✓ risk_assessor    421 entries, 100% success, +0.000 slope, on 0.80 → autonomous 0.95
✓ scope_estimator  285 entries, 100% success, +0.000 slope, on 0.80 → autonomous 0.95
✓ task_decomposer  422 entries, 100% success, +0.000 slope, on 0.80 → autonomous 0.95
```

Experience-store inspection:

| Agent | Total rows | Tagged `observed` | Tagged `review-gated` / `acted` / `emitted` |
|---|---:|---:|---:|
| risk_assessor    | 425 | 425 | **0** |
| scope_estimator  | 288 | 288 | **0** |
| task_decomposer  | 426 | 426 | **0** |

And `.agentis/memo/` contains **zero** `plan_reviewer:*:*_drafted` markers, meaning the review-gated acting path has never written a draft.

In other words: the agents have produced hundreds of `"observed"` rows from the unconditional observe step (`learn(topic, ..., "success", ["observed", ...])` at e.g. risk_assessor.ag:57), and **nothing** from the tier-gated acting branch that 0.8 is supposed to enable. The auto-promote heuristic still says ✓ promote to 0.95.

The "100% success" signal is meaningless here: observe-step rows are hardcoded to `"success"` regardless of quality, and the tier-gated acting branch has never executed. The agent could be on `0.5` or `0.95` and the observe-row stream would look identical — so the signal cannot distinguish tiers at all.

## Root cause (hypothesis)

`tools/auto-promote.sh` reads `.agentis/experience/*.jsonl`, aggregates per-agent `outcome == "success"` over a window, and emits ✓ when the rate clears a threshold. It does not appear to condition on:

[...]

Issue #10, opened 2026-04-10
planning gitlab-api.sh ships unused endpoints


ylohnitram:
## Problem

`dev-apprenticeship/planning/scripts/gitlab-api.sh` implements endpoints that are not called by any agent:

- `update-issue` — no callers
- `issue-events` — no callers
- `labels` — no callers

```
$ grep -rn 'update-issue\|issue-events\|labels' dev-apprenticeship/planning/agents/
(no matches)
```

## Impact

Dead code. Either these endpoints are planned for future use (in which case they should be marked as such in a comment with a link back to the agent that will use them) or they are speculative and should be dropped. Shipping a "full" API wrapper before agents need it is not how the code-review colony's wrapper was built, and the inconsistency makes the surface area harder to reason about.

## Scope

`dev-apprenticeship/planning/scripts/gitlab-api.sh`

The code-review colony's wrapper may have a similar dead-code pattern and should be checked in the same pass.

## Related

Reported by QA on #8 as MEDIUM-3.

ylohnitram:
## Decision

Delete the three unused endpoint blocks (`update-issue`, `issue-events`, `labels`) from `dev-apprenticeship/planning/scripts/gitlab-api.sh`, along with the now-orphaned `gl_put` helper and the `json_field` helper (only consumer was `update-issue`). Update the file's usage header to match.

## Scope

Single file: `dev-apprenticeship/planning/scripts/gitlab-api.sh`.

- Header usage lines 14, 16, 17 — drop the three usage entries.
- `gl_put` function lines 45-52 — delete.
- `json_field` helper lines 54-59 — delete if the remaining `post-note` case is refactored to inline the same python3 one-liner that `code-review/scripts/gitlab-api.sh` already uses at its own line ~102. Otherwise keep `json_field` for `post-note`. Decide during implementation, prefer inlining to match the sibling colony.
- `update-issue` case lines 120-136 — delete.
- `issue-events` case lines 97-100 — delete.
- `labels` case lines 138-140 — delete.

Search anchors (line numbers may drift as edits stack): `gl_put()`, `json_field()`, `update-issue)`, `issue-events)`, `labels)`.

## Concrete fix

After edits, the `case "$CMD" in ... esac` block in the planning file should contain only: `issues`, `issue`, `issue-notes`, `post-note`, and the `*)` unknown-command fallback. The header comment block must list exactly those commands. The file should shrink from 147 lines to roughly 95-100.

[...]

PR #243, opened 2026-04-21
feedback-loop: extend reality-check to autonomous-tier writes (#203)

### LINKED ISSUE #203: feedback-loop: extend reality-check to autonomous-tier emissions
Follow-up from #195 / PR #202 (QA finding #15).

## Problem

The reality-check pattern introduced by #195 operates only at the `propose` and `review-gated` tiers for the labeler pilot. The autonomous branch (at `dev-apprenticeship/triage/agents/labeler.ag:254` on the feat/195-feedback-loop branch, or the analogous line on main once merged) still emits:

```
learn("label", "issue " + to_string(action.issue_id), action.labels, "success", [scope, "triage", "acted"]);
```

on the happy path — no reality check. Rationale documented in [`doc/feedback-loop.md`](../doc/feedback-loop.md#out-of-scope): at autonomous the agent writes labels to GitLab itself, so there is no separable operator signal "did they agree with my suggestion" — the agent and the operator are effectively the same voice at that point.

But the operator can still **revert** the autonomous write: remove the label, replace it, or close the issue. That reversal is the ground-truth "you were wrong" signal for autonomous-tier actions, just on a longer horizon.

## Proposed pattern: longer-horizon autonomous check

Analogous to the existing propose-tier memo, but with different cadence and triggers:

1. **Stash autonomous verdict** after each autonomous-tier `update-issue --add-labels` call. Memo: `labeler:autonomous_verdict:<iid>` → `[ts, "labels_csv"]`.
2. **Scan autonomous verdicts** at start of each tick. For each verdict older than N minutes (e.g. 30) and younger than M hours (e.g. 48):
   - Query `get-issue <iid>` for current state.
   - Compare: are our labels still there?
   - Full overlap → `outcome="success"` (`["acted", "triage"]`).
   - Partial erosion (some removed) → `outcome="partial"`.
   - All removed or replaced → `outcome="fail"` (the reversal signal).
3. **Clear verdict** after scoring or after M hours (ageout).
4. **Tie-break for multi-label writes**: one verdict per `(iid, tick)` tuple; stale verdicts from earlier writes on the same iid overwrite.

## Why separate issue

- The propose / review-gated pattern is about "did the operator apply my suggestion?". The autonomous pattern is about "did the operator revert my action?". Different semantic, different cadence (propose: minutes; autonomous: hours).
- Requires a multi-slot memo schema (one pending verdict per iid) rather than the single-slot schema the pilot uses. That's a non-trivial schema change.
- Currently has low urgency: no labeler in the federation is at autonomous tier today because auto-promote's `0.8 → 0.95` step requires ≥ 120 clean acting rows (per `doc/auto-promote.md#0.8---0.95-review-gated---autonomous`), and #195 has not yet produced any rows. This follow-up lands naturally after the pilot accumulates propose-tier data.

## Scope

**In scope:**
- Implement for labeler (second pilot), mirroring the #195 shape.
- Update `doc/feedback-loop.md` to replace the "autonomous out-of-scope" caveat with a reference to the autonomous pattern.
- Document the multi-slot memo schema in the doc.

**Out of scope:**
- Fan-out to the other 20 agents (same as #195, one follow-up per agent).
- Tying the autonomous reject signal to auto-promote's demote path — let the data land first; calibrate later.

## Success criteria

- Autonomous labeler action followed by operator label removal within 48 h produces a single `outcome="fail"` row in the experience store with tag `["<scope>", "triage", "acted"]`.
- Autonomous action not reverted within 48 h produces a single `outcome="success"` row after the 48 h ageout.
- No double-counting: a single autonomous write produces exactly one post-hoc verdict row regardless of how many ticks elapsed.

## Related

- #195 — parent pattern (propose / review-gated).
- #186 / #192 — classifier the honest signal lands into.
- #163 — calibration follow-up that consumes both reject sources.

ylohnitram:
## Summary

Extends the [#195](https://github.com/Replikanti/agentis-colonies/issues/195) reality-check pattern to the labeler's **autonomous** branch. Closes #203 as a labeler-pilot — fan-out to the other 20 agents is tracked in follow-ups per agent (each needs its own ground-truth signal defined).

## What's different from the propose path

The propose / review-gated path scores against *"did the operator apply our suggestion?"* — signal is separable because the operator's reaction happens *after* the agent's suggestion. Autonomous is different: the agent *is* the writer, so the separable signal becomes *"did the operator revert the write?"* — and that takes time to settle.

- **Multi-slot memo** (one blob per in-flight iid + an index CSV) instead of single-slot, so multiple parallel autonomous writes can soak simultaneously.
- **Soak window: 30 min.** An autonomous write is scored no earlier than 30 minutes after it lands.
- **Ageout: 48 h.** Longer than the propose-path 24 h to match the slower operator-response horizon on already-applied labels.
- **Compare has no signal==0 arm.** We wrote the labels, so empty label set = reversal, not silence. Strict 1/2/3 output.
- **Two-row pattern per autonomous action.** At-write `learn("success", ..., "acted")` preserves the acting-path fitness signal [#186](https://github.com/Replikanti/agentis-colonies/issues/186) already consumes; post-soak `learn(<outcome>, ..., "acted")` lands in the same tag bucket and averages in — so a consistently-reverted agent sees its acting fitness drift down despite at-write optimism.

## Ordering invariant

`record_autonomous_verdict(...)` is called **before** the bare at-write `learn("success", ..., "acted")`. If the memo write fails, the `learn()` still lands (acceptable asymmetry — slightly optimistic fitness, which the overall two-row pattern is designed to correct over time). If `learn()` fails after a successful memo write, we get the soak row with no at-write row (also fine — soak row is the honest one).

Enforced by `tools/test-labeler-autonomous-verdict.sh` (awk line-ordering check between `if my_tier == \"autonomous\"` and the first matching `learn()`).

## Memo schema

```
labeler:autonomous_verdict:<iid>  -> "[ts, iid, \"labels_csv\"]"
labeler:autonomous_verdict_index  -> CSV of iids with pending blobs
```

Index is the only structure the scanner reads; blobs are read per-iid. A missing blob (index drift) triggers a self-heal that drops the iid from the index without emitting a learn row.

## Changes

- `dev-apprenticeship/triage/agents/labeler.ag` — 6 new functions (`append_autonomous_index`, `record_autonomous_verdict`, `score_one_autonomous`, `append_csv`, `eval_autonomous_at`, `evaluate_autonomous_verdicts`). Scanner wired into `tick()` next to the existing `evaluate_label_verdict()`. Autonomous branch records the verdict before the at-write `learn()`. Iterates via tail recursion (runtime convention — `.ag` has no loops).
- `doc/feedback-loop.md` — remove the old "autonomous out of scope" bullet, add a full autonomous-tier-extension section covering memo schema, soak / ageout rationale, two-row pattern, signal interpretation, and the ordering invariant.
- `tools/test-labeler-autonomous-verdict.sh` — structural regression (18 assertions: entry points, memo keys, tick() wiring, ordering invariant, propose path intact, signal==0 arm absent in autonomous compare, soak / ageout constants).
- `dev-apprenticeship/CHANGELOG.md` — entry under `[Unreleased]`.

## Test plan

- [x] `./tools/colony-lint.sh` — 83 passed, 0 failed, 5 skipped (shellcheck skips; CI runs it)
- [x] `./tools/test-labeler-autonomous-verdict.sh` — 18 / 18
- [x] Existing `test-learn-idempotency.sh` / `test-dashboard-*` / `test-kill-*` — all green (untouched surface)
- [x] CI green across all checks

### DIFF
--- a/dev-apprenticeship/CHANGELOG.md
+++ b/dev-apprenticeship/CHANGELOG.md
@@ -17,6 +17,29 @@ is asserted until multi-version CI is in place.
 
 ### Added
 
+- **Labeler autonomous-tier reality check (pilot)**
+  ([#203](https://github.com/Replikanti/agentis-colonies/issues/203)).
+  Extends the #195 reality-check pattern to the labeler's autonomous
+  branch, where the agent writes labels to GitLab directly and the
+  ground-truth signal is "did the operator revert the write?" rather
+  than "did the operator apply our suggestion?". New memo schema is
+  multi-slot per-iid (`labeler:autonomous_verdict:<iid>` + an index
+  CSV `labeler:autonomous_verdict_index`) so multiple in-flight writes
+  can soak in parallel; the single-slot propose-path idiom remains
+  untouched. Soak window 30 min, ageout 48 h (longer than the propose
+  path's 24 h to match the slower human-response horizon on an
+  already-applied label). Two-row pattern per autonomous action:
+  at-write `learn("success", ..., "acted")` preserves the existing
+  acting-path fitness signal for #186, post-soak
+  `learn(<outcome>, ..., "acted")` lands in the same tag bucket and
+  averages in — so a consistently-reverted agent sees its acting
+  fitness drift down despite at-write optimism. Full pattern
+  documented in
+  [`doc/feedback-loop.md`](../doc/feedback-loop.md#autonomous-tier-extension-203-labeler-pilot);
+  structural regression in `tools/test-labeler-autonomous-verdict.sh`.
+  Fan-out to the other 20 agents is tracked per-agent in follow-up
+  tickets.
+
 ### Changed
 
 ### Deprecated
--- a/dev-apprenticeship/triage/agents/labeler.ag
+++ b/dev-apprenticeship/triage/agents/labeler.ag
@@ -193,12 +193,183 @@ fn evaluate_label_verdict() -> void {
     memo_write("labeler:pending_verdict", "");
 }
 
+// #203: longer-horizon reality check for autonomous-tier writes.
+//
+// The propose / review-gated path above uses a single-slot
+// `labeler:pending_verdict` blob that the next tick overwrites. That
+// shape is fine there because the signal we want ("did the operator
+// apply our suggestion?") settles in minutes and a stale entry is
+// never wrong for long.
+//
+// Autonomous is different. The agent writes labels directly to GitLab,
+// and the ground-truth signal is "did the operator REVERT my write?"
+// — a question that only makes sense after the operator has had time
+// to notice and react. We soak each autonomous write for at least 30
+// minutes before scoring, and accept the verdict as stale ("aged out
+// without signal") after 48 hours. During that window the agent may
+// emit further autonomous writes for other issues, so we need one
+// slot per iid, not the single-slot idiom the propose path uses.
+//
+// Memo schema:
+//   labeler:autonomous_verdict:<iid>  -> "[ts, iid, \"labels_csv\"]"
+//   labeler:autonomous_verdict_index  -> CSV of iids with pending blobs
+//
+// The index is the only structure the per-tick scanner reads to
+// decide which iids to check. Blobs are read per-iid inside the
+// scanner, and a missing blob (drift) triggers a self-heal that drops
+// the iid from the index without emitting a learn row.
+
+// Append an iid to the autonomous-verdict index, deduping so the same
+// iid never appears twice. We store the freshest occurrence at the
+// tail (matches the "latest write wins" semantic for repeated writes
+// on the same issue).
+fn append_autonomous_index(iid: int) -> void {
+    let existing = recall_latest("labeler:autonomous_verdict_index");
+    let iid_str = to_string(iid);
+    let append_cmd = "EX=" + shell_escape(existing) + " N=" + shell_escape(iid_str) +
+        " python3 -c 'import os; ex=os.environ[\"EX\"]; n=os.environ[\"N\"]; parts=[p for p in ex.split(\",\") if p.strip() and p.strip()!=n]; parts.append(n); print(\",\".join(parts))'";
+    let next = try { exec sh append_cmd; } catch e { iid_str; };
+    memo_write("labeler:autonomous_verdict_index", next);
+}
+
+// Record a pending autonomous verdict for a specific iid. Called
+// immediately after a successful `update-issue --add-labels` in the
+// autonomous branch. Writing the blob BEFORE the at-write `learn()`
+// (see tick()) means a `learn()` failure does not leave a phantom iid
+// in the index; an `append_autonomous_index` failure means the blob
+// is unreachable from the scanner and gets GC'd on drift-cleanup when
+// it eventually expires.
+fn record_autonomous_verdict(issue_id: int, labels_csv: string) -> void {
+    let ts_raw = try { exec sh "date +%s"; } catch e { "0"; };
+    let blob = "[" + ts_raw + "," + to_string(issue_id) + ",\"" + labels_csv + "\"]";
+    memo_write("labeler:autonomous_verdict:" + to_string(issue_id), blob);
+    append_autonomous_index(issue_id);
+    print("[labeler] autonomous verdict recorded: issue", issue_id, "labels", labels_csv);
+}
+
+// Score a single pending autonomous verdict. Returns "keep" when the
+// verdict is still soaking (< 30 min) and should remain in the index
+// for a future tick; returns "drop" when the verdict is scored,
+// aged out (> 48 h), or was drift-orphaned (blob missing). Only the
+// scored-and-emitted path appends to the experience store.
+fn score_one_autonomous(issue_id: int) -> string {
+    let key = "labeler:autonomous_verdict:" + to_string(issue_id);
+    let blob = recall_latest(key);
+    if len(blob) < 3 {
+        // Drift: index references a blob that was already cleared or
+        // never written (the `record` path writes blob first, then
+        // appends to index — if `append` failed the blob exists but
+        // the index never held this iid; if we got here, the index
+        // did hold it, so the blob was cleared by an earlier scan).
+        return "drop";
+    };
+    let age = verdict_age_seconds(blob);
+    let soak_s = 1800;
+    if age < soak_s {
+        // Still soaking — operator hasn't had time to react. Leave
+        // the blob and keep the iid in the index for the next tick.
+        return "keep";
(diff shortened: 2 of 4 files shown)


ylohnitram:
## QA Report — PR #243 (#203 autonomous verdict)

### Verdict: ship

Priority-1-to-10 invariants from the PR notes all verified. No correctness bugs; three non-blocking LOW findings (1 docs, 1 theoretical correctness hardening, 1 test coverage). All three **addressed inline in 82832c2** — see addendum.

### Summary
- `./tools/colony-lint.sh`: 83 passed, 0 failed, 5 skipped (shellcheck skips — CI runs it)
- `./tools/test-labeler-autonomous-verdict.sh`: 18/18 (pre-fix) → 20/20 (post-fix)
- GitHub Actions `colony-lint`: green on both commits
- check-exec-sh: 0 findings on the new surface
- `agentis` parser accepts `labeler.ag`

### Findings

| 1 | INFO | correctness | Ordering invariant `record_autonomous_verdict` < at-write `learn("success", "acted")` confirmed inside `if my_tier == "autonomous"` branch | - [x] verified |
| 2 | INFO | correctness | Propose-path single-slot `labeler:pending_verdict` idiom unchanged; `record_label_verdict()` still called in review-gated and propose branches | - [x] verified |
| 3 | INFO | correctness | Memo key prefix `labeler:autonomous_verdict:` and index key `labeler:autonomous_verdict_index` consistent across writer / reader / test | - [x] verified |
| 4 | INFO | correctness | `eval_autonomous_at` tail recursion terminates cleanly on empty `iid_raw`; no off-by-one on the final iid | - [x] verified |
| 5 | INFO | correctness | `append_autonomous_index` dedup via python3 filter runs correctly; freshest iid at tail | - [x] verified |
| 6 | INFO | correctness | Autonomous compare emits only 1/2/3 (no signal==0 arm); empty `sug` falls to 3 (fail) which is a safe corrupted-memo default | - [x] verified |
| 7 | INFO | correctness | `shell_escape()` covers all dynamic values in `append_cmd`, `match_cmd`, and the `eval_autonomous_at` position helper | - [x] verified |
| 8 | INFO | CB | Per-tick extra work bounded; ~50 CB/hop, ~8 iids/tick, fits inside `cb 400` with overflow carried forward in the index | - [x] verified |
| 9 | INFO | docs | Doc section `Autonomous-tier extension (#203, labeler pilot)` memo keys / soak (1800 s) / ageout (172800 s) / two-row pattern / signal mapping all match code | - [x] verified |
| 10 | INFO | tests | 18 structural assertions cover the regression surface | - [x] verified |
| 11 | LOW | docs | Drift-comment at `score_one_autonomous` L259-263 misdescribed the append-failure mode — `append_autonomous_index` unconditionally writes the index even on its catch branch, so "the index never held this iid" is not a reachable state | - [x] addressed in 82832c2 |
| 12 | LOW | correctness | `append_autonomous_index` catch-branch fell back to bare `iid_str`, silently clobbering any peer iids already queued in the index on a subprocess failure (python3 missing / OOM / exec_foreign revoked) | - [x] addressed in 82832c2 |
| 13 | LOW | tests | Structural test did not pin the catch-branch fallback value nor the end-of-scan index rewrite in `evaluate_autonomous_verdicts` | - [x] addressed in 82832c2 |

### Details

#### Findings 1-10 — verified invariants
- **#1 ordering:** `record_autonomous_verdict(action.issue_id, action.labels)` at L432 (now L439 post-fix) precedes `learn("label", ..., "success", [scope, "triage", "acted"])` at L433 (now L440). Enforced structurally by the test and visually by reading `tick()`.
- **#2 propose intact:** `labeler:pending_verdict` memo still written in both review-gated (L446) and propose (L458) branches via `record_label_verdict(action.issue_id, action.labels)`. `evaluate_label_verdict()` still called from `tick()` alongside the new `evaluate_autonomous_verdicts()`.
- **#3 memo key consistency:** 8 references in the agent + 3 in the test, all using the same prefix/key.
- **#4 tail-recursion terminator:** `eval_autonomous_at` returns `acc` when `iid_raw` is empty (python prints empty string on out-of-range position); `exec sh` trims the trailing newline (same pattern as `parse_int(now_raw)` for `date +%s` in the existing propose path).
- **#5 dedup:** python3 filter `[p for p in ex.split(",") if p.strip() and p.strip()!=n]` followed by `parts.append(n)` gives correct "latest wins at tail".
- **#6 no signal==0 in autonomous:** `print(1 if sug and sug.issubset(act) else (2 if sug & act else 3))` — strict 1/2/3.
- **#7 shell_escape coverage:** confirmed for all dynamic values; `to_string(issue_id)` for the int-typed iid in `cur_cmd` follows the established propose-path precedent at L157.
- **#8 CB budget:** unchanged `cb 400;` header; per the PR note estimate, ~50 CB per scan hop × ~8 iids/tick leaves headroom for the main work.
- **#9 doc alignment:** memo keys (`labeler:autonomous_verdict:<iid>`, `labeler:autonomous_verdict_index`), soak (1800 s), ageout (172800 s), two-row pattern, and strict 1/2/3 signal mapping all agree between `doc/feedback-loop.md §Autonomous-tier extension` and `labeler.ag`.
- **#10 test completeness:** 18 assertions structurally pin the regression surface (entry points, memo keys, ordering, tier-branch invariants, constants).

#### Finding 11 — Drift-comment
Replaced the speculative "if append failed → index never held this iid" rationale (which contradicted the unconditional index memo_write on the catch branch) with a concise list of realistic drift causes (external clear, prior scan cleared blob but failed to rewrite index) and a one-line action summary.

#### Finding 12 — Catch-branch clobber
```diff
-    let next = try { exec sh append_cmd; } catch e { iid_str; };
+    let next = try { exec sh append_cmd; } catch e { append_csv(existing, iid_str); };
```
Preserves any previously-queued peer iids on a subprocess failure. A brief duplicate (if `iid_str` was already in `existing`) is acceptable — the next successful scan normalizes it; a silently-lost peer is not — its blob would leak indefinitely without an index reference.

#### Finding 13 — Test hardening
Added 2 awk-scoped asserts:
- `append_autonomous_index` catch-branch contains `append_csv(existing, iid_str)` (regression guard against the fallback ever weakening back to plain `iid_str` or `""`).
- `evaluate_autonomous_verdicts` function body contains the `memo_write("labeler:autonomous_verdict_index", ...)` (regression guard if the end-of-scan rewrite ever gets moved into `eval_autonomous_at` or dropped entirely, which would turn the scanner into a no-op after the first still-soaking iid).

### Addendum — all three LOW findings addressed inline (82832c2)

Post-fix: `test-labeler-autonomous-verdict.sh` 20/20, colony-lint 83/0/5, CI green on both commits.

Issue #172, opened 2026-04-17
federation-dashboard.sh: heredoc bug at line 962 — eliminate ALL remaining heredocs


ylohnitram:
## Problem

After PR #171 (#170) fix, macOS users still hit a bash heredoc syntax error, this time at line 962 of `tools/federation-dashboard.sh`.

The remaining 7 heredocs in the script are vulnerable to the same family of macOS bash 3.2 parser bugs that #170 attempted to eliminate. The previous fix only extracted Python heredocs nested in `\$()` — but the JS/CSS/HTML heredocs still contain content (e.g. \`\\'\` JS string escapes at line 962) that confuses the parser.

## Root cause

Line 962:
\`\`\`js
return '<th onclick="doSort(\\'' + col + '\\')">' + label + '<span class="sort-arrow">' + arrow + '</span></th>';
\`\`\`

The \`\\'\` escape sequence inside a \`<<'JSEOF'\` heredoc trips macOS bash 3.2 / 5.3 parser even without \`\$()\` nesting.

## Fix

Mirror the #170 approach more aggressively: extract ALL heredocs from \`federation-dashboard.sh\`. Specifically:

1. Move CSS + HTML + JS to a single \`tools/federation-dashboard.html.template\` static file
2. Add \`tools/federation-dashboard-renderer.py\` that takes context vars and emits \`index.html\` by reading the template + filling sentinels
3. \`federation-dashboard.sh\` calls the renderer instead of \`cat\` heredocs
4. Add regression test: \`grep -c '^[[:space:]]*<<' tools/federation-dashboard.sh\` must equal 0

This eliminates the entire class of bash heredoc bugs permanently.

Issue #170, opened 2026-04-17
federation-dashboard.sh: syntax error at line 348 after v1.3.1 pull (bash heredoc not consumed)


ylohnitram:
## Summary

`tools/federation-dashboard.sh` fails to start with a bash syntax error on line 348 after the latest pull (agentis-colonies with v1.3.1 changes):

```
tools/federation-dashboard.sh: line 348: syntax error near unexpected token `line'
```

The error occurs on both macOS bash 3.2.57 and Homebrew bash 5.3.9 — it is NOT a bash-version issue.

## Reproduction

```bash
cd agentis-colonies/dev-apprenticeship
./dashboard.sh 8420
```

Output:
```
Federation Dashboard: dev-apprenticeship
Discovered: 5 colonies, 21 agents

/Users/martinholy/git/factory/agentis-colonies/dev-apprenticeship/../tools/federation-dashboard.sh: line 348: syntax error near unexpected token `line'
```

Dashboard does not start. Port 8420 never opens.

## What's at line 348

Python code inside a `<<'PY'` heredoc (single-quoted = should be verbatim, no bash parsing):

```python
                for line in lines[-100:]:
                    line = line.strip()
```

Bash interprets `for line in ...` as a bash for-loop instead of passing it to Python as heredoc content.

## Diagnostic findings

1. **Heredoc boundaries look correct:**
   - Opens at line 124: `COLLECTOR_JSON="$(python3 - ... <<'PY' "${ALL_AGENTS[@]}"`
   - Closes at line 454: `PY` (alone on line, no trailing whitespace, confirmed via hexdump: `50 59 0a`)
   - `$()` properly closes on line 455: `)"` 
   - No stray `PY` delimiter between lines 125-453 (verified with `awk '/^PY/{print NR": "$0}'`)

2. **`bash -n` (syntax check) reports NO errors** — the parse-level syntax is valid according to bash 5's parser.

3. **The error occurs at RUNTIME**, not at parse time — yet it's reported as a "syntax error". This suggests the heredoc is correctly parsed but something at runtime causes bash to re-interpret the heredoc content as bash code (possibly via an `eval`, `source`, or a subshell execution path that reconstructs the command).

4. **File encoding:** UTF-8, no BOM, no Windows line endings.

5. **Both bash 3.2 and bash 5.3 produce the identical error on the identical line.** Neither is immune.

6. **Federation itself runs fine** — 21/21 healthy, 100% tick success. Only the dashboard is affected.

[...]

PR #246, opened 2026-04-22
fix(auto-promote): portable lock + extract config parser (#245)

### LINKED ISSUE #245: auto-promote.sh not portable on macOS (flock dependency + bash 3.2 incompatible shebang)
## Summary

`tools/auto-promote.sh` has two independent macOS-portability issues that together make the auto-promote sidecar a no-op on every macOS host. I hit both this morning while setting up the scheduler against the `dev-apprenticeship` federation on a MacBook (`agentis v1.4.4`).

## Environment

- macOS (Darwin 25.4.0, Apple silicon)
- Default `/bin/bash` 3.2.57 (Apple-shipped; GPL reasons)
- Fresh federation restart with `.auto-promote-install.toml` (`enabled = true`)

---

## Issue 1: `flock` is a hard dependency and doesn't exist on stock macOS

`tools/auto-promote.sh:72`:

```bash
exec 200>"$LOCK_FILE"
if ! flock -n 200; then
    echo "Another auto-promote instance is running. Exiting."
    exit 0
fi
```

On a stock macOS install `flock` is not present (it ships as part of `util-linux` on Linux, which Apple doesn't bundle). Every invocation fails with:

```
./tools/auto-promote.sh: line 72: flock: command not found
Another auto-promote instance is running. Exiting.
```

The exit code is 0, so the sidecar keeps calling the script every `interval_s` and keeps silently no-op'ing. From the operator's perspective the sidecar appears to run; in reality nothing ever promotes.

### Suggested fix

Graceful degradation: only take the lock if `flock` is available, otherwise fall back to a pid-file check (or a python `fcntl`-based helper, since the script already depends on python3). Example skeleton:

```bash
if command -v flock >/dev/null 2>&1; then
    exec 200>"$LOCK_FILE"
    if ! flock -n 200; then
        echo "Another auto-promote instance is running. Exiting."
        exit 0
    fi
else
    # pid-file guard for macOS / systems without util-linux
    if [ -f "$LOCK_FILE" ] && kill -0 "$(cat "$LOCK_FILE")" 2>/dev/null; then
        echo "Another auto-promote instance is running (pid $(cat "$LOCK_FILE")). Exiting."
        exit 0
    fi
    echo $$ > "$LOCK_FILE"
    trap 'rm -f "$LOCK_FILE"' EXIT
fi
```

Alternatively, document `brew install util-linux` in the setup docs and `PATH` instructions, plus have `install.sh` / `doctor` fail-fast on missing `flock` with a clear hint.

---

## Issue 2: Shebang `#!/bin/bash` targets macOS bash 3.2, which crashes on the eval heredoc around line 287

`tools/auto-promote.sh:1`: `#!/bin/bash`

On macOS this resolves to the Apple-shipped bash 3.2.57. The eval block starting at line 119:

```bash
eval "$(python3 - "$CONFIG_FILE" <<'PYCONFIG'
...
PYCONFIG
)"
```

crashes on bash 3.2 with:

```
./tools/auto-promote.sh: eval: line 287: syntax error near unexpected token `('
```

(Line numbers shift slightly with any added/removed lines above; on the current HEAD the token is the closing `)` of the subshell on what was line 284 pre-my-local-comments.)

The same script runs cleanly under bash 5.x (I verified with `/opt/homebrew/bin/bash tools/auto-promote.sh …` — 21 agents evaluated, proper `SKIP …` lines, journal write works).

### Suggested fix

Easiest: change the shebang to one that doesn't silently fall back to bash 3.2 on macOS. `#!/usr/bin/env bash` is the usual choice, but on macOS `env` can still resolve to `/bin/bash` depending on `PATH` ordering. Two options:

1. Rewrite the eval heredoc into a form bash 3.2 parses (move the python block into a separate `.py` helper file, or source a temp file).
2. Bump the minimum bash version in the script header comment and have `tools/colony-lint.sh` (or a new `check-bash-version.sh`) fail with a clear error when `$BASH_VERSION < 4.0`, pointing users to `brew install bash` + instructions to put `/opt/homebrew/bin` ahead of `/bin` in `PATH`.

Option 1 is cleaner (no new dependency). Option 2 is cheaper.

---

## Repro

On any macOS host with default `/bin/bash` and no `util-linux`:

```bash
cd agentis-colonies
./tools/auto-promote.sh dev-apprenticeship
# => "flock: command not found" + silent exit 0
```

Install `flock` via `brew install util-linux` and put it on `PATH` (or symlink it into `~/.local/bin`), re-run:

```bash
./tools/auto-promote.sh dev-apprenticeship
# => "eval: line 287: syntax error near unexpected token `('"
```

Only with both problems patched (e.g. `#!/opt/homebrew/bin/bash` + `flock` in `PATH`) does the script actually evaluate agents.

## Impact

- macOS operators (including contributors running the dev federation locally) have no working auto-promote unless they discover and patch both problems themselves.
- Silent failure mode is especially bad: the sidecar keeps firing every 30 min, the dashboard shows "21 skipped", and there is nothing in logs that points at `flock` being the root cause (the false-positive lock message is misleading).
- The dev-apprenticeship federation's entire learning loop depends on auto-promote; on macOS it effectively never runs.

## Workaround I'm using locally (not committed)

- Added `/opt/homebrew/opt/util-linux/bin/flock` symlinked into `~/.local/bin` (which is in my `PATH`).
- Patched shebang to `#!/opt/homebrew/bin/bash`.

Happy to send a small PR for the graceful-`flock`-fallback + a rewritten eval block if you'd like; flag here whether you'd prefer Option 1 or Option 2 above.

ylohnitram:
## Summary

- Replace `flock(1)` call (util-linux, not on stock macOS) with a `fcntl.flock` Python helper. Parent shell holds the fd open for the run, helper acquires the lock and exits.
- Extract the `eval "$(python3 - <<'PYCONFIG' ... PYCONFIG)"` block into `tools/auto-promote-config-parser.py`. Eliminates the macOS bash 3.2 parser bug that made the heredoc unusable. Same fix pattern as #170 / #172 on `federation-dashboard.sh`.
- Shebang to `#!/usr/bin/env bash` as a secondary guard.

Before this PR, every macOS invocation of `auto-promote.sh` silently no-op'd, logged a misleading "Another auto-promote instance is running" line, and the dev-apprenticeship learning loop effectively never ran on macOS hosts.

Plan lives in https://github.com/Replikanti/agentis-colonies/issues/245#issuecomment-4295823905.

## Test plan

- [x] `tools/test-auto-promote.sh` — 23 passed, 0 failed (was 19; added Test 10 "no heredocs", Test 11 "lock contention + release", reshaped Test 8 to invoke the real parser helper instead of a mirror)
- [x] `tools/colony-lint.sh` — 83 passed, 0 failed, 5 skipped
- [x] `tools/test-make-federation-bundle.sh` — 11 passed, 0 failed (confirms the two new helpers are in `BUNDLE.manifest` and included in the real federation tarball)
- [x] End-to-end Linux smoke test: `tools/auto-promote.sh dev-apprenticeship` with no daemons running → "Federation not running, no-op", exit 0
- [x] End-to-end Linux smoke test: config parse path verified by `bash -c 'eval "$(python3 auto-promote-config-parser.py auto-promote-config.yaml)"'` — every `CFG_*` var populated
- [ ] macOS E2E: needs verification on the reporter's MacBook (@ylohnitram). Both helpers use only POSIX `fcntl.flock` + stdlib, no Linux-only syscalls.

## Files

- **New:** `tools/auto-promote-lock.py` (~50 LoC)
- **New:** `tools/auto-promote-config-parser.py` (~170 LoC, verbatim lift of PYCONFIG body + main)
- **Edit:** `tools/auto-promote.sh` — shebang, lock block, config eval (net -165 LoC)
- **Edit:** `tools/test-auto-promote.sh` — 3 tests reshaped / added
- **Edit:** `dev-apprenticeship/BUNDLE.manifest` — add the 2 helpers
- **Edit:** `dev-apprenticeship/CHANGELOG.md` — Fixed entry under Unreleased
- **Edit:** `CLAUDE.md` — auto-promote.sh row: no-heredoc invariant + 2 helper rows

Closes #245


ylohnitram:
## QA report — additional macOS heredoc found and fixed (bdb826b)

Ran a full QA pass on the PR after CI went green. One finding, fixed in the latest commit before merge.

### Finding: third heredoc still present

`tools/auto-promote.sh` had **three** embedded-Python blocks, not two. The PR initially extracted two (`PYCONFIG`, `flock(1)` replacement) but left `<<'PYEVAL'` on the decisions block (previously line 168), which is the same class of macOS bash 3.2 parser bug — the entire command substitution silently no-ops on macOS.

**Why test 10 missed it:** the regex was `^[[:space:]]*<<`, which only catches heredoc markers at the start of a line. PYEVAL sat at the end of a backslash-continued line after a long argument list, so the pattern never saw it.

**Severity:** high — same failure class as the one this PR claims to fix. Shipping as-is would have left auto-promote broken on macOS (just at a later decision point) and probably spawned a #247 from the same reporter.

### Fix in bdb826b

- Extracted PYEVAL (287 LoC) to `tools/auto-promote-decisions.py`. Called with an explicit script path so the 11 positional args keep their `sys.argv[1..11]` indices.
- Strengthened test 10 regex to `<<['\"]?[A-Z_]+` on any non-comment line; would now catch the continued-line case.
- Added helper to `BUNDLE.manifest` + `CLAUDE.md` tools table.
- Updated CHANGELOG wording to reflect both heredocs extracted.

### Verification

- `tools/test-auto-promote.sh`: 23/23 pass (test 10 now validates zero heredocs under the stronger pattern).
- `tools/colony-lint.sh`: 83 passed, 0 failed, 5 skipped.
- `make-federation-bundle.sh`: all three helpers (`auto-promote-config-parser.py`, `auto-promote-lock.py`, `auto-promote-decisions.py`) present in the release tarball.
- `bash -n tools/auto-promote.sh`: syntax OK; no Python at column 0 after the PYEVAL body deletion.
- `grep -vE '^[[:space:]]*#' tools/auto-promote.sh | grep -cE "<<['\"]?[A-Z_]+"` → 0.

### Other checks (all clean)

- No `Avvoka` / `the_factory` references anywhere in the diff.
- No Czech or em-dashes in new content.
- `python3 -c "..."` inline forms (lines 41, 149, 195 etc.) are **not** heredocs and do not trigger the macOS bug — left as-is.
- `.ag` files untouched; ADR-0001 tier semantics unchanged.

PR is ready from my side.

PR #253, opened 2026-04-23
feat(dashboard): extract federation-dashboard as standalone component (#252)

### LINKED ISSUE #252: Extract federation-dashboard as separately-versioned component
## Why

`tools/federation-dashboard*` files are listed in `dev-apprenticeship/BUNDLE.manifest`, so every dashboard-only commit (#248 PR A/B/C, #170, #172, ...) forces a federation version bump. Semantically the federation hasn't changed. The dashboard is generic (`auto-discovers colonies/agents`) and is meant to serve every federation, but it has its lifecycle hard-bound to dev-apprenticeship.

Symptom from #248: three dashboard-only PRs would force a `dev-apprenticeship-v0.3.3` release for nothing else. The operator reads the federation changelog and sees entries that have nothing to do with their `.ag` agents.

## What should happen

Extract `federation-dashboard/` as a separately-versioned top-level component with its own:
- `VERSION`
- `CHANGELOG.md`
- `BUNDLE.manifest`
- `README.md` (operator-facing)
- `install.sh` (or just a tarball + extract)
- tag scheme `federation-dashboard-v<X.Y.Z>`
- release workflow (analogous to `.github/workflows/release.yml`, but triggered on `federation-dashboard-v*` tags)

The federation `install.sh` then either downloads a pinned dashboard version (`DASHBOARD_VERSION=0.X.Y`), or the operator installs the dashboard separately.

## Proposed layout (open to discussion)

```
federation-dashboard/
  VERSION
  CHANGELOG.md
  README.md
  BUNDLE.manifest
  install.sh                                 # extracts into the operator's ~/.local/share/federation-dashboard/
  bin/
    federation-dashboard                     # entry point (was tools/federation-dashboard.sh)
  lib/
    federation-dashboard-collector.py
    federation-dashboard-renderer.py
    federation-dashboard-history.py
    federation-dashboard-server.py
    federation-dashboard.html.template
```

## Shared dependencies — needs a decision

The dashboard calls these helpers, which are **not** dashboard-exclusive:
- `tools/auto-promote-decisions.py` — collector calls `--preview` (#248 PR A)
- `tools/resolve-tick-interval.py` — collector uses it for tick-interval lookup
- `tools/kill-federation.sh` — server invokes it from the `/kill` endpoint
- `tools/parse-toml.sh` — collector parses fed configs

Three options:

**(a)** Dashboard calls these helpers **from the federation directory** it's pointed at (`<fed-dir>/tools/...`). Assumes every federation ships them in its bundle (dev-apprenticeship already does). Clean separation, but duplication across every federation bundle.

**(b)** Dashboard **vendors** the helpers — copies live in `federation-dashboard/lib/`. Versioned with the dashboard, federation bundles no longer need to ship them. Risk: drift (two copies of `auto-promote-decisions.py` in the repo).

**(c)** Hoist into a top-level `lib/` shared layer. Both federation and dashboard depend on it. Architecturally cleanest, but a third release scheme (lib-v0.X.Y).

Recommendation: **(a) for v0.1.0, (c) as follow-up once there are more federations**. I.e. dashboard expects `<fed-dir>/tools/auto-promote-decisions.py` to exist; if it doesn't, gracefully degrades (skips the Promote Candidates panel instead of crashing).

## Migration

1. **Create `federation-dashboard/`**:
   - `git mv tools/federation-dashboard*.{sh,py,html.template} federation-dashboard/lib/` (except the entry point)
   - entry point → `federation-dashboard/bin/federation-dashboard`
   - write `federation-dashboard/VERSION` (`0.1.0`)
   - migrate dashboard-only changelog entries from `dev-apprenticeship/CHANGELOG.md` (#248 A/B/C, #170, #172, #143, #144, #160, #161, #163, #167, #177, #186 dashboard parts, ...) into `federation-dashboard/CHANGELOG.md` as a historical catalog → v0.1.0 = "first dashboard-as-component release; consolidates everything previously shipped under dev-apprenticeship 0.1.x–0.3.2"

2. **Update `dev-apprenticeship/`**:
   - Remove dashboard files from `BUNDLE.manifest`
   - `dashboard.sh` wrapper either (a) assumes the dashboard is installed at `$HOME/.local/share/federation-dashboard/bin/federation-dashboard`, (b) downloads it on first run, or (c) fails with an instruction to "install federation-dashboard separately"
   - `install.sh` §X adds a step: "Install federation-dashboard? [Y/n]" → curl `federation-dashboard-v<pinned>.tar.gz` from the release asset → extract into `~/.local/share/federation-dashboard/`

3. **Release infrastructure**:
   - `.github/workflows/release.yml` — extend the trigger or duplicate into a new `release-dashboard.yml` that fires on `federation-dashboard-v*` tags
   - `tools/make-federation-bundle.sh` — either generic enough already (probably not), or add `tools/make-dashboard-bundle.sh`
   - `colony-lint.sh` / `check-changelog.sh` — teach them about two changelog contexts (`dev-apprenticeship/` vs `federation-dashboard/`)

4. **Documentation**:
   - Top-level `README.md` — add a section about the dashboard as a separate component
   - `CLAUDE.md` — rewrite the `## Tools` section: drop dashboard files, add a `## federation-dashboard component` section
   - `doc/federation-dashboard.md` — stays, but path references update

5. **Tests**:
   - `tools/test-timeline-rendering.sh` — paths update to `federation-dashboard/lib/...`
   - Possibly move into `federation-dashboard/test/`
   - `tools/test-make-federation-bundle.sh` — assert that dashboard files are **not** in the federation bundle (regression guard)

## Open questions

- **Dashboard install location** — RESOLVED: XDG_DATA_HOME-aware. Install to `${XDG_DATA_HOME:-$HOME/.local/share}/federation-dashboard/` with a symlink to `${XDG_BIN_HOME:-$HOME/.local/bin}/federation-dashboard`. `install.sh` warns if `~/.local/bin` is not on PATH. State (index.html, history.json, sockets) is generated inside the federation directory it's pointed at, not under `$HOME`.
- **Federation install integration** — RESOLVED: opt-in prompt with default Yes in `dev-apprenticeship/install.sh`, downloads the pinned version from a release asset. Pin lives in `dev-apprenticeship/.dashboard-version`. `--no-dashboard` flag for operators who want to skip. `dashboard.sh` wrapper resolves: (1) `$FEDERATION_DASHBOARD_BIN` env, (2) XDG install path, (3) `command -v federation-dashboard`, (4) helpful error. Federation declares a soft minimum (`requires dashboard >= 0.1.0`); strict semver on the dashboard mitigates pin drift.
- **Dashboard version for existing dev-apprenticeship 0.3.2 installs** — RESOLVED: invasive migration is fine (single-operator project). `dev-apprenticeship 0.3.3` install.sh removes leftover `tools/federation-dashboard*` from prior installs without prompting. CHANGELOG entry calls out the extraction; no soft-prompt or opt-in.
- **First dashboard release version** — RESOLVED: `0.1.0`. Clean-slate semver under the new package identity. The previous dashboard code shipped inside `dev-apprenticeship` 0.1.x–0.3.2 but never had its own version contract, so there is nothing to continue from.
- **Release notes for federation-dashboard v0.1.0** — RESOLVED: minimal note. v0.1.0 entry is "First release as a standalone component. Code extracted from `dev-apprenticeship 0.3.2`. For history prior to extraction, see `git log -- tools/federation-dashboard*`." No reconstructed historical catalog — pre-extraction history stays in `dev-apprenticeship/CHANGELOG.md` where it actually happened.

## Acceptance criteria

- [ ] `federation-dashboard/VERSION` exists, contains `0.1.0`
- [ ] `federation-dashboard/CHANGELOG.md` in Keep-a-Changelog format
- [ ] `tools/federation-dashboard*` files removed / relocated
- [ ] `dev-apprenticeship/BUNDLE.manifest` no longer references dashboard files
- [ ] `dev-apprenticeship/dashboard.sh` works against a separately-installed dashboard
- [ ] `tools/test-make-federation-bundle.sh` fails when dashboard files leak into a federation bundle
- [ ] `.github/workflows/` triggers on `federation-dashboard-v*` tags and creates a release with the tarball
- [ ] `colony-lint.sh` / `check-changelog.sh` understand both changelog contexts
- [ ] `CLAUDE.md` updated
- [ ] `doc/federation-dashboard.md` paths updated
- [ ] First dashboard release `federation-dashboard-v0.1.0` is cut and tagged

## Dependency

This should land **before** the dev-apprenticeship 0.3.3 release, so 0.3.3 can pin a dashboard version instead of carrying a vendored copy.

ylohnitram:
Closes #252.

## Summary

- Dashboard moved to `federation-dashboard/` with its own `VERSION` (0.1.0), `CHANGELOG.md`, `README.md`, `BUNDLE.manifest`, and XDG-aware `install.sh`.
- New release flow: tag `federation-dashboard-v<X.Y.Z>` → `.github/workflows/release-dashboard.yml` → `tools/make-dashboard-bundle.sh` (portable sha256, excludes `__pycache__`/`*.pyc`).
- `dev-apprenticeship/install.sh` gains step 8 (opt-in install of the pinned dashboard; opt out via `FEDERATION_DASHBOARD_SKIP=1`). `dev-apprenticeship/.dashboard-version` pins the floor (currently `0.1.0`).
- `dev-apprenticeship/dashboard.sh` is now a resolver wrapper: `$FEDERATION_DASHBOARD_BIN` → XDG default → `command -v federation-dashboard`.
- Dashboard binary resolves federation-shared tools via `<fed-dir>/tools/` then `<fed-dir>/../tools/`, and **gracefully degrades** instead of hard-asserting: `/kill` → 503, Promote Candidates renders empty, tick interval falls back to 60000ms.
- `tools/check-changelog.sh` rewritten as loop over `COMPONENTS=("dev-apprenticeship/" "federation-dashboard/")`. Adding a new versioned component is one line.

## Files

| Before | After |
|---|---|
| `tools/federation-dashboard.sh` | `federation-dashboard/bin/federation-dashboard` |
| `tools/federation-dashboard-*.py` | `federation-dashboard/lib/federation-dashboard-*.py` |
| `tools/federation-dashboard.html.template` | `federation-dashboard/lib/federation-dashboard.html.template` |

## Cross-platform

The component needs to install cleanly on both Linux and macOS (stock bash 3.2, BSD coreutils):

- `federation-dashboard/install.sh` — no GNU-only sed flags (`sed -E 's/^# \?//'` replaced with explicit `show_help()` heredoc).
- `tools/make-dashboard-bundle.sh` — `sha256sum` → `shasum -a 256` fallback; `ln -sfn` (portable both ways).
- `federation-dashboard/bin/federation-dashboard` — zero heredocs (already enforced by `test-timeline-rendering.sh` test 19).

## Compatibility

`federation-dashboard >= 0.1.0` ↔ `dev-apprenticeship >= 0.3.3`. The standalone dashboard can serve any federation that satisfies its floor, including layouts where the federation and its shared tools live in different parent trees.

Net behaviour for an operator who runs `install.sh` and then `dashboard.sh` is unchanged — the difference is that dashboard fixes can now ship without forcing a federation re-release.

## Test plan

- [x] `tools/colony-lint.sh` — 83 passed, 0 failed, 6 skipped.
- [x] `tools/test-timeline-rendering.sh` — all 23 tests pass against new paths.
- [x] `tools/test-kill-endpoint.sh` — all 5 tests pass (fixture symlinks `kill-federation.sh` into `<fed-dir>/tools/`).
- [x] `tools/test-dashboard-fedpath.sh` — all 5 tests pass (federation-local, symlinked, cwd-fallback layouts).
- [x] `tools/test-make-federation-bundle.sh` — all 11 tests pass; dashboard files correctly blacklisted from `dev-apprenticeship/` bundles.
- [x] `tools/test-check-changelog.sh` — all 6 tests pass against the loop-driven rewrite.
- [x] `tools/make-dashboard-bundle.sh 0.1.0` produces a 49KB tarball; extract → `install.sh` → working symlink → dashboard boots against a real federation and renders 139KB of HTML.
- [ ] Post-merge: tag `federation-dashboard-v0.1.0`, verify `release-dashboard.yml` creates the GitHub release with bundle + CHANGELOG-extracted notes.
- [ ] Post-merge: bump `dev-apprenticeship` to 0.3.3 (CHANGELOG entry already in `[Unreleased]`).


ylohnitram:
## QA report — PR #253

End-to-end install/uninstall round-trip on Linux + boot against a real federation + graceful-degradation behaviour all verified. Bundle hygiene clean (no `__pycache__`, no `*.pyc`, no `BUNDLE.manifest`, no `dev-apprenticeship/` leakage, no contributor tooling). Resolver in `dev-apprenticeship/dashboard.sh` falls through correctly when `FEDERATION_DASHBOARD_BIN` is bad. `tools/check-changelog.sh` loop hard-fails on a `federation-dashboard/VERSION`-only PR. Colony-lint baseline: **84 passed, 0 failed, 6 skipped** (was 83 before — new test added below).

### Findings

| ID | Severity | Finding | Status |
|----|----------|---------|--------|
| F1 | Major | `federation-dashboard/BUNDLE.manifest:18` advertises a regression test (`tools/test-make-dashboard-bundle.sh`) that didn't exist; the dashboard tarball's contributor-only / pyc-free guarantees were enforced by hand-rolled bash patterns in `tools/make-dashboard-bundle.sh:118-127` with **zero coverage**. | ✅ FIXED in 4e68154 — added `tools/test-make-dashboard-bundle.sh` (12 tests covering version-mismatch rejection, runtime-file inclusion, exec-bit preservation, BUNDLE.manifest exclusion, zero `__pycache__`/`*.pyc`, no federation/contributor leakage, single top-level dir, end-to-end install + uninstall round-trip into a scratch XDG prefix). Auto-discovered by colony-lint. |
| F2 | Minor | `federation-dashboard/install.sh:11` header comment said `--prefix /opt/x` would fall back to `/opt/x/bin if no XDG_BIN_HOME`, but the implementation always uses XDG (lines 19, 51). The `--help` text on line 34 was already correct. | ✅ FIXED in 4e68154 — header comment now matches implementation (bin always XDG; `XDG_BIN_HOME` relocates the symlink). |
| F3 | Minor | `dev-apprenticeship/dashboard.sh:49` hardcoded `0.1.0` in the install-instructions error block as a fallback when `.dashboard-version` couldn't be read. Would silently rot if the federation bumped its pin. | ✅ FIXED in 4e68154 — fallback removed. If the pin file is unreadable, the error now points the operator at `dev-apprenticeship/install.sh` (which can recreate it) instead of printing a stale literal. |
| F4 | Minor | `federation-dashboard/bin/federation-dashboard:34-36` printed the no-args usage banner to **stdout**; "Federation directory not found" on line 41 correctly used stderr. Inconsistent within the same script. | ✅ FIXED in 4e68154 — usage banner now goes to stderr. |
| F5 | Minor | `federation-dashboard/bin/federation-dashboard:241` ran the Python server in the foreground, so the `trap '...' INT TERM` on line 232 could not fire until python had exited on its own. A service manager that delivers SIGTERM to the wrapper PID alone (not the pgid) could leave an orphan Python listening on the port. | ✅ FIXED in 4e68154 — server now backgrounded, wrapper `wait`s on it, trap kills both children explicitly. Smoke-tested: `kill -TERM <wrapper-pid>` on Linux takes the port down within 1s. |
| F6 | Minor | `federation-dashboard/install.sh` doesn't validate that the parent of `--prefix /nonexistent/deep/path` exists before `mkdir -p "$DATA_DIR"` — silently creates the whole tree on a typo. | ☐ UNFIXED — `mkdir -p` matches standard installer ergonomics; warning on every novel prefix would also cost the legitimate "first install on a clean machine" case. Operator can `--uninstall` to clean up if they realise they typo'd. Not worth a UX gate for the user's solo-operator workflow. |
| F7 | Minor | `tools/make-dashboard-bundle.sh:115` glob `"$COMP_DIR"/.[!.]*` would expand to its own literal text under `set -u` if no dotfiles existed (the `[ -e "$entry" ] || continue` guard handled it, but the intent was implicit). | ✅ FIXED in 4e68154 — `shopt -s nullglob dotglob` makes the dotfile-or-not behaviour explicit; collapsed the dotglob into the main `*` loop. |
| F8 | Nit | `federation-dashboard/CHANGELOG.md:16` references `git log -- tools/federation-dashboard*` for pre-extraction history without anchoring the working directory. | ☐ UNFIXED — `git log -- <pathspec>` works from anywhere inside a working tree, so the only ambiguity would be running it from outside the repo entirely (in which case `git` itself errors clearly). Not worth a doc tweak. |

### Verdict

**Mergeable.** All Blocker/Major findings fixed; 5 of 6 Minors fixed. Two intentionally skipped Nits (F6, F8) are documented above. Latest tip: `4e68154` — colony-lint 84/0/6, all dashboard test suites green, end-to-end SIGTERM teardown verified.

ylohnitram:
### QA report correction — F6 was actually a real bug

On re-evaluation, my original justification for leaving F6 UNFIXED was wrong: I claimed `--uninstall` could clean up a typo'd `--prefix`, but `--uninstall` only removes `$DATA_DIR/federation-dashboard` and the symlink — it doesn't touch the `mkdir -p`-created ancestor chain. A typo like `--prefix /opt/frederation-dashb` would silently materialise the whole typo'd path and leave the garbage on disk forever.

| ID | Status (updated) |
|----|------------------|
| F6 | ✅ FIXED in `d9ec95e` — `install.sh` now distinguishes an explicit `--prefix` from the XDG default; explicit prefixes whose parent doesn't exist are rejected with exit 3 + clear stderr message. XDG default path unchanged (those ancestors are creatable-in-place per spec). New regression test (`test-make-dashboard-bundle.sh` test 13) asserts exit !=0 AND that no directory is created anywhere under the typo'd path. Colony-lint 84/0/6; bundle tests 13/13. |
| F8 | ☐ Still UNFIXED — nit only. `git log -- <pathspec>` resolves pathspecs relative to the repo root from any cwd inside the working tree, so no cwd anchoring is actually needed. |

Tip now at `d9ec95e`.

Issue #288, opened 2026-04-24
federation-dashboard wrapper calls `agentis daemon list --json` without cd'ing to federation dir — renders every agent as "stopped" when run with wrong cwd


ylohnitram:
The `federation-dashboard` bash wrapper invokes `agentis daemon list --json` from whatever cwd it inherits at launch. `agentis` resolves `.agentis/` via cwd (no `--fed-dir` flag), so if the caller doesn't `cd` to the federation root first, the call returns `[]`. The collector then fabricates per-agent placeholders with `state="stopped"` and `health="unknown"`, and the dashboard renders every agent as dead even though the federation is fully alive.

### Reproduction

Run the dashboard from any cwd that isn't the federation root (or an ancestor that happens to contain a `.agentis/`):

```bash
systemd-run --user --unit=fed-dashboard \
    /home/<me>/.local/bin/federation-dashboard \
      /home/<me>/path/to/dev-apprenticeship 8420
# no --working-directory → systemd defaults cwd to $HOME
```

Inspect the generated HTML:

```bash
curl -s http://localhost:8420/ | python3 -c '
import sys, re, json
html = sys.stdin.read()
m = re.search(r"\"agents\"\s*:\s*(\[.+?\])\s*,\s*\"experience", html, re.DOTALL)
agents = json.loads(m.group(1))
from collections import Counter
print({k: dict(Counter(a.get(k) for a in agents)) for k in ("state","health")})
'
# → {'state': {'running': 21}, 'health': {'healthy': 21}}   (expected)
# → {'state': {'stopped': 21}, 'health': {'unknown': 21}}   (what you see)
```

Meanwhile from a real shell in the federation root:

```bash
cd /path/to/agentis-colonies
agentis daemon list --json | python3 -c 'import sys,json;print(len(json.load(sys.stdin)))'
# → 21  ← agentis finds .agentis/ relative to cwd
```

### Evidence of the wrapper's assumption

`federation-dashboard/bin/federation-dashboard` around line 127:

```bash
local DAEMONS
DAEMONS="$(agentis daemon list --json 2>/dev/null || echo '[]')"
```

No `cd` before the call, no `--fed-dir` arg to agentis, no env var. The wrapper knows `FED_DIR` from argv[1] but never uses it for agentis lookups — only for passing into the collector. Same pattern repeats wherever the wrapper re-invokes agentis during regen.

### Impact

[...]

Issue #286, opened 2026-04-24
dashboard /start endpoint kills the federation it just launched


ylohnitram:
The dashboard has a **"Start Federation"** button that POSTs to `/start`. This endpoint shells out to `start-federation.sh` via `subprocess.run(..., timeout=60)` (see `federation-dashboard/lib/federation-dashboard-server.py` around line 637):

```python
result = subprocess.run(
    ['bash', start_script],
    capture_output=True, text=True,
    cwd=fed_dir, timeout=60,
)
```

But `dev-apprenticeship/start-federation.sh` never exits under normal operation — after spawning the 21 daemons and the optional auto-promote sidecar, it sits on a `while :; do sleep "$AP_INTERVAL"; …` loop (start-federation.sh:137-138) that keeps the sidecar alive for the lifetime of the federation (default 30 min intervals).

So every `/start` click:

1. Python spawns `bash start-federation.sh` as a subprocess child.
2. `start-federation.sh` launches 21 `agentis daemon …` children + the sidecar loop.
3. Python waits on the subprocess → hits the 60s timeout → `subprocess.run` sends **SIGTERM** to its bash child.
4. The bash child dies → kernel propagates the signal down the process group → **all 21 agents that just finished spawning die too**, plus the sidecar.
5. Python's `/start` handler returns an error response. From the operator's perspective the button "stopped" the agents it was supposed to start.

**Reproduction**

1. Stop the federation (or start from a clean state: `agentis daemon prune`, empty `.agentis/daemon/`).
2. Open the dashboard, click **Start Federation**.
3. Watch `agentis daemon list` — 21 agents spawn, then disappear within ~60s.
4. `journalctl --user -u fed-dashboard.service` shows no crash of the dashboard itself — just the 60s `subprocess.SubprocessError: TimeoutExpired` path.

**Root-cause pairing**

This is the `/start` counterpart to #274 (auto-promote false DEGRADED): the sidecar is designed to be long-lived, and every control path that wraps `start-federation.sh` in a bounded call or process group will kill the federation it meant to manage.

**Fix options**

1. **Detach the subprocess entirely** (preferred). Use `subprocess.Popen` with `start_new_session=True`, do not `wait()`/`communicate()`, return 202-Accepted with a "federation spawn kicked off, poll `agentis daemon list` for state". Dashboard doesn't own the federation's lifetime anymore:
   ```python
   subprocess.Popen(
       ['bash', start_script],
       cwd=fed_dir,
[...]

Issue #279, opened 2026-04-24
github-api.sh: normalize_* funcs pass body via env var, exceeds MAX_ARG_STRLEN on closed-PR history


ylohnitram:
After the #277 workaround lets `forge-api.sh` actually dispatch to `github-api.sh`, the implementation colony still fails every tick with:

```
[code_writer] GitLab poll failed: ... ExecFailed: exec sh failed (exit 126):
  /.../dev-apprenticeship/implementation/scripts/github-api.sh: line 140:
  /usr/bin/python3: Argument list too long
```

**Root cause**

`github-api.sh` normalization helpers push the full HTTP body into an environment variable and then fork `python3`:

```bash
# normalize_pulls (line 137-140), same pattern in normalize_issue, normalize_mr_changes, etc.
normalize_pulls() {
    local DATA
    DATA="$(cat)"
    DATA="$DATA" python3 <<'PY'
import os, json
data = json.loads(os.environ["DATA"])
...
PY
}
```

On Linux, `execve(2)` imposes a per-string limit of `MAX_ARG_STRLEN = PAGE_SIZE * 32 = 131072` bytes on any single element of `argv` / `envp` (see `include/uapi/linux/binfmts.h`). The total `ARG_MAX` (2 MB on x86_64) is much larger, but no single env var can exceed 128 KB. Passing `DATA` of any size ≥ 128 KB triggers `E2BIG` → `execve` fails before python3 even starts, manifesting as the kernel's terse `Argument list too long` string at exec-time.

**Reproduction**

```bash
$ token=$(grep -oP '(?<=token = ")[^\"]+' dev-apprenticeship/triage/config/colony.toml)
$ curl -s -H "Authorization: Bearer $token" \
    "https://api.github.com/repos/Replikanti/agentis-colonies/pulls?state=closed&per_page=20&sort=updated" \
    | wc -c
436671
```

Any mature repo's closed-PR history will blow through 128 KB at the default `per_page=20` (`merge-requests` arm, line 433). Smaller repos avoid it but are not safe; anything with 30+ rich-bodied PRs lands in this failure mode on the first real tick.

**Why the existing `--per-page` flag doesn't save us**

`merge-requests` accepts `--per-page`, but `code_writer.ag` (and most .ag callers) don't pass it — they expect the forge wrapper to return data in a single call. Forcing the agents to page and lowering the default fixes nothing at the root.

**Fix**

Switch every normalize_* helper from env-var transport to stdin:

```diff
[...]

PR #294, opened 2026-04-24
fix(dashboard): route large payloads via temp-files to bypass MAX_ARG_STRLEN (#293)

### LINKED ISSUE #293: federation-dashboard wrapper passes history.json / collector JSON as argv — Linux MAX_ARG_STRLEN (128KB) exceeded after ~8h of tick accumulation
`federation-dashboard/bin/federation-dashboard:195` invokes the renderer with the full serialized `$COLLECTOR_JSON`, `$HISTORY`, and `$REMEDIATION` payloads as **positional argv arguments** to `python3`. Linux's per-argv-string kernel limit is `MAX_ARG_STRLEN = PAGE_SIZE * 32 = 131072 bytes` — same class of bug #279 fixed for `github-api.sh`, different file.

`$HISTORY` grows monotonically (one snapshot per sidecar tick, pruned only after 7 days per federation-dashboard's own comment). After ~8h of runtime on this federation it reached **131 698 bytes = 128.6 KB**, one byte over the kernel limit — dashboard startup began failing with:

```
/home/<me>/.local/bin/federation-dashboard: line 195: /usr/bin/python3: Argument list too long
```

systemd service `fed-dashboard.service` went through the same pattern: repeated spawn attempts, each one hitting `execve` E2BIG before the python process could start. No stderr from python; the shell's own error is all we get.

### Reproduction

Let a dashboard run against a federation with `auto-promote.enabled=true` (sidecar writes to `.dashboard/history.json` every tick). After ~8-24h of uptime, `wc -c .dashboard/history.json` climbs past 131072. Next dashboard restart (or initial regen after reboot) fails with the above error. Shorter path: `head -c 140000 /dev/urandom | base64 > .dashboard/history.json` (forces-fail on the very next regen).

### Why `$COLLECTOR_JSON` can also trip it

On a 21-agent federation the collector output is ~30-50 KB today; on a 40-agent federation with deep recent-experience tails it will cross 128 KB inside a week. Even if `$HISTORY` is capped, a second bomb is waiting.

### Fix options

1. **Temp-file transport** (cleanest). The bash wrapper writes each large payload to a `mktemp` file; `renderer.py` takes paths on argv and reads them:
   ```bash
   HIST_FILE=$(mktemp --suffix=.json)
   REMED_FILE=$(mktemp --suffix=.json)
   COLL_FILE=$(mktemp --suffix=.json)
   trap 'rm -f "$HIST_FILE" "$REMED_FILE" "$COLL_FILE"' EXIT
   printf '%s' "$HISTORY"        > "$HIST_FILE"
   printf '%s' "$REMEDIATION"    > "$REMED_FILE"
   printf '%s' "$COLLECTOR_JSON" > "$COLL_FILE"
   python3 "$LIB_DIR/federation-dashboard-renderer.py" \
       "$LIB_DIR/federation-dashboard.html.template" \
       "$HTML_FILE" \
       "$FED_NAME" "$FED_NAME_JS" "$COLONY_COUNT" "$AGENT_COUNT" \
       "$EPOCH" "$TIMESTAMP" \
       "@$COLL_FILE" "@$HIST_FILE" "@$REMED_FILE" \
       "$COLONY_LIST_JS"
   ```
   `renderer.py` checks each arg for the `@` prefix and `json.load`s the file. Small values like `$FED_NAME` still pass inline. No new dependencies.

2. **Stdin multiplex**. Feed one framed JSON (`{"collector":..., "history":..., "remediation":...}`) over stdin; `renderer.py` reads stdin once. Slightly more invasive to the renderer signature but cleaner on the bash side.

3. **Env var** (rejected). Same `MAX_ARG_STRLEN` per-string limit applies to env vars; only trades one hit for another.

Option 1 is the minimum viable fix. The collector invocation at `bin/federation-dashboard:127-140` (whatever that shape is now post-#252) should get the same treatment preemptively.

### Workaround (operators hit by this today)

```bash
# truncate history to last N snapshots — keeps trend visible, gets under 128KB
python3 -c '
import json
p = ".dashboard/history.json"
d = json.load(open(p))
json.dump(d[-200:], open(p, "w"))
' \
  && ./dashboard.sh    # or systemctl --user restart fed-dashboard.service
```

Restart the dashboard; regen will succeed on the now-smaller file. history.json will grow back to the limit over ~8h, so this is a stopgap only.

### Impact

P1 for any long-running dashboard deployment. The limit bites silently (no warning as it approaches 128 KB), the failure mode is cryptic (bash's own `Argument list too long` line, no python traceback), and the diagnostic rabbit hole looks identical to whatever SIGTERM / crash-loop theory the operator is currently chasing. My session spent a lot of time on the wrong hypothesis before finally `wc -c .dashboard/history.json` made it obvious.

### Out of scope (neutral phrasing)

- Separate from #279 (github-api.sh normalizers fixed in #281/#282). Same class, different code path.
- The dashboard `/start` detach (#286 → #289) and wrapper-cwd (#288) bugs are orthogonal.
- No change to `.ag` files, no VERSION bump, no install.sh involvement.

ylohnitram:
Fixes #293.

## Summary

Dashboard wrapper passed `$COLLECTOR_JSON`, `$HISTORY`, `$REMEDIATION`, and `$DAEMONS` as positional argv to `python3`. After ~8 h of history accumulation, the JSON blobs cross Linux's `MAX_ARG_STRLEN` (128 KB per single argv string) and the dashboard startup fails with `Argument list too long` before any helper opens its template. Same class of failure as the one addressed in #279 (also a 128 KB inline-argv overrun), but in a different code path (`federation-dashboard/bin/federation-dashboard` -> `lib/*.py`).

## Approach

Spool large payloads to `mktemp` files and pass `@<path>` argv tokens to the Python helpers. Renderer (argv[9..11]) and collector (argv[1]) detect the prefix and read the file via a tiny `_read()` helper; non-`@` values pass through unchanged so any external caller is unaffected.

Cleanup uses an in-function `RETURN` trap rather than an `EXIT` trap, because `generate()` is also called from the regen-only path and from the 60 s background loop in normal mode — an `EXIT` trap on the wrapper would leak files for the lifetime of the dashboard process.

## Files changed

| File | Net delta |
|------|-----------|
| `federation-dashboard/bin/federation-dashboard` | +33 / -2 |
| `federation-dashboard/lib/federation-dashboard-renderer.py` | +21 / -3 |
| `federation-dashboard/lib/federation-dashboard-collector.py` | +12 / -1 |
| `federation-dashboard/CHANGELOG.md` | +10 / 0 |
| `tools/test-dashboard-argv-overflow.sh` | +135 / 0 (new) |

## Test plan

- [x] `bash -n federation-dashboard/bin/federation-dashboard` clean
- [x] `python3 -c 'import ast; ast.parse(...)'` clean for renderer + collector
- [x] New regression test `tools/test-dashboard-argv-overflow.sh` stages a 240 KB synthetic `history.json` with a unique sentinel, runs `DASHBOARD_REGEN_ONLY=1 federation-dashboard $FED_DIR`, asserts exit 0 and that the rendered `index.html` contains the sentinel — proves the temp-file plumbing reaches `{{HISTORY}}` substitution. Skips gracefully when the dashboard binary is not on PATH.
- [x] `tools/colony-lint.sh` passes 106 / 0 / 1 (baseline 105 + new test).
- [x] `test-timeline-rendering.sh` test 19 (zero heredocs) still green.

## Related

Same class of MAX_ARG_STRLEN overflow as the issue tracked in #279 (different file). #286 and #288 were the prior Fixed entries in this CHANGELOG section.

### DIFF
--- a/federation-dashboard/CHANGELOG.md
+++ b/federation-dashboard/CHANGELOG.md
@@ -9,6 +9,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Dashboard wrapper no longer crashes with `Argument list too long` after
+  several hours of accumulated history. `$COLLECTOR_JSON`, `$HISTORY`,
+  `$REMEDIATION`, and `$DAEMONS` now travel to the Python helpers via
+  temp files (`@<path>` argv prefix) instead of inline argv strings, so
+  they do not hit Linux's 128 KB `MAX_ARG_STRLEN` per-argv cap. Renderer
+  and collector accept both forms (`@`-prefix reads the file, plain
+  string passes through unchanged) — same class of failure as
+  [#279](https://github.com/Replikanti/agentis-colonies/issues/279) but
+  in a different code path
+  ([#293](https://github.com/Replikanti/agentis-colonies/issues/293)).
 - `/start` endpoint no longer SIGTERMs `start-federation.sh` (and its 21
   just-spawned agents) when the sidecar loop runs past 60s. Handler now
   detaches the subprocess with `start_new_session=True` and returns 202
--- a/federation-dashboard/bin/federation-dashboard
+++ b/federation-dashboard/bin/federation-dashboard
@@ -156,12 +156,35 @@ generate() {
         LOG_DIR=".agentis/logs"
     fi
 
+    # --- Temp-file argv routing (#293) ---
+    # Linux MAX_ARG_STRLEN caps any single argv string at 128 KB. After ~8 h
+    # of accumulation, $COLLECTOR_JSON / $HISTORY / $REMEDIATION / $DAEMONS
+    # cross that threshold and python3 fails with "Argument list too long".
+    # Spool large payloads to temp files and pass `@<path>` argv tokens; the
+    # Python helpers detect the prefix and read the file (non-`@` values
+    # still pass through inline for backward compat). Same class as #279
+    # (github-api.sh) but a different code path.
+    local DAEMON_FILE COLL_FILE HIST_FILE REMED_FILE
+    # Plain `mktemp` (no GNU-only --suffix flag) for macOS bash 3.2 portability;
+    # consumers locate the file by its argv slot, not by extension.
+    DAEMON_FILE="$(mktemp)"
+    COLL_FILE="$(mktemp)"
+    HIST_FILE="$(mktemp)"
+    REMED_FILE="$(mktemp)"
+    # In-function cleanup (RETURN trap) — generate() is also called from the
+    # regen-only path and from the 60 s background loop, so an EXIT trap on
+    # the wrapper would leak files for the duration of the long-running
+    # process. RETURN fires on every return path (success, set -e, return).
+    trap 'rm -f "$DAEMON_FILE" "$COLL_FILE" "$HIST_FILE" "$REMED_FILE"' RETURN
+
+    printf '%s' "$DAEMONS" > "$DAEMON_FILE"
+
     # --- Comprehensive per-agent data collection ---
     # Single python3 invocation collects all enriched data: experience stats,
     # .ag descriptions, log lines, PID liveness, event timeline, and
     # confidence change history. Outputs a single JSON blob consumed by JS.
     local COLLECTOR_JSON
-    COLLECTOR_JSON="$(python3 "$LIB_DIR/federation-dashboard-collector.py" "$DAEMONS" "$AGENT_COLONY_MAP" "$FED_DIR" "$EPOCH" "$EXPERIENCE_DIR" "$LOG_DIR" "$DASH_DIR" "$COLONY_LIST_PY" "$FED_TOOLS_DIR" "${ALL_AGENTS[@]}")"
+    COLLECTOR_JSON="$(python3 "$LIB_DIR/federation-dashboard-collector.py" "@$DAEMON_FILE" "$AGENT_COLONY_MAP" "$FED_DIR" "$EPOCH" "$EXPERIENCE_DIR" "$LOG_DIR" "$DASH_DIR" "$COLONY_LIST_PY" "$FED_TOOLS_DIR" "${ALL_AGENTS[@]}")"
     if ! echo "$COLLECTOR_JSON" | python3 -c 'import sys,json;json.loads(sys.stdin.read())' 2>/dev/null; then
         COLLECTOR_JSON='{"agents":[],"experience_counts":{"total":0},"events":[],"confidence_changes":[]}'
     fi
@@ -186,6 +209,10 @@ generate() {
     done
     COLONY_LIST_JS="[${COLONY_LIST_JS%,}]"
 
+    printf '%s' "$COLLECTOR_JSON" > "$COLL_FILE"
+    printf '%s' "$HISTORY" > "$HIST_FILE"
+    printf '%s' "$REMEDIATION" > "$REMED_FILE"
+
     # --- HTML generation ---
     # All HTML/CSS/JS lives in federation-dashboard.html.template. The renderer
     # substitutes 10 named sentinels and writes index.html atomically. This
@@ -201,9 +228,9 @@ generate() {
         "$AGENT_COUNT" \
         "$EPOCH" \
         "$TIMESTAMP" \
-        "$COLLECTOR_JSON" \
-        "$HISTORY" \
-        "$REMEDIATION" \
+        "@$COLL_FILE" \
+        "@$HIST_FILE" \
+        "@$REMED_FILE" \
         "$COLONY_LIST_JS"
 }
 
--- a/federation-dashboard/lib/federation-dashboard-collector.py
+++ b/federation-dashboard/lib/federation-dashboard-collector.py
@@ -12,7 +12,11 @@
 # and re-parsing the heredoc body as bash code at runtime.
 #
 # Args (positional):
-#   1: daemons_json        — JSON array from `agentis daemon list --json`
+#   1: daemons_json        — JSON array from `agentis daemon list --json`,
+#                            or "@<path>" to read from file (#293 — bypass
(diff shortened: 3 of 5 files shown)


ylohnitram:
## QA Report — PR #294

**Verdict:** ✅ **ship-it**

Every one of the 10 verification checkpoints from the QA directive passes. Temp-file plumbing plus `@<path>` argv prefix correctly bypasses `MAX_ARG_STRLEN` on all four large payloads (`$DAEMONS`, `$COLLECTOR_JSON`, `$HISTORY`, `$REMEDIATION`); both helpers preserve full backward compat for callers that still pass plain inline JSON; `RETURN`-trap cleanup leaks zero temp files across repeated regens; colony-lint baseline is 106/0/1 as advertised; scope is minimal and intentional; CI is green; live smoke against the real 50 KB-history federation exits clean.

| # | Area | Result | Notes |
|---|------|--------|-------|
| 1 | Closing-issue graph | ✅ | `closingIssuesReferences == [293]` exactly (`gh pr view 294 --json closingIssuesReferences -q '.closingIssuesReferences[].number'` → `293`). `#279 / #286 / #288` in the body use neutral phrasing (“tracked in”, “prior Fixed entries”), no spillover. |
| 2 | Temp-file routing: `$DAEMONS` → collector argv[1] | ✅ | `bin/federation-dashboard` line 180 `printf '%s' "$DAEMONS" > "$DAEMON_FILE"`; line 187 passes `"@$DAEMON_FILE"` as argv[1] to the collector. |
| 3 | Temp-file routing: `$COLLECTOR_JSON`/`$HISTORY`/`$REMEDIATION` → renderer argv[9..11] | ✅ | `bin/federation-dashboard` lines 212-214 spool, lines 222-234 pass `@$COLL_FILE` / `@$HIST_FILE` / `@$REMED_FILE` as argv[9..11]. |
| 4 | Cleanup trap | ✅ | Line 178 `trap 'rm -f "$DAEMON_FILE" "$COLL_FILE" "$HIST_FILE" "$REMED_FILE"' RETURN` is an in-function trap; correctly chosen over EXIT because `generate()` runs in the 60 s background loop (EXIT would leak for the lifetime of the process). Verified: regen-before `ls /tmp/tmp.*` = 13, regen-after = 13 → no leak. |
| 5 | Renderer backward compat | ✅ | `_read()` applied only to `sys.argv[9..11]`; `sys.argv[3..8,12]` are passed through untouched (renderer.py lines 61-71). Probed with plain-string argv: output correctly contains `COLL={"agents":[]} HIST=[] REM=[] FED=fedname`. |
| 6 | Collector backward compat | ✅ | `_read()` applied only to `sys.argv[1]` (collector.py line 40); argv[2..N] pass through unchanged. Probed with plain-string `'[]'` at argv[1]: rc=0, `agents=1`, JSON output valid. |
| 7 | Test harness `tools/test-dashboard-argv-overflow.sh` | ✅ | Executable (`755`), generates 243536-byte synthetic `history.json` (>200 KB, comfortably past the 128 KB MAX_ARG_STRLEN cap), drives `DASHBOARD_REGEN_ONLY=1`, asserts rc=0 + sentinel in index.html, skips gracefully when dashboard binary is missing. Timestamps anchored to `now` so the 7-day prune does not silently truncate the fixture. Standalone run: 2 passed / 0 failed / 0 skipped. |
| 8 | Colony-lint baseline | ✅ | `tools/colony-lint.sh` → **106 passed, 0 failed, 1 skipped** (baseline 105 + new `test-dashboard-argv-overflow.sh`). Matches target. |
| 9 | Scope check | ✅ | `git diff main...colonies-293-dashboard-argv-overflow --stat` lists exactly 5 files: `federation-dashboard/bin/federation-dashboard` (+33/-2), `.../lib/federation-dashboard-renderer.py` (+21/-3), `.../lib/federation-dashboard-collector.py` (+12/-1), `federation-dashboard/CHANGELOG.md` (+10/0), `tools/test-dashboard-argv-overflow.sh` (+151 new). No `dev-apprenticeship/`, `.ag`, `install.sh`, `VERSION`, or `BUNDLE.manifest` touch. |
| 10 | PR-body language safety | ✅ | `grep -iE '(fix(es|ed)?\|close[sd]?\|resolve[sd]?)\s+#[0-9]+'` returns exactly one match: `Fixes #293.` Other issue references (`#279`, `#286`, `#288`) use neutral phrasing. |
| 11 | CI | ✅ | `colony-lint / SUCCESS` on latest commit (1m29s). |
| 12 | Live smoke | ✅ | `DASHBOARD_REGEN_ONLY=1 bin/federation-dashboard dev-apprenticeship` against the real 54556-byte `.dashboard/history.json` → rc=0, "Discovered: 5 colonies, 21 agents". New code path handles the current federation state cleanly. Port 9999 not bound (regen-only mode exits before spawning the server). |
| 13 | Additional syntax checks | ✅ | `bash -n federation-dashboard/bin/federation-dashboard` clean; `python3 -c 'ast.parse(...)'` clean for both renderer + collector. |

### Observations (non-blocking)

- **`RETURN` trap choice is correct and documented.** Lines 174-177 of `bin/federation-dashboard` explain exactly why an `EXIT` trap would leak for the life of the wrapper (the background regen loop calls `generate()` every 60 s). Good engineering.
- **`mktemp` without `--suffix` is deliberate.** Line 168 comments explicitly flag macOS bash 3.2 portability — GNU-only `--suffix` is avoided. Consumers locate files by argv slot, not extension. Matches the dashboard's existing portability posture.
- **Collector docstring updated in lockstep** (argv[1] definition now mentions `@<path>` and cites #293). Renderer module docstring similarly updated with the new `@<path>` contract and cross-ref.
- **CHANGELOG Unreleased entry is precise** (bytes, root cause, fix shape, back-references #279 / #286 / #288). Ready for the next dashboard release-PR.
- **Sanity assert inside the test** (`assert len(data) > 200*1024`) makes the regression test non-silent: if a future refactor lets the fixture shrink below the pre-fix threshold, the test fails loud rather than masking the bug.

### Blockers

None. Safe to merge.

Issue #359, opened 2026-04-26
dashboard: intent-driven tab restructure (Status / Cost / Recovery / Progress / Config) + 4 bug fixes


ylohnitram:
## Bug + redesign — fundamental rework

The v0.6.0 + #357 redesign cut down the long-scroll problem but the **tab categorization is conceptually wrong**. Operator quote (translated):

> "You've got the categories badly designed. You need to think about it more in product terms, not as logical grouping of individual features. Look — I log in and what do I want to see? Is it running? What all is running, such that it means it's running? How much is it eating? Won't it eat me alive in an hour? Or a day? Or a month?! How much will I spend? When it dies, how fast can I get it back to operational? How is it progressing (in anything)? How do I configure it? How do I configure as much at once as possible? Stop the goddamn flicker every minute!"

This is a **product-driven** taxonomy, not a feature-driven one. Each tab maps to one operator question, not one feature group.

## Proposed tab structure (intent-driven)

Replace the current 6-tab cut (Overview / Agents / Promotions / Learning / Cost / Config) with a 5-tab cut keyed to operator questions:

| Tab | Question | Content |
|---|---|---|
| **Status** | *"Běží to? Co všechno běží?"* — Is it running? What all is running so that we can say it's running? | Top: federation health verdict (✅ Running / ⚠️ Degraded / ❌ Down). 21 agents pulse-grid (color = state, click → drill-in). 2 sidecars status pills + last-tick age. Last-error timestamp + count. **No charts, no bars, no tables — pure single-glance verdict.** |
| **Cost** | *"Kolik to žere? Nesežere mě to za hodinu / den / měsíc? Kolik utratím?"* | Per-minute burn rate ($/min, tokens/min). Projected spend at current rate: 1h, 1d, 1w, 1m (operator wants to see "if I leave it running for a month, am I OK?"). Today / 7d / 30d cumulative. Spending vs caps (cost-cap progress bar — if installed). Per-colony cost split. Token volume separated from $. |
| **Recovery** | *"Když to chcípne, jak rychle ho dám do provozuschopnýho stavu?"* | Bulk-restart all stopped. Per-agent restart. Per-sidecar restart. Federation kill switch (with confirm). Per-agent log tail (last 20 lines). Last 10 watchdog kills (forensic — when did things die). **All actionable.** |
[...]

PR #367, opened 2026-04-27
fix(dashboard): sentinel-injection-safe single-pass renderer (#366)

### LINKED ISSUE #366: dashboard: renderer sentinel-injection blanks the dashboard when an agent mentions {{HISTORY}} in experience.out
## Bug — sentinel injection in `federation-dashboard-renderer.py` blanks the dashboard

**Severity: P0** (live dashboard outage). Already mitigated locally; this issue tracks the permanent fix.

## Symptom

Operator opens `http://localhost:8420/`, sees the tab bar but **all tabs are empty**, and the SSE status dot is not bright green. Hard refresh does not help. `/snapshot.json` is valid; `/events` SSE channel is alive; `/dashboard.json` is 404 (correct — there is no such endpoint).

## Root cause

`federation-dashboard-renderer.py:77-78` substitutes the 10 named template sentinels in a Python `for sentinel, value in substitutions.items(): html = html.replace(sentinel, value)` loop. **The substitution is sequential, not single-pass**, so the value of an earlier sentinel can contain the literal text of a later sentinel and get re-substituted.

In the live incident:

1. The `risk_assessor` agent processed issue #362, whose body referenced the template sentinels by name (it discusses `{{HISTORY}}` and `{{COLLECTOR_JSON}}` and other sentinels in the "test-suite rewrite" section).
2. The agent wrote a `risk_assessment` experience entry whose `out` field contained the literal text `(\`{{HISTORY}}\` rendering, Phase Readiness assertions from test 22)`.
3. The collector built `collector_json` (passed to renderer as `{{COLLECTOR_JSON}}`); the JSON serialization preserved the `{{HISTORY}}` substring inside the `timeline[].payload.out` string.
4. The renderer substituted `{{COLLECTOR_JSON}}` first → 3.5 MB blob written into the template, including the literal text `{{HISTORY}}` somewhere in it.
5. The next iteration substituted `{{HISTORY}}` → 1.5 MB history JSON injected into the *collector data string literal* in the template, breaking `const data = {…}` so the JSON parser in the browser threw on `Unexpected token`. JS dispatch died, every renderer skipped, tabs were left empty, SSE health dot stuck on init color.

## Reproduction (deterministic)

Add a row to any agent's experience JSONL where the `out` field contains the literal substring `{{HISTORY}}`. Wait for the next dashboard refresh. The browser's `Console` will show:

```
SyntaxError: Unexpected token '['
    at <anonymous>:1578
```

…and every tab is blank, including the Status pulse-grid (the green-dot complaint).

## Fix proposal

Replace the sequential `str.replace` loop in `federation-dashboard-renderer.py` with a single-pass `re.sub` over the joined sentinel pattern, where the replacement function looks up the matched sentinel in the substitutions dict and returns the value as-is. A single pass cannot re-substitute injected sentinels.

```python
import re
pattern = re.compile('|'.join(re.escape(s) for s in substitutions))
html = pattern.sub(lambda m: substitutions[m.group(0)], html)
```

This is functionally identical when no value contains a sentinel literal, and **safe** when one does: the replacement function returns the value verbatim and the regex engine's scan position advances past the value, so any sentinel literal inside the value is skipped.

## Tests

Add a new `tools/test-timeline-rendering.sh` test:

> **renderer is sentinel-injection-safe: when COLLECTOR_JSON contains literal `{{HISTORY}}`, the rendered output keeps the literal as-is and does NOT inject the history payload.**

Implementation: pipe a synthetic substitution where `collector_json` value is `{"timeline":[{"out":"{{HISTORY}}"}]}` and `history_json` is `[{"sentinel":true}]`; assert the output contains `{{HISTORY}}` exactly once and does NOT contain `"sentinel":true`.

## Severity & scope

- **P0** because it's already happened to a live operator and is fully reproducible from any future agent that mentions sentinel names in its `out` field. Risk_assessor and pretty much any planning-tier agent will hit this once it processes a dashboard-related issue.
- One-file change in `federation-dashboard/lib/federation-dashboard-renderer.py`.
- One new test in `tools/test-timeline-rendering.sh` (~30 LOC).
- No version bump in this PR — PATCH bump (federation-dashboard 0.6.1) deferred to next release PR.

## Mitigation already applied locally

The offending `risk_assessor` JSONL line was rewritten in-place to neutralize the sentinel literal (`{{HISTORY}}` → `{{ HISTORY }}` with internal spaces). The dashboard recovered after one refresh cycle. The renderer fix prevents recurrence.

## Out of scope

- Escaping in the agent's experience writer. Agents can write any `out` text; the renderer is the safe layer.
- Reworking the sentinel approach (e.g., switching to `{{ "FED_NAME" | substitute }}` or a templating library). Single-pass `re.sub` is sufficient and minimal.
- Deletion of the broken JSONL entry. The mitigation already in place keeps the entry's content (preserves the audit trail) by neutralizing the sentinel.

ylohnitram:
## Summary
- Single-pass `re.sub` over all 10 template sentinels — replaces the previous sequential `html.replace(sentinel, value)` loop that allowed an agent's experience `out` field containing `{{HISTORY}}` to be re-substituted when injected via `{{COLLECTOR_JSON}}`.
- Adds regression test 59 in `tools/test-timeline-rendering.sh` (rendered output must contain `{{HISTORY}}` exactly once and `[{"sentinel":true}]` exactly once when COLLECTOR_JSON value contains `{{HISTORY}}`).

## Why P0
Live operator dashboard hit this 30 minutes ago — tabs visible but empty, SSE status dot stuck on init color. `risk_assessor` agent had processed a dashboard-related issue whose body literally mentioned `{{HISTORY}}`; its experience JSONL row carried the literal forward to the renderer; renderer's iterative substitution turned the collector data string into corrupted JSON; browser threw on `const data = ...;` parse; entire JS dispatch died.

Fully reproducible: any future agent that mentions a sentinel name in `out` reproduces the outage.

## Mitigation already applied
The offending JSONL line was rewritten in-place to neutralize the sentinel (`{{HISTORY}}` → `{{ HISTORY }}`). The dashboard recovered after one refresh cycle. This PR prevents recurrence.

## Files
- `federation-dashboard/lib/federation-dashboard-renderer.py` — `+8 -2` LOC: import `re`, replace sequential loop with `re.compile + pattern.sub(lambda)`. Behaviour-identical when no value contains a sentinel literal; safe when one does.
- `tools/test-timeline-rendering.sh` — `+38 -0` LOC: new test 59.

## Out of scope
- Escaping in the agent's experience writer (agents can write arbitrary text; the renderer is the safe layer).
- Reworking the sentinel approach (single-pass `re.sub` is sufficient).
- federation-dashboard PATCH bump → deferred to next release PR.

## Test plan
- [x] `bash tools/test-timeline-rendering.sh` — `Results: 59 passed, 0 failed`.
- [x] CI colony-lint must pass.
- [ ] Live operator confirms dashboard renders normally after refresh (already confirmed pre-PR).

Closes #366.

### DIFF
--- a/federation-dashboard/lib/federation-dashboard-renderer.py
+++ b/federation-dashboard/lib/federation-dashboard-renderer.py
@@ -35,6 +35,7 @@ to temp files; values without the `@` prefix still pass through unchanged
 for backward compatibility.
 """
 import os
+import re
 import sys
 
 
@@ -74,8 +75,16 @@ def main():
     with open(template_path, 'r', encoding='utf-8') as f:
         html = f.read()
 
-    for sentinel, value in substitutions.items():
-        html = html.replace(sentinel, value)
+    # Single-pass substitution (#366): a value (notably {{COLLECTOR_JSON}}) can
+    # contain the literal text of another sentinel — e.g. an agent's experience
+    # row whose `out` field discusses "{{HISTORY}}" while reasoning about a
+    # template change. A sequential `for s in subs: html.replace(s, v)` would
+    # then re-substitute that literal in the next iteration and inject the
+    # history payload into the collector data string, breaking
+    # `const data = ...;` parsing in the browser. re.sub() advances past the
+    # replacement value, so any sentinel literal inside a value is left intact.
+    pattern = re.compile('|'.join(re.escape(s) for s in substitutions))
+    html = pattern.sub(lambda m: substitutions[m.group(0)], html)
 
     tmp = output_path + '.tmp.' + str(os.getpid())
     with open(tmp, 'w', encoding='utf-8') as f:
--- a/tools/test-timeline-rendering.sh
+++ b/tools/test-timeline-rendering.sh
@@ -2332,6 +2332,51 @@ else
     fail "58: bulk-apply checkbox group wiring regressed (#359)"
 fi
 
+# --- #366 test 59: renderer is sentinel-injection-safe. When COLLECTOR_JSON
+#     contains the literal text of another sentinel ({{HISTORY}}), the
+#     rendered output keeps the literal as-is and does NOT inject the
+#     history payload. Sequential str.replace would re-substitute and
+#     break `const data = ...;` parsing in the browser, blanking the
+#     dashboard. The fix uses a single-pass re.sub. ---
+T59_DIR="$(mktemp -d)"
+T59_TEMPLATE="$T59_DIR/template.html"
+T59_OUTPUT="$T59_DIR/output.html"
+cat > "$T59_TEMPLATE" <<'TPL'
+<!doctype html><html><body>
+const data = {{COLLECTOR_JSON}};
+let history = {{HISTORY}};
+let remediation = {{REMEDIATION}};
+</body></html>
+TPL
+# Note the {{HISTORY}} literal embedded inside the COLLECTOR_JSON value.
+T59_COLLECTOR='{"timeline":[{"out":"agent reasoned about {{HISTORY}} sentinel"}]}'
+T59_HISTORY='[{"sentinel":true}]'
+T59_REMEDIATION='[]'
+T59_OK=0
+if python3 "$RENDERER_PY" \
+        "$T59_TEMPLATE" \
+        "$T59_OUTPUT" \
+        "fed" '"fed"' "1" "1" "0" "ts" \
+        "$T59_COLLECTOR" "$T59_HISTORY" "$T59_REMEDIATION" "[]" \
+        2>/dev/null
+then
+    # The {{HISTORY}} literal must survive inside the COLLECTOR_JSON value
+    # (one occurrence). The history payload [{"sentinel":true}] must appear
+    # exactly once — at the {{HISTORY}} substitution site, NOT also injected
+    # into the COLLECTOR_JSON spot.
+    T59_HIST_LIT_COUNT="$(grep -cF '{{HISTORY}}' "$T59_OUTPUT" 2>/dev/null || echo 0)"
+    T59_PAYLOAD_COUNT="$(grep -cF '[{"sentinel":true}]' "$T59_OUTPUT" 2>/dev/null || echo 0)"
+    if [ "$T59_HIST_LIT_COUNT" = "1" ] && [ "$T59_PAYLOAD_COUNT" = "1" ]; then
+        T59_OK=1
+    fi
+fi
+if [ "$T59_OK" -eq 1 ]; then
+    pass "59: renderer is sentinel-injection-safe — {{HISTORY}} inside COLLECTOR_JSON is not re-substituted (#366)"
+else
+    fail "59: renderer re-substituted a sentinel literal inside another sentinel's value (#366)"
+fi
+rm -rf "$T59_DIR"
+
 echo ""
 echo "Results: $PASS passed, $FAIL failed"
 [ "$FAIL" -eq 0 ]


ylohnitram:
## QA report (#367)

| Check | Result | Notes |
|---|---|---|
| Fix correctness (re.sub semantics) | ✅ | `re.sub` advances scan position past the matched **input** range, not the replacement output, so any sentinel literal embedded in a replacement value is never re-scanned. Verified with `pattern.sub(..., '{{A}} and {{C}}')` where `{{A}}` → `'B {{C}} D'`: output `'B {{C}} D and X'` — the `{{C}}` injected via `{{A}}` is left intact while the standalone `{{C}}` is correctly substituted. |
| All 59 tests pass | ✅ | `bash tools/test-timeline-rendering.sh` → `Results: 59 passed, 0 failed`. Includes test 16 (renderer ast-parses), 17/18 (sentinel substitution mechanics), 22 (Phase Readiness still rendered), plus all 5-tab restructure tests from #359. |
| Sentinel-prefix collision (FED_NAME vs FED_NAME_JS) | ✅ | Even though Python's `re` uses leftmost-**first** alternation (not leftmost-longest), the current 10-sentinel set is naturally collision-safe: every sentinel ends with the fixed `}}` terminator, so `{{FED_NAME}}` cannot match a prefix of `{{FED_NAME_JS}}` — the `}}` boundary is required. Programmatically verified that no sentinel is a substring of another. Empirical test: `{'{{FED_NAME}}': 'NAME', '{{FED_NAME_JS}}': 'NAME_JS'}` over `'A={{FED_NAME}} B={{FED_NAME_JS}}'` → `'A=NAME B=NAME_JS'` (correct). |
| Regression test 59 fails on old code | ✅ | Reproduced the old sequential loop on test 59's synthetic input: `{{HISTORY}}` literal count = **0** (re-substituted), payload count = **2** (injected into both `{{HISTORY}}` slot AND inside `{{COLLECTOR_JSON}}` value). Test 59 asserts both must equal 1 → both assertions would fail under the old code. Test 59 is correctly catching the failure mode. |
| No unrelated diff | ✅ | `git diff main..HEAD --stat` → only two files changed: `federation-dashboard/lib/federation-dashboard-renderer.py` (+11/-2) and `tools/test-timeline-rendering.sh` (+45/-0). Total +56/-2 LOC. Inline comment in renderer thoroughly explains the rationale. |
| Doc updates | ✅ | CLAUDE.md describes the renderer as "substitutes 10 named sentinels" — that contract is unchanged. The new safety guarantee is additive. CHANGELOG / VERSION bump deferred to next release PR per project convention; PR body acknowledges this in `Out of scope`. |

**Verdict: ✅ ship-it**

Additional notes:
- **Byte-identical safe-path behaviour confirmed.** When no replacement value contains a sentinel literal (the overwhelmingly common case), the new `re.sub` produces output byte-identical to the old sequential loop. Length matches, content matches.
- **Fragility note (not blocking).** `re.compile('|'.join(re.escape(s) for s in {}))` would raise `re.error: empty pattern` on an empty substitutions dict. Today the dict is statically populated with 10 fixed entries so this cannot occur, but worth flagging for any future refactor that makes population conditional. A defensive guard (`if substitutions:` / fall back to identity) would cost one line.
- **Dict iteration order.** The fix relies on Python 3.7+ insertion-order-preserving dicts to keep the alternation deterministic. Standard for the runtime targeted by the rest of this codebase.
- **Mitigation already applied to the live federation** (offending JSONL row neutralized to `{{ HISTORY }}`); this PR prevents recurrence for any future agent that mentions a sentinel name in `out`.

CI: colony-lint **SUCCESS**. Mergeable: **YES**. P0 hotfix is ready to ship.

Issue #378, opened 2026-04-28
Dashboard renders 'sidecar not installed' when sidecar is actively running (split-brain detection)


ylohnitram:
## Observed

When `<fed>/.auto-promote-install.toml` is missing but the sidecar process is in fact alive (recent ticks in `<fed>/.agentis/logs/auto-promote.log`, fresh `<fed>/.agentis/logs/auto-promote.sidecar_started_at`), the dashboard's per-sidecar pill renders `not installed` and the status banner falls back to "no sidecar" semantics.

This happens because `start-federation.sh` reads the install file once at boot to decide whether to spawn the sidecar; if the file is later removed (operator `rm`, accidental cleanup, restored federation tree, etc.), the running process keeps ticking but the dashboard's `installed` probe — `os.path.isfile(install_file)` in `federation-dashboard/lib/federation-dashboard-collector.py:560` — flips to `false`. `_sidecar_status()` (same file, line ~1521) then short-circuits to `'not-installed'` regardless of the log liveness fields it already collected.

Concrete repro:

1. Bring up `dev-apprenticeship` with auto-promote enabled.
2. Wait for one tick (auto-promote.log grows; sidecar.started_at file present).
3. `rm <fed>/.auto-promote-install.toml`.
4. Open the dashboard → per-sidecar pill says `not installed`, age column says `not installed`, restart button is disabled — even though the underlying loop is still ticking on its 30-min cadence.

The same gap exists symmetrically for `cost-cap`: `.cost-cap.toml` + `cost-cap.log` + `cost-cap.sidecar_started_at`.

## Why it matters

The current status text is misleading. The dashboard claims the sidecar is not installed, but the federation is actively being scheduled by it. Two symptoms operators have hit:

- After `kill-federation.sh`-then-`start-federation.sh` cycles where the install file vanished between runs, the operator can't tell from the dashboard whether the sidecar is running. They have to `pgrep` to find out.
- The "restart" button is disabled, so the recovery path the dashboard offers doesn't apply (there's nothing to restart — the sidecar is fine; the install file is what's missing).

## Proposal

Make the collector detect a `running-without-install` state and the template render an actionable label.

Detection rule (collector, both sidecars):

- `install_file_present = os.path.isfile(install_file)` (current value of `installed`)
[...]

PR #379, opened 2026-04-28
feat(dashboard): detect orphaned sidecars (running, install file missing) (#378)

ylohnitram:
## Summary

- A sidecar process can survive after its install file (`.auto-promote-install.toml` / `.cost-cap.toml`) is deleted — `start-federation.sh`'s loop keeps appending to the log. The dashboard previously labelled these `not installed`, misleading operators into thinking the sidecar is dead.
- Detection cross-references log mtime against install-file presence. Threshold is 4× the production default interval (7200s) — matches the `down` cutoff already used by `_sidecar_status()` so orphan and healthy installs share the same liveness ceiling.
- New collector field `data.sidecars[].running_orphan` (boolean) and new enum value `data.sidecars[].status === 'orphan'`. Pill colour is orange (same palette as `degraded`); restart button stays disabled with a tooltip pointing operators at `install.sh §7`. Federation health banner stays HEALTHY (the loop is scheduling) — adds a detail-line hint instead. Existing `2× interval silence` DEGRADED predicate is unchanged.
- VERSION 0.6.0 -> 0.6.1; CHANGELOG entry under `[Unreleased] / Added`.

Addresses #378.

## Test plan

- [x] `bash tools/test-dashboard-sidecar-grace.sh` — 3/3 PASS (regression check, the existing `2× interval` grace path is untouched).
- [x] `bash tools/test-dashboard-sidecar-orphan.sh` — 5/5 PASS (new test, fixtures A-E):
  - A auto-promote orphan -> `('orphan', true)`
  - B cost-cap orphan -> `sidecars[1] = ('orphan', true)`
  - C stale orphan (log >7200s) -> `('not-installed', false)` (pins 4×1800s threshold)
  - D installed + ticking -> `('healthy', false)`
  - E installed but never-ticked -> `('silent', false)`
- [x] `bash tools/colony-lint.sh` — 141 passed, 0 failed, 3 skipped (the new test is auto-discovered).
- [x] `GITHUB_BASE_REF=main bash tools/check-changelog.sh` — `CHANGELOG consistent with federation-dashboard/ diff`.


ylohnitram:
## QA Report — #378 / PR #379

**Verdict:** ✅ ship-it

### Test results

| Check | Result | Notes |
|-------|--------|-------|
| `tools/test-dashboard-sidecar-orphan.sh` | ✅ 5/5 | Fixtures A-E all green; pins 4×1800s threshold via fixture C |
| `tools/test-dashboard-sidecar-grace.sh` | ✅ 3/3 | Regression — `2× interval` grace path untouched |
| `tools/colony-lint.sh` | ✅ 141/0/3 | New test auto-discovered; 3 skips (`test-kill-*` destructive, kept off CI) |
| `GITHUB_BASE_REF=main bash tools/check-changelog.sh` | ✅ consistent | `CHANGELOG consistent with federation-dashboard/ diff` |
| `tools/test-make-dashboard-bundle.sh` | ✅ 13/13 | Bundle script integrity unaffected by VERSION bump 0.6.0 → 0.6.1 |
| Live federation smoke test | ✅ orphan detected | With `.auto-promote-install.toml` temporarily renamed: collector emits `sidecar.running_orphan=true`, `sidecars[0].status='orphan'`, `sidecars[0].running_orphan=true`. Install file restored; collector stderr clean. |
| Diff sanity (no abs paths, no debug, branch ordering, exec bit) | ✅ | No `/home/...` paths in committed files; no leftover `print()`/`console.log()`/`pdb` (the `print()` matches in the new test are intentional Python JSON helpers); orphan branch precedes `not-installed` in `_sidecar_status()` (lines 1548-1551); orphan branch precedes `!sidecar.installed` in `renderFedHealthBanner()` (lines 2065-2070); `.sidecar-pill.orphan` CSS at line 862, reachable via `renderSidecarStatus()` line 2446 (`class=\"sidecar-pill ${status}\"`); `tools/test-dashboard-sidecar-orphan.sh` is `chmod +x`. |
| Edge cases (time-skew, missing `last_tick_ts`, cost-cap banner) | ✅ | Time-skew (future tick): `(now - future) < 7200` → `True`, orphan fires — safe outcome (recent file touch implies live loop). Missing `last_tick_ts`: explicit `is not None` guard at lines 625 and 1549 → `running_orphan` stays `False`. Cost-cap banner.json/state.json paths (lines 1457-1492) only mutate `status`/`reasons`/`metrics` etc., never touch `installed`, so an orphan cost-cap with stale banner files keeps `installed=False`. |
| PR body (Addresses #378, no auto-close) | ✅ | "Addresses #378" present; zero hits for `fix/close/resolve #378` (case-insensitive). |

### Findings

| # | Severity | Finding | Result |
| 1 | info | Both `data.sidecar` (singular auto-promote dict, used by `renderFedHealthBanner()`) and `data.sidecars[]` (array, used by `renderSidecarStatus()` / `renderStatusVerdict()`) carry `running_orphan` symmetrically. CSS pill class `.sidecar-pill.orphan` matches the `degraded` orange palette as planned. | ✅ |
| 2 | info | Time-skew handling is implicit-but-correct: a future `last_tick_ts` produces a negative diff that satisfies `< ORPHAN_INFER_S`, so orphan fires. Defensible behaviour — a fresh log mtime in any direction means the loop is touching the file. | ✅ |
| 3 | minor | `renderFedHealthBanner()` only checks `sidecar.running_orphan` (auto-promote singular). It does not check `cost_cap.running_orphan`, so a cost-cap orphan with healthy auto-promote will not appear in the banner detail line. The per-sidecar pill in the Status tab (`renderSidecarStatus()`) does cover this case correctly via `data.sidecars[1]`, so the operator still sees it — just not in the federation banner. Plan's Step 2e explicitly scoped the banner change to the auto-promote `sidecar.*` namespace, so this is consistent with the agreed scope, but a follow-up to teach the banner about cost-cap orphans would close the symmetry. | ⚠️ |
| 4 | info | Threshold is pinned via fixture C (log mtime at `NOW - (ORPHAN_INFER_S + 100)` resolves to `not-installed` not `orphan`), so any future tuning of `ORPHAN_INFER_S` shows up as a test failure rather than silent semantic drift — exactly as the plan called for. | ✅ |
| 5 | info | VERSION bump 0.6.0 → 0.6.1 (PATCH) is defensible per CLAUDE.md: a new optional dict key + new enum value normally trip MINOR, but `data.sidecars[]` is internal-component-only (not consumed outside `federation-dashboard.html.template`) and this is functionally a defensive UX fix. CHANGELOG entry under `[Unreleased] / Added` is concise and complete. | ✅ |

### Recommendation

Ship it. All required checks pass, the live smoke test on `dev-apprenticeship/` confirmed end-to-end orphan detection (collector emitted `status='orphan'` + `running_orphan=true` for `sidecars[0]` with the install file temporarily renamed), and the install file was restored cleanly with no residual side effects. The single ⚠️ minor finding (banner does not surface cost-cap orphans, only auto-promote) is consistent with the plan's scope and has a working fallback in the Status tab's per-sidecar pill — operators will still see the cost-cap orphan, just not reiterated in the banner. Worth a quick follow-up issue if symmetric banner coverage is desired, but not a merge blocker.

PR #487, opened 2026-05-09
feat(tribes-bench): seed initial tribe pool > DEATH_THRESHOLD (finite-pool follow-up)

ylohnitram:
## Summary
- #485 finite-pool R&D substrate landed zero-sum reward + carrying-capacity drain, but the cold-start economy (pool starts at 0) makes Goldilocks tuning impossible: any reasonable DEATH_THRESHOLD fires before exponential reward income from replicates can build up.
- Add `STAGE3_TRIBE_INITIAL_POOL` knob so tribes start with capital. Defaults to 0 = OFF (legacy cold-start, byte-identical to v1.7.4 behaviour).
- Plumbed: env var → bootstrap printf → 5 × `start-colony.sh` memo seed `tribe-<name>:pool` from `$INITIAL_POOL`.

## Empirical motivation

| Smoke | METABOLIC | DEATH | INITIAL_POOL | Spawns | Mortality |
|-------|-----------|-------|--------------|--------|-----------|
| #15 | 5 | (no cascade) | 0 | 18 | 0% |
| #16 | 50 | 2000 | 0 | 5 | 50% |
| #17 | 15 | 2000 | 0 | 4 | 56% |
| #18 | 15 | 500 | 0 | 18 | 0% |

DEATH=2000 (#16, #17) kills tribes before exponential phase lands (cold-start drain hits DEATH within first ~20 ticks). DEATH=500 (#18) never triggers because metabolic drain doesn't reach it — but mortality stays at 0%. There is no static DEATH/METABOLIC pair that produces both ≥10 spawns AND ≥30% mortality from a zero pool.

With `INITIAL_POOL=5000 + DEATH=2000 + METABOLIC=15`:
- pool drops below DEATH at ~tick 40 (5000 − 75/tick × 40 = 2000) without rewards
- replicates fed by `reward=200/replicate` stretch the lifeline → burst phase before death pressure
- POOL_CAP=10000 caps runaway growth (zero-sum competition still active)

## Acceptance

- Default INITIAL_POOL=0 byte-identical to v1.7.4 behaviour. Verified by lint pass + zero-knob smoke runs upstream.
- Smoke #19 (next, with INITIAL_POOL=5000 / DEATH=2000 / METABOLIC=15) will validate that both ≥10 spawns AND ≥30% mortality emerge in the same run.

## Out of scope

- agentis-core changes (this is colonies-only; the runtime contract is unchanged).
- Adaptive INITIAL_POOL per tribe rep / population.

## Test plan
[...]

Issue #485, opened 2026-05-09
Stage 3 finite shared resource pool — zero-sum tribe economy for niche/hierarchy emergence (R&D)


ylohnitram:
## Severity

P2 — R&D emergence research substrate. Sister-issue to #470 (CB calibration) but addresses a different layer: the *reward economy*, not the *cognitive budget*.

## Background

Smoke #14 (post-v1.7.4 multi-gen replication) plateau'd at ~5 spawns over 30 min with 30% mortality. Static `DEATH_THRESHOLD = 2000` produces neither exponential growth nor 50%+ mortality — pool drains too fast for replicas to accumulate budget for second-gen replication, but not fast enough to crash whole tribes.

Static threshold is the "Option 1" arm of the Malthusian-tension trilemma:

| Policy | Defense fit | R&D emergence fit |
|---|---|---|
| Option 1: dynamic threshold (scales with population) | ✅ predictable equilibrium, procurement-friendly | ❌ boring, suppresses transients |
| Option 2: **finite shared pool** (zero-sum) | ❌ adversary-exploitable, no survival guarantee | ✅ classic substrate for niche / hierarchy / cooperation emergence |
| Option 3: boom-bust (low threshold, long pilot) | ❌ chaotic, unacceptable for tactical | ✅ r/k-strategy phenotype divergence |

This issue scopes Option 2 — the R&D emergence path. Option 1 is filed separately as a defense-product shape.

## Concept

Today: every TP gets unconditional `REWARD_FULL=200` (or `REWARD_SUBSEQUENT=20` for non-first-finders) credited to the hunter's CB pool via `bug_ledger.jsonl`.

Proposed: each tribe has a **capped shared resource pool** (`tribe-<name>:resource_pool` memo, initial e.g. 10000) that depletes on every TP credit. Replenishes at a slow fixed rate via the Stage 3 rotation timer (e.g. +50 per rotation interval).

Effect:
- More hunters in a tribe = faster pool depletion = smaller per-TP reward
- Hunters compete for slice of fixed budget
- Some hunters consistently win (high TP rate, position in code, prompt-variant fit) → accumulate CB → replicate
- Losers drain own CB → die via existing pool-drain path
- **Equilibrium = carrying capacity** defined by replenishment rate ÷ per-tick drain
- Population finds steady-state at the carrying capacity, NOT at static `DEATH_THRESHOLD`

[...]

Issue #459, opened 2026-05-06
Stage 4: Emergence research program (5-9 month roadmap, awaiting commit)


ylohnitram:
## Status

**This is a roadmap stub, not an active milestone.** No work begins until external interest or funding aligns. #439 (Stage 3) is infrastructure-complete with apparatus validated — what tonight's session demonstrated. Stage 4 would scale the apparatus from "infrastructure validated" to "real emergence research program" with the population sizes, time scales, and search-space richness that the literature shows is actually required.

Filing this issue creates a public record of what serious work would look like, so an external reader (researcher, advisor, funder) can see the realistic scope before deciding whether to engage.

## Goal

Test the hypothesis that an LLM agent ecosystem under selection pressure produces specialists that no human authored — **strong emergence** in the OpenAI-hide-and-seek / DeepMind-AlphaStar / Lenat-EURISKO sense, not weak emergence in the Schelling-segregation sense.

## What Stage 3 actually demonstrated, and why it isn't enough

Stage 3 (#439) ran 5 specialty tribes across 2 docker-isolated nodes for 30 minutes against a 2-crate planted-bug palette. Result: 4 of 5 tribes saturated their bug class (5/5 hits), one tribe struggled, 23 verified findings total. That replicates the Stage 2 N=3 single-node result on a multi-node infrastructure. **Methodologically, it confirms structural specialization beats generalist brute force**, with σ ≤ 1.0 on TPs across runs.

What Stage 3 did **not** demonstrate:

- Replication — `replicate(self_node_addr())` returned empty 22 of 22 attempts, federation handshake never consummated between containers
- Death — no tribe starved, rewards exceeded costs throughout
- Mutation — scaffolded but never invoked because no replicate succeeded
- Novel specialists — every finding came from the 5 hardcoded specialty prompts

The apparatus is **bijective with the search space**: 5 tribes, 5 bugs, 5 specialty classes. Even if every prereq fired perfectly, there is no slack for novel specialists to inhabit. The setup was never going to produce strong emergence regardless of runtime fixes.

## What the literature says is actually required

| Constraint | Stage 3 had | Real-emergence research needs |
|---|---|---|
| Population size | 5 tribes | 100-1000+ |
| Wall clock | 30 min smoke / 6h target | Days to weeks, or massively parallel |
| Action / strategy space | 15 hardcoded variants | Open-ended (LLM-driven prompt mutation) |
[...]

PR #22, opened 2026-04-10
docs: remove em dashes from root README and tools scripts

ylohnitram:
## Summary

Follow-up from #19 QA scope gap. Removes the remaining em dashes in the repo root so the writing-style rule is uniformly enforced.

- `README.md`: 4 rewrites in overview prose
- `tools/colony-lint.sh`: 1 rewrite in the header comment
- `tools/new-colony.sh`: 5 rewrites in the header comment and the "Next steps" scaffolding output

After this merges, `grep -rn '—' .` returns zero matches repo-wide.

Closes #20

## Test plan

- [x] `grep -rn '—' .` returns zero matches
- [x] `./tools/colony-lint.sh` passes (25 passed, 0 failed, 5 skipped)
- [x] `bash -n` passes for both modified scripts (shellcheck runs in CI)
- [x] Each rewrite preserves original meaning

PR #489, opened 2026-05-09
feat(tribes-bench): per-hunter age mortality (workaround for shared-state RMW race)

ylohnitram:
## Summary
- Add per-hunter age counter keyed on PPID (no shared-state race). Hunter suicides via \`agentis daemon stop\` when age > \`hunter:max_age\`.
- New env var \`STAGE3_HUNTER_MAX_AGE\` (default 0 = OFF). Tribe-wide cascade still applies as fallback.
- 5 × \`hunter.ag\` + 5 × \`start-colony.sh\` + 1 × \`run-stage3-docker.sh\`. 190 insertions.

## Why

Smokes #15-#21 exposed shared-state RMW race in the tribe-pool drain pattern: 5 hunters concurrent read-modify-write on \`tribe-X:pool\` lose updates. Cap drain at ~1 hunter/min regardless of population. Smoke #21 (post-v1.7.5 colony plumb) confirmed colony plumb landed (pool drained 22× faster) but reproduced the race shape — pool dropped 5000→1052 in 10min then tribe-wide cascade killed everyone in one wave.

## Why per-hunter, not atomic memo_decrement

Considered shipping a runtime \`memo_increment\` builtin in agentis-core, but rejected: a convenience primitive without authority/quorum model would silently break under Byzantine adversary (defense use case — single compromised agent drains shared counter undetected). Per-hunter age keeps the runtime contract clean (each hunter writes its OWN unique key, no concurrency primitive needed), preserves the defense positioning, and unblocks the colonies-side R&D goal.

## Implementation

\`hunter.ag\` (5 tribes, identical block inserted after existing metabolic block):
- Read \`hunter:max_age\` memo (default 0 = OFF, byte-identical to v1.7.5)
- Get \`\$PPID\` via \`exec sh\` (stable per daemon, unique across tribe-mates + replicas thanks to v1.7.5 colony plumb)
- Read+increment+write \`hunter:\$PPID:age\` (no race — only this hunter writes this key)
- If age > max_age: \`learn(\"self_death\")\` + lookup own agent_id from \`.pid\` registry + \`agentis daemon stop \$own_id --async\`

\`start-colony.sh\` (5 tribes): seed \`hunter:max_age\` from \`\$HUNTER_MAX_AGE\` env, default 0.

\`run-stage3-docker.sh\`: \`STAGE3_HUNTER_MAX_AGE\` env var + plumb into bootstrap printf as \`HUNTER_MAX_AGE=%s\`.

## Acceptance

- Default OFF (HUNTER_MAX_AGE unset) byte-identical to v1.7.5.
- Smoke #22 with \`HUNTER_MAX_AGE=20\` + \`INITIAL_POOL=20000\` (pool starvation pushed far out so per-hunter is the dominant mortality) demonstrates **both** ≥10 spawns AND ≥30% mortality in the same run, with replication continuing through hunter deaths (not the smoke #21 mass-extinction shape).

## Test plan
- [x] \`colony-lint.sh\` passes (168 passed, 0 failed)
- [ ] Stage3 docker smoke #22 with HUNTER_MAX_AGE=20 / INITIAL_POOL=20000

Refs: #485 (finite-pool), #487 (initial pool seed), Replikanti/agentis-core v1.7.5 (M106 colony plumb), #459 (Stage 4 R&D).

### DIFF
--- a/tribes-bench/tools/run-stage3-docker.sh
+++ b/tribes-bench/tools/run-stage3-docker.sh
@@ -56,6 +56,13 @@
 #                              and any positive METABOLIC_COST drives
 #                              the tribe immediately into death-pressure
 #                              regime before the burst phase can land).
+#   STAGE3_HUNTER_MAX_AGE      #487 follow-up: per-hunter age cap. Each
+#                              hunter increments its own age counter
+#                              (keyed on PPID, no shared-state RMW race)
+#                              and suicides via `agentis daemon stop`
+#                              when age > MAX_AGE. Default 0 = OFF (no
+#                              per-hunter age mortality; tribe-wide
+#                              cascade still applies as before).
 #   STAGE3_LLM_BACKEND         llm.backend value injected into both
 #                              hermetic configs. Default: openai (#445).
 #   STAGE3_OPENAI_MODEL        Model id when STAGE3_LLM_BACKEND=openai.
@@ -165,6 +172,12 @@ DEATH_THRESHOLD="${STAGE3_DEATH_THRESHOLD:-300}"
 TRIBE_POOL_CAP="${STAGE3_TRIBE_POOL_CAP:-0}"
 TRIBE_METABOLIC_COST="${STAGE3_TRIBE_METABOLIC_COST:-0}"
 TRIBE_INITIAL_POOL="${STAGE3_TRIBE_INITIAL_POOL:-0}"
+# #487 follow-up: per-hunter age mortality. Each hunter has its own
+# monotonic age counter keyed on PPID; suicides via `agentis daemon stop`
+# when age > HUNTER_MAX_AGE. Eliminates the shared-state RMW race the
+# tribe-pool drain pattern hits. Default 0 = OFF (byte-identical to
+# v1.7.5; tribe-wide cascade still applies as before).
+HUNTER_MAX_AGE="${STAGE3_HUNTER_MAX_AGE:-0}"
 LLM_BACKEND="${STAGE3_LLM_BACKEND:-openai}"
 OPENAI_MODEL="${STAGE3_OPENAI_MODEL:-gpt-4o-mini}"
 OPENAI_ENDPOINT="${STAGE3_OPENAI_ENDPOINT:-https://api.openai.com/v1/chat/completions}"
@@ -431,8 +444,8 @@ write_bootstrap() {
             # file to seed the tribe-<name>:bug_ledger memo from. Without
             # it, hunters verify findings but the JSONL ledger never
             # grows (visible in experience but missing from bug-ledger).
-            printf 'DEATH_THRESHOLD=%s POOL_CAP=%s METABOLIC_COST=%s INITIAL_POOL=%s AGENTIS_ROOT=/run-root/.agentis TARGET_DIR=targets-stage2/smallvec-v0.6.13 TARGET_FILE=lib.rs BUGS_MANIFEST=/run-root/.agentis/sandbox/targets-stage2/bugs.json VERIFIER_PATH=/run-root/tools/verify-finding-stage2.sh BUG_LEDGER_PATH=/run-root/bug-ledger.jsonl bash /run-root/%s/scripts/start-colony.sh > /run-root/%s.log 2>&1 &\n' \
-                "$DEATH_THRESHOLD" "$TRIBE_POOL_CAP" "$TRIBE_METABOLIC_COST" "$TRIBE_INITIAL_POOL" "$tribe" "$tribe"
+            printf 'DEATH_THRESHOLD=%s POOL_CAP=%s METABOLIC_COST=%s INITIAL_POOL=%s HUNTER_MAX_AGE=%s AGENTIS_ROOT=/run-root/.agentis TARGET_DIR=targets-stage2/smallvec-v0.6.13 TARGET_FILE=lib.rs BUGS_MANIFEST=/run-root/.agentis/sandbox/targets-stage2/bugs.json VERIFIER_PATH=/run-root/tools/verify-finding-stage2.sh BUG_LEDGER_PATH=/run-root/bug-ledger.jsonl bash /run-root/%s/scripts/start-colony.sh > /run-root/%s.log 2>&1 &\n' \
+                "$DEATH_THRESHOLD" "$TRIBE_POOL_CAP" "$TRIBE_METABOLIC_COST" "$TRIBE_INITIAL_POOL" "$HUNTER_MAX_AGE" "$tribe" "$tribe"
         done
         printf 'while [ ! -e /run-root/.shutdown ]; do sleep 5; done\n'
         printf 'exit 0\n'
--- a/tribes-bench/tribe-alpha/agents/hunter.ag
+++ b/tribes-bench/tribe-alpha/agents/hunter.ag
@@ -213,6 +213,40 @@ fn tick(reason: string) -> void {
     };
 
 
+    // #487 follow-up: per-hunter age mortality. Each hunter has its own
+    // monotonic age counter keyed on its PPID (stable for the daemon's
+    // lifetime, unique across tribe-mates and replicas). When age exceeds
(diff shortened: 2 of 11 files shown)


ylohnitram:
## QA Report — PR #489 (per-hunter age mortality)

Static-analysis pass against the 11-file diff cross-referenced with smokes #15-#21 evidence + the agentis runtime sandbox-profile + M106 spawn paths. No smoke run executed (per time budget).

### Findings

|---|---|---|---|
| 1. Default-OFF byte-identity | ✅ | n/a | `start-colony.sh` always seeds `hunter:max_age=0` (one extra memo write at colony start, not per-tick). Hunter reads it once per tick (~5 CB, negligible vs 2M budget); `max_age > 0` short-circuits the entire block — no `exec sh` fork, no memo write, no learn(). Functional byte-identity holds. |
| 2. Self-PID stability | ✅ | n/a | Default `SandboxProfile::Basic` (`agentis-core/src/evaluator/mod.rs:494`) spawns `sh -c <code>` directly from daemon-inner, so `$PPID = daemon-inner PID`, stable for daemon lifetime. No `unshare`/`sandbox-exec` wrapping in tribes-bench (no override). M106 spawns each replica via `Command::new(exe).arg("daemon").arg(&source_path)` — distinct OS PID, distinct `$PPID`, distinct memo key. Empty-PID degradation gated by `if len(self_pid) > 0` — hunter falls through to normal tick (acceptable). |
| 3. Memo write race-free | ✅ | n/a | Per-PID key `hunter:<PPID>:age` is written by exactly one process (the hunter that owns that PPID). 60s tick interval + serialized per-daemon ⇒ no self-RMW race. Memo footprint ~23 keys at peak (5 originals + ~18 replicas) — well below `memo.max_keys=500` default. |
| 4. Suicide command shell safety | ✅ | n/a | All `$VAR` substitutions are shell-side (`$PPID`, `$AGENTIS_ROOT`, `$own_id`); zero .ag-side concatenation into the static string passed to `exec sh`. The `\\$` literal in the .ag source emits `\$` to the shell, which under double-quotes anchors literal `$` in the grep regex — verified empirically: `grep -l "^$PPID\$"` matches only the single `<id>.pid` line where the daemon-inner PID sits, not the `<id>.watchdog.pid` (which holds the watchdog's PID, not the daemon-inner's). `xargs -I{} basename {} .pid` is POSIX; `[ -n "$own_id" ]` guards the empty-grep case. |
| 5. Colony-lint clean | ✅ | n/a | `./tools/colony-lint.sh` → 168 passed, 0 failed, 3 skipped (matches PR claim). |
| 6. Cascade interaction | ✅ | n/a | Per-hunter suicide path uses `agentis daemon stop --async --grace-ms 1000 --timeout-ms 5000`. The existing tribe-pool cascade fires from a separate code path (pool < death_threshold) in a different branch of `tick()`. Both paths short-circuit with `return;` — a hunter cannot double-kill itself in one tick. Across hunters they are independent processes; cascade-killer signaling siblings via `daemon stop` is idempotent (already-stopping daemon swallows the second stop file). |
| 7. v1.7.5 colony plumb integration | ✅ | n/a | M106 path: `agentis-core/src/colony/worker.rs:1869-1893` (per `agentis-core/CLAUDE.md` v1.7.4 entry) translates `0x10 → ["--enable-replication", "--allow-replica-replication"]` and spawns replica daemon-inner with its own PID. `$PPID` inside replica's `exec sh` resolves to that replica's daemon-inner PID — distinct from parent and siblings. Per-hunter age block works for replicas without modification. |
| 8. Memo footprint | ✅ | n/a | Per the prompt: 5 originals + ~18 replicas at peak (smoke #20) × 1 key each = ~23 keys. Default ceiling 500. Watchdog respawn creates orphan `hunter:<old-PPID>:age` rows — not GC'd, but accumulation over a single 6h smoke is bounded by `~max_concurrent_daemons × max_respawn_count` which still sits well below 500. Documented as expected behavior in PR comment block. |
| 9. Plumbing trace | ✅ | n/a | `STAGE3_HUNTER_MAX_AGE` → `run-stage3-docker.sh:180` (`HUNTER_MAX_AGE="${STAGE3_HUNTER_MAX_AGE:-0}"`) → bootstrap printf line 447-448 (`HUNTER_MAX_AGE=%s` slot 5, with `"$HUNTER_MAX_AGE"` argument 5) → start-colony.sh consumes `${HUNTER_MAX_AGE:-0}` → `agentis memo set "hunter:max_age" ...` → hunter.ag `recall_latest("hunter:max_age")`. End-to-end consistent across all 5 tribes (alpha/beta/delta/epsilon/gamma; each tribe's `start-colony.sh` adds the same line at line 202, and each `hunter.ag` adds the identical 34-line block after the existing metabolic block). |
| 10. Documentation header | ✅ | n/a | `run-stage3-docker.sh:56-62` header comment block + `:175-179` inline definition both correctly describe the pattern (per-hunter age, `daemon stop` suicide, default 0 = OFF, byte-identical to v1.7.5, tribe-wide cascade still active as fallback). Consistent with the .ag comment block. |
| 11. Env-var input validation (minor) | ⚠️ | dismissed: matches existing pattern | `STAGE3_HUNTER_MAX_AGE` is NOT added to the integer-validation `for` loop at `:199-209`. Non-numeric values (`STAGE3_HUNTER_MAX_AGE=foo`) pass through silently → memo set to `"foo"` → `parse_int("foo")=0` → block skipped (soft-fail). Same shape as the existing `TRIBE_POOL_CAP` / `TRIBE_METABOLIC_COST` / `TRIBE_INITIAL_POOL` env vars (also missing from the validation loop). Not a regression; documented behavior for the #485 family. |
| 12. CHANGELOG entry (minor) | ⚠️ | dismissed: tribes-bench at VERSION 0.0.0 (research scaffold) | `tribes-bench/CHANGELOG.md` `[Unreleased]` section not updated. The federation is not formally release-tracked (VERSION = 0.0.0; per CLAUDE.md only `dev-apprenticeship/` and `federation-dashboard/` are versioned components in `check-changelog.sh`). Optional polish: add an `[Unreleased]` bullet referencing this PR for the eventual tribes-bench release — but not blocking. |
| 13. Test coverage for env-var plumbing (minor) | ⚠️ | optional follow-up | `tools/test-tribes-bench-run-stage3-docker.sh` does not assert on `HUNTER_MAX_AGE=%s` plumbing in the bootstrap printf. Existing tests cover dispatch shape, not specific env-var pass-through (same gap as #485 / #487 ships before). Not blocking; consider an optional follow-up adding a one-liner assertion to lock the plumbing if a future refactor reorders the printf args. |

### Notable observation (informational, not a finding)

Watchdog respawn invalidates the per-PID counter: a daemon-inner crash → watchdog respawn produces a new PID → fresh `hunter:<NEW_PID>:age=0`, while the prior `hunter:<OLD_PID>:age=N` becomes orphan memo cruft. For a HUNTER_MAX_AGE-driven policy this means watchdog respawn effectively resets the age clock — a feature, not a bug, since respawn already implies "this hunter died once and was rebuilt." Worth knowing when interpreting smoke #22 self-death telemetry: if a hunter respawns under watchdog, its `hunter:self_death` learn() rows can stretch past `max_age × tick_interval` wall time.

### Verdict

✅ **ship-it**

All HIGH-severity test plan items pass. Three LOW/MINOR observations (env-var validation, CHANGELOG, test coverage) are pattern-matching with existing #485/#487 shape — none block merge. Default-OFF guarantee verified via static analysis: with `STAGE3_HUNTER_MAX_AGE` unset, the only observable change vs main is one extra `agentis memo set "hunter:max_age" "0"` at colony-start time (5 calls total across the federation, one per `start-colony.sh`) plus one extra `recall_latest` + `parse_int` per hunter tick. Both are functionally inert.

Recommend running smoke #22 (`HUNTER_MAX_AGE=20 INITIAL_POOL=20000`) post-merge to validate the dynamics-side acceptance criteria the PR header describes (≥10 spawns AND ≥30% mortality, replication continuing through hunter deaths).

PR #602, opened 2026-05-17
fix(claim-auditor): cross-federation memo recall — probe laptop-node/.agentis (closes #601)

ylohnitram:
Closes #601.

## Summary

`claim-auditor/tools/run-auditor.sh` was unable to recover the original math-foundry problem context (formulator's `problem_text`, `answer`, `novelty_claim`) for any NOVEL/BORDERLINE row it processed. The orchestrator's `os.path.isdir(<run-dir>/.agentis)` probe always failed because math-foundry writes its memo store one level deeper at `<run-dir>/laptop-node/.agentis/`. The `have_upstream_memo` flag stayed False, no `agentis memo get` was ever invoked, and the orchestrator fell back to seeding the ledger row JSON as the "problem text".

End-to-end result on first smoke (12 min, 6 BORDERLINE rows audited): every verdict was `NEEDS_HUMAN` with confidence 0.4 and reasoning beginning "The problem input is a system event record (novelty verdict classification), not a mathematical claim" — the auditor was reasoning correctly given garbage input.

## Fix

Probe both directory shapes — `<run-dir>/.agentis/` and `<run-dir>/laptop-node/.agentis/` — and set the subprocess cwd to whichever exists. This is the same pattern already used (correctly) by `preprint-foundry/tools/run-preprint.sh`. ~15 line change, one file.

## Validation

- `bash -n claim-auditor/tools/run-auditor.sh`: PASS
- `tools/colony-lint.sh`: 310 passed, 0 failed, 3 skipped
- Manual probe: from a math-foundry run dir, the patched logic finds `laptop-node/.agentis/memo` and `agentis memo get formulator:36:problem_text:tick-5` returns the actual Schur Ramsey problem text.

## Test plan

- [ ] `tools/colony-lint.sh` passes
- [ ] Smoke run against big run #2 reconstructed ledger produces at least one `VERIFIED_NEW` verdict on the known-NOVEL P_2 find (tick 82) and at least one `KNOWN_PRIOR` on a classical BORDERLINE (e.g., Schur Ramsey, tick 5)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #594, opened 2026-05-17
fix(math-foundry): pipeline wiring + Claude CLI backend + seed corpora

ylohnitram:
## Summary

Phase 1 scaffold (#593) shipped the math-foundry runtime but a 5-tribe smoke produced silent downstream colonies — no surprises, problems, verdicts, or novelty memos written end-to-end. This PR makes the chain produce, and adds the Claude CLI backend plus the missing seed paper corpora.

### Pipeline wiring fixes
- **PID self-registration**: explorer writes its own pid under `replay:current_explorer_pid` (other colonies already did this for their respective keys; explorer was the missing link).
- **Phased read offsets**: noticer reads `upstream_tick = tick_idx - 1`, formulator `-2`, verifier `-3`, novelty `-4`. Memo writes use upstream_tick so settlement aligns on a common reference (HOLD_PERIOD=4 already matched the depth).
- **Single-LLM verifier**: dropped the internal solver step. Was two serial LLM calls (~7 min/tick on Opus medium), overrunning the 90s tick window. Verifier is now strictly a problem-quality referee.
- **Verifier prompt rewrite**: ACCEPT when problem is well-defined; answers_agree and solver_rigorous become deprecated metadata.
- **Novelty trigger relaxed**: accepts both ACCEPT and NEEDS_REVISION verdicts (without this, NEEDS_REVISION blocked novelty entirely for empirical computational problems).
- **DSL syntax**: agentis DSL parser does not accept `&&` — rewrote `verdict_label != "ACCEPT" && verdict_label != "NEEDS_REVISION"` as nested `if`/`else`. Was producing a parse error on daemon startup, killing novelty silently.

### Claude CLI backend
`run-foundry.sh` adds `llm.backend=claude` path: spawns `claude -p --output-format json --model {opus|sonnet|haiku} --tools "" --system-prompt minimal --effort {low|medium|high}`. Threaded via `FOUNDRY_CLAUDE_MODEL`, `FOUNDRY_CLAUDE_EFFORT`, `FOUNDRY_HOST_CLAUDE_DIR`. Host `.claude` directory bind-mounted `rw,z` (SELinux relabel per tribes-bench #540 precedent).

### Seed corpora
`data/papers/{number_theory,combinatorics,abstract_algebra,graph_theory}.json` — 6 arxiv abstracts per topic, each carrying `description` and `compute_hints` for sympy/numpy/networkx. Without these the orchestrator exits with `missing corpus for topic` before spawn.

## Smoke validation

[...]

PR #604, opened 2026-05-17
fix(claim-auditor): use replay:current_formulator_pid for memo recall (closes #603)

### LINKED ISSUE #603: claim-auditor: orchestrator uses novelty pid instead of formulator pid
After fix #602, orchestrator now finds the math-foundry memo store but looks up wrong keys. Ledger row carries pid=novelty_daemon_pid (33 for big run #2). Formulator memos are written under pid=formulator_daemon_pid (36). So agentis memo get formulator:33:problem_text:tick-5 returns empty and orchestrator falls back to seeding ledger row JSON.

Fix: in claim-auditor/tools/run-auditor.sh, before per-tick loop, recover the formulator and explorer pids from upstream memo: upstream_recall(replay:current_formulator_pid) and upstream_recall(replay:current_explorer_pid). Use these instead of row.get(pid) for formulator/explorer-keyed lookups.

Test plan: claim:problem_text:tick-0 should contain actual problem text. At least one VERIFIED_NEW or KNOWN_PRIOR verdict instead of all NEEDS_HUMAN.

ylohnitram:
Closes #603.

## Summary

Second bug in the same code path as #602. After fixing the laptop-node path probe, the orchestrator successfully reaches the upstream memo store but dereferences the wrong pid: it uses `row.get("pid")` (novelty daemon pid = 33) when querying formulator keys (which are written under formulator daemon pid = 36).

Result: all cross-federation memo recalls return empty, fallback seeds the ledger row JSON again, searchers and auditor produce another batch of `NEEDS_HUMAN`.

## Fix

Before the per-tick loop, recover formulator + explorer pids from `replay:current_formulator_pid` and `replay:current_explorer_pid` (math-foundry's self-registered keys per PR #594). Use those for all formulator/explorer-keyed lookups. The novelty pid is still passed through to the audit-trail as `claim:source_pid`.

## Validation

- `bash -n claim-auditor/tools/run-auditor.sh`: PASS
- `tools/colony-lint.sh`: 310 passed, 0 failed, 3 skipped
- Manual recall: `agentis memo get replay:current_formulator_pid` returns 36; `agentis memo get formulator:36:problem_text:tick-5` returns the Schur Ramsey problem text.

## Test plan

- [ ] After merge, smoke run against big run #2 ledger produces non-empty `claim:problem_text:tick-0` (= Schur Ramsey, not ledger JSON).
- [ ] At least one `VERIFIED_NEW` or `KNOWN_PRIOR` verdict in audit-ledger (not uniform `NEEDS_HUMAN`).

## Out of scope (= follow-ups if needed)

- `preprint-foundry/tools/run-preprint.sh` may have the same latent bug in its math-foundry recall path; check after this merges.

### DIFF
--- a/claim-auditor/tools/run-auditor.sh
+++ b/claim-auditor/tools/run-auditor.sh
@@ -398,6 +398,13 @@ def upstream_recall(key):
 
 source_run_id = os.path.basename(upstream_dir) or "unknown"
 
+# The ledger row's `pid` is the novelty daemon's pid (it emitted the
+# verdict event), not the formulator's. Formulator/explorer write memo
+# keys under their own daemon pids, so we look those up once from the
+# upstream's self-registered replay keys.
+upstream_formulator_pid = upstream_recall("replay:current_formulator_pid")
+upstream_explorer_pid = upstream_recall("replay:current_explorer_pid")
+
 for idx in range(total_ticks):
     if idx >= len(candidates):
         with open(log_path, "a") as log:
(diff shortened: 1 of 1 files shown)


ylohnitram:
Local CI: colony-lint 310/0/3, bash -n PASS. Manual recall confirms upstream_recall now finds formulator memo keys correctly. Billing-blocked CI, admin merge.

Issue #705, opened 2026-05-21
federation-dashboard: auto-regen serves stale collector verdict; manual /refresh flips to correct state


ylohnitram:
## Problem

`federation-dashboard` reports stale `pid_alive=false` verdicts even when:
- `FEDERATION_DASHBOARD_STALENESS_TICKS=60` is set in the dashboard process env (verified via `/proc/<pid>/environ`).
- Daemon memo timestamps are within the 60×60s = 3600s freshness window (e.g. 1937s = 32 min ago < 3600s).
- Manual collector probe with the same env returns `pid_alive=true` for the same agents.

**Force `POST /refresh` immediately flips the dashboard to correct verdict.** That means the auto-regen cycle is serving a stale collector verdict somewhere between dashboard restart and the next forced refresh.

## Reproducer

1. Kill running dashboard process, restart with `FEDERATION_DASHBOARD_STALENESS_TICKS=60 setsid bash -c "federation-dashboard <fed-dir> 8420 &"`.
2. Verify env in `/proc/$DASH_PID/environ` — `STALENESS_TICKS=60` present.
3. Wait 5-10 minutes for auto-regen.
4. Probe `curl http://localhost:8420/` → 0/N agents alive even though memo timestamps are fresh enough for the 60×60s window.
5. `curl -sX POST http://localhost:8420/refresh` → next probe correctly shows N/N agents alive.

Observed live during research-foundry Run #9 (research-foundry/runs/20260521T001537Z/) — daemons heartbeating correctly, memo store healthy, but dashboard banner stuck on DEGRADED until manual `/refresh`.

## Hypotheses

Three possible root causes (need to inspect `federation-dashboard-server.py` to confirm which):

### A. Auto-regen subprocess inherits stale env

The dashboard server's auto-regen watcher may have been spawned **before** the operator restarted the parent process with the new STALENESS env. Server-side timer/thread holds a reference to an old subprocess invocation that uses the pre-restart default `STALENESS_TICKS=3`. Manual `/refresh` is a `do_POST` handler that builds env via `dict(os.environ)` at request time — picks up the current env.

### B. Cached collector output served verbatim

[...]

Issue #700, opened 2026-05-20
federation-dashboard: STALENESS_TICKS=3 too aggressive for listen-driven agents (noticer/skeptic always show DEGRADED in research-foundry)


ylohnitram:
## Problem

Dashboard reports noticer + skeptic as `pid_alive=false` in research-foundry runs, even after #697/#698 fix (top-of-tick heartbeat). Run #5 observation at 22:42:57Z:

- noticer:last_check = 22:37:43Z (5m14s ago, 314s)
- skeptic:last_check = 22:38:29Z (4m28s ago, 268s)
- explorer:last_check = 22:41:53Z (1m04s ago, 64s) ✓
- formulator/verifier/auditor/editor: ~2 min ago ✓

`STALENESS_TICKS=3 × tick_interval=60s = 180s` window → noticer (314s) and skeptic (268s) both exceed → false-positive stale.

## Root cause

`research-foundry/noticer/agents/noticer.ag` and `skeptic/agents/skeptic.ag` are **event-driven listeners**, not regular tick agents:

```
// noticer.ag
listen("math-foundry:exploration_done") triggers the agent

// skeptic.ag  
listen("math-foundry:notice_done") triggers the agent
```

`fn tick()` is called ONLY when the upstream event fires:
- noticer waits for explorer's `exploration_done` → explorer's `prompt()` call can take 90-120s, so noticer ticks every 90-120s, not every 30s.
- skeptic waits for noticer's `notice_done` → adds another 30-90s delay → skeptic ticks every 2-4 minutes.

The top-of-tick heartbeat from #698 fires when tick() executes — but for these listen-driven agents, "tick() executes" doesn't mean "every 30s". The dashboard's freshness probe assumes daemon ticks at `tick_interval` cadence, which doesn't match listen-driven semantics.

## Why explorer / formulator / verifier / etc. work

Most colonies are regular `tick()` agents with `recommend(<topic>)` patterns that fire each 30s daemon tick. They write heartbeat at the predictable cadence the dashboard expects.

Only noticer + skeptic (and possibly a few others — needs survey) use `listen()`.

### A. Bump STALENESS_TICKS default in dashboard (smallest, recommended)

`federation-dashboard/lib/federation-dashboard-collector.py` STALENESS_TICKS=3 hardcoded. Bump to `10` (= 10 × 60s = 600s = 10 min) for research-foundry. Covers typical inter-event gaps for listen-driven agents.

[...]

Issue #697, opened 2026-05-20
research-foundry: 15/18 .ag agents skip last_check on early-return gates — dashboard DEGRADED despite live daemons


ylohnitram:
## Problem

`federation-dashboard` post-#686 (memo-freshness liveness) reports 15/18 research-foundry agents as `pid_alive=false` → DEGRADED banner persists even when daemons are ticking and the container is healthy.

Confirmed via 3rd live run (`runs/20260520T212733Z/`, container Up 11 minutes):

- 18/18 daemons running per `pgrep -af "agentis daemon-inner"`.
- Only 3 `<colony>:last_check` memos written: `explorer:last_check` (48 entries), `noticer:last_check` (23), `skeptic:last_check` (18).
- Other 15 colonies: zero `last_check` entries despite their daemons being alive.

## Root cause

CLAUDE.md mandates `memo_write("<agent_name>:last_check", now)` at the end of every tick. The convention assumes `fn tick()` reaches end. But in research-foundry, **15/18 colonies have early-return gates** in `fn tick()` that exit before the syntactic end:

| Colony | last_check at line | early `return;` before it |
|--------|--------------------|-----------------------------|
| `explorer` | 614 | 0 |
| `noticer` | 360 | 0 |
| `skeptic` | 398 | 0 |
| `formulator` | 448 | 5 (lines 39, 44, 55, 63, 68 — gates on missing upstream) |
| `verifier` | 384 | similar |
| `novelty` | similar | similar |
| ...12 more | | |

When formulator's upstream (skeptic verdict) isn't ready, it `return;` at line 39 → memo_write never executes. Daemon is alive and ticking, just stuck on a missing-upstream gate. Dashboard correctly reports "no memo activity in last 3*60s" → flips to `pid_alive=false`.

This is a **producer-consumer semantic mismatch**, not a dashboard bug:
- Dashboard expects `last_check` to track "process is alive" liveness.
- `.ag` pattern writes `last_check` to track "tick reached end" completion — different semantics.

## Proposed fix

Move `memo_write("<colony>:last_check", now)` to the **START** of `fn tick()` body in all 15 affected `.ag` files (between parameter parsing and any early-return gate). Idea: liveness = "daemon entered tick", not "tick produced output".

This matches what the dashboard wants and what `pid_alive` actually means. The 3 colonies that work (explorer/noticer/skeptic) effectively already do this by having no early returns.

[...]

Issue #683, opened 2026-05-19
federation-dashboard: PID-based liveness false-negatives containerized federations (research-foundry, tribes-bench)


ylohnitram:
## Problem

Federation-dashboard reports `DEGRADED` banner and `PID ALIVE ❌` red crosses in the Recovery tab when monitoring research-foundry (and tribes-bench), even though every daemon is healthy and ticking.

## Root cause

`lib/federation-dashboard-collector.py:230` does `os.kill(pid, 0)` from the **host** to test daemon liveness. But research-foundry / tribes-bench daemons run **inside a podman container** with a container-internal PID namespace. The host cannot signal those PIDs (`kill -0` returns ESRCH), so:

- `rec['pid_alive']` resolves to `False` for every daemon.
- Every `state=running` daemon shows up in the Recovery tab as PID-ALIVE-X.
- The DEGRADED banner triggers (per template line 1984: `any running daemon with pid_alive === false`).

Verified by reading the registry inside the container:

```
podman exec research-foundry-laptop bash -c 'pgrep -c -f "agentis daemon-inner"'  # 39
# from host:
kill -0 476  # /bin/bash: kill: (476) - No such process
```

## Why this works for dev-apprenticeship today

dev-apprenticeship spawns `start-colony.sh` host-side; daemon PIDs ARE host PIDs and `os.kill(pid, 0)` resolves correctly. The mismatch only surfaces for containerized federations.

## Proposed fix

Add a `--container <name>` flag to `federation-dashboard`. When set:
- Collector resolves daemon PIDs by invoking `podman exec <container> kill -0 <pid>` instead of `os.kill(pid, 0)`.
- Or: collector reads `/proc/<host-side-pidmap>` if podman exposes it (rootless slirp4netns doesn't, so probably no).
- Or: collector trusts `agentis daemon list` state field plus a freshness check (memo write within last N seconds) — bypasses PID-based liveness entirely.

Simplest of the three: **freshness-based liveness**. Each daemon's `<agent>:last_check` memo (already written every tick by the `colony-lint`-mandated pattern) gets compared against `now - N×tick_interval`. If memo is fresh, daemon is alive; PID check skipped.

This eliminates the host/container PID-namespace mismatch entirely and degrades gracefully when `agentis daemon list` is stale (per [feedback_daemon_liveness_check](https://github.com/Replikanti/agentis-colonies/issues/) memo).

## Acceptance criteria

[...]

PR #708, opened 2026-05-21
fix(auto-promote): memo-freshness liveness check unblocks lifecycle for containerized federations (#706)

### LINKED ISSUE #706: auto-promote + cull-replicas trust 'agentis daemon list' state (lies for containerized federations) — entire lifecycle/evolution silently no-op'd
## Problem

`tools/auto-promote.sh` and `tools/cull-replicas.sh` both trust `agentis daemon list` state field to decide whether a daemon is alive. The registry reports `"state": "zombie"` for all daemons in containerized federations (research-foundry, tribes-bench) even when the daemons are demonstrably alive and ticking. Result: **auto-promote skips 16/18 daemons every cycle, cull cycle no-ops, lifecycle defaults (#682) achieve nothing, and Phase 5 cross-run fitness aggregation has nothing to aggregate.**

This is the same `agentis daemon list` liveness lie that #683 documented and #686/#697 fixed on the dashboard side. Auto-promote + cull haven't been migrated to memo-evidence.

## Reproducer

Live observation from research-foundry Run #10 (`runs/20260521T013154Z/`, container Up ~2 min):

```
[2026-05-21T01:31:56Z] auto-promote: SKIP theorist (theorist): effective_state='zombie' not running
[2026-05-21T01:31:56Z] auto-promote: SKIP skeptic (skeptic): effective_state='zombie' not running
[2026-05-21T01:31:56Z] auto-promote: SKIP verifier (verifier): effective_state='zombie' not running
... 14 more SKIPs ...
[2026-05-21T01:31:57Z] auto-promote: Done. Promotes=0, Evolves=0, Skips=18
```

Meanwhile inside the container:
```
podman exec research-foundry-laptop pgrep -c -f "agentis daemon-inner"
# 39  ← all alive
```

After Run #9 finishing (30-tick run, 60 min):
- `persistent/run-history.jsonl`: `by_specialty: {}` (empty — no fitness data aggregated)
- `persistent/fittest_specialties.json`: `ranked: []` (empty — no specialties promoted)
- replication-ledger.jsonl: only the 5 initial spawn rows, **zero replicate/cull/respawn events for the whole run**

So the federation runs the full 30-tick cycle but doesn't evolve at all — the entire #682 lifecycle work + Phase 5 cross-run learning is silently no-op'd by the registry lie.

## Root cause

`tools/auto-promote-decisions.py` and `tools/cull-replicas.sh` both query `agentis daemon list --json` and read the `state` field. For containerized federations, that field is unreliable (registry holds stale records from previous starts, daemons spawned via in-container watchdog don't update the host-visible registry, etc. — see #683 thread for full diagnosis). Dashboard side migrated to **memo-freshness check** (`<colony>:last_check` memo + STALENESS_TICKS window, #697 + #700). Auto-promote + cull haven't.

## Fix

Apply the same memo-freshness liveness check used by `federation-dashboard-collector.py` (after #697 / #700):

```python
# Instead of:
state = daemon.get('state')
if state != 'running':
    skip("effective_state='zombie' not running")

# Use:
last_check = read_memo(f"{colony}:last_check")  # ISO-8601
if last_check and (now() - parse(last_check)) < staleness_window:
    treat_as_alive = True
```

Apply to:
1. `tools/auto-promote-decisions.py` — the `effective_state` derivation.
2. `tools/cull-replicas.sh` — the daemon-iteration loop that decides which pids to consider for culling.
3. (Optional) `tools/auto-promote.sh` — wrapper logic if it also checks state independently.

Match dashboard's `STALENESS_TICKS * tick_interval_ms` formula; reuse `FEDERATION_DASHBOARD_STALENESS_TICKS` env knob OR add federation-aware `AUTOPROMOTE_STALENESS_TICKS` (recommend reuse for simplicity).

## Acceptance criteria

- [ ] Auto-promote sidecar on research-foundry Run with all 18 daemons heartbeating sees ≥ 16/18 as "running" (not "zombie").
- [ ] Daemons that legitimately stopped heartbeating (`last_check` > staleness window) correctly flagged as zombie.
- [ ] At least one promote OR evolve OR replicate event observed in a standard 30-tick research-foundry run (assuming explorer prompts produce non-trivial claims).
- [ ] `persistent/run-history.jsonl` after run completion has non-empty `by_specialty` for at least one specialty (= fitness aggregation has data).
- [ ] No regression on dev-apprenticeship (which has reliable registry state because daemons spawn host-side).

## Severity

**Critical** — the entire lifecycle / evolution / cross-run-learning value proposition of research-foundry is silently no-op'd. Operators see "auto-promote running, sidecar healthy" in the dashboard but no actual promotions happen for the lifetime of the federation.

## Related

- #683 (dashboard PID-kill → memo-freshness migration, same lie, same fix-shape)
- #686 / #697 / #700 (dashboard side fixes that proved memo-freshness works)
- #682 (lifecycle defaults ON, blocked from achieving anything by this bug)
- #626 / #692 / #693 / #695 (Phase 5 cross-run persistence, gets empty `by_specialty` due to this)
- Memo `feedback_daemon_liveness_check`: "agentis daemon list lže o liveness — vždy cross-check přes /proc / pgrep -af 'agentis daemon-inner', ne grep na registry."

## Discovered via

Live Run #10 observation immediately after spawn. 18 SKIPs in first auto-promote sidecar tick = 100% false-negative rate.

ylohnitram:
## Summary

- `tools/auto-promote-decisions.py` containerized branch now trusts `<agent>:last_check` memo freshness over `effective_state`, mirroring the dashboard's #683 logic.
- Daemon registry `effective_state` lies after sleep/reboot — the field stays `zombie` even when the in-process tick loop keeps writing the memo. Pre-#706 every containerized agent fell into the `effective_state='zombie' not running` skip branch, blocking promotions, evolves, and (downstream) cull-replicas inputs.
- Memo freshness is PID-namespace independent: the predicate works whether the dashboard / sidecar runs inside or outside the container.

## Background

This continues the registry-liveness fix chain: #683 (dashboard PID liveness → memo freshness), #697 (per-repo fallback), #700 (env-configurable `FEDERATION_DASHBOARD_STALENESS_TICKS`), #705 (banner alignment). The dashboard was correct after #683, but `auto-promote-decisions.py` still consumed the stale registry field — so the dashboard rendered HEALTHY while the sidecar refused to promote anything.

## Implementation

The three helpers (`_read_memo_raw`, `_parse_last_check_epoch`, `_resolve_tick_interval_ms`) are mirrored verbatim from `federation-dashboard/lib/federation-dashboard-collector.py:85-134` with a `keep in sync with #683/#697/#700/#705` banner. `STALENESS_TICKS` reuses the same `FEDERATION_DASHBOARD_STALENESS_TICKS` env knob so a single setting controls both the dashboard and the sidecar.

The containerized gate is now:

```python
if containerized:
    last_check_epoch = _parse_last_check_epoch(_read_memo_raw(fed_dir, agent_name + ':last_check'))
    fresh = False
    if last_check_epoch is not None:
        tick_ms = _resolve_tick_interval_ms(agent_name, colony, fed_dir)
        fresh = (time.time() - last_check_epoch) < STALENESS_TICKS * (tick_ms / 1000.0)
    effective_state = d.get('effective_state') or state
    if not fresh and effective_state != 'running':
        # skip with reason "memo last_check stale and effective_state=... not running"
```

Host-side (`containerized=False`) path is untouched — `dev-apprenticeship/` still uses `os.kill(pid, 0)`.

## Test plan

- [x] `bash tools/test-auto-promote.sh` — 37/0 (35 baseline + 2 new tests for #706)
  - Test 18: fresh memo + stale registry → decision does NOT carry `effective_state=... not running` reason
  - Test 19: no memo + stale registry → decision IS skipped with the new memo-aware reason
- [x] `bash research-foundry/tools/test-run-research.sh` — 95/0
- [x] `bash research-foundry/tools/test-replicate-gates.sh` — 29/0
- [x] `bash research-foundry/tools/test-jitter.sh` — 72/0
- [x] `bash research-foundry/tools/test-last-check-early.sh` — 72/0
- [x] `bash tools/colony-lint.sh` — 343/0/4 (baseline)
- [x] `python3 -m ast tools/auto-promote-decisions.py` — syntax clean

## Out of scope

- `tools/cull-replicas.sh` — inherits the fix transitively via `auto-promote-decisions.py` JSON output.
- `federation-dashboard/` — already memo-freshness-driven since #683.

---
Closes #706

### DIFF
--- a/tools/auto-promote-decisions.py
+++ b/tools/auto-promote-decisions.py
@@ -48,6 +48,7 @@
 #
 # Contract is frozen by doc/auto-promote.md and test-auto-promote.sh.
 
+import datetime
 import hashlib
 import json
 import math
@@ -55,6 +56,71 @@ import os
 import subprocess
 import sys
 import time
+
+
+# Mirrored from federation-dashboard-collector.py:85-134; keep in sync with
+# #683/#697/#700/#705. #706 ports the same memo-freshness liveness predicate
+# from the dashboard to this sidecar so the containerized branch below no
+# longer trusts the daemon registry's stale `effective_state` field (the
+# in-process tick loop continues to write `<agent>:last_check` even when
+# the registry record is mis-flagged as `zombie` after sleep/reboot).
+try:
+    STALENESS_TICKS = max(1, int(os.environ.get("FEDERATION_DASHBOARD_STALENESS_TICKS", "3")))
+except (TypeError, ValueError):
+    STALENESS_TICKS = 3
+
+
+def _read_memo_raw(fed_dir, key):
+    try:
+        proc = subprocess.run(
+            ['agentis', 'memo', 'get', key],
+            cwd=fed_dir, capture_output=True, text=True, timeout=2,
+        )
+    except (subprocess.SubprocessError, OSError):
+        return None
+    if proc.returncode != 0:
+        return None
+    raw = (proc.stdout or '').strip()
+    return raw or None
+
+
+def _parse_last_check_epoch(raw):
+    if not raw:
+        return None
+    try:
+        dt = datetime.datetime.strptime(raw, "%Y-%m-%dT%H:%M:%SZ")
+        return dt.replace(tzinfo=datetime.timezone.utc).timestamp()
+    except (TypeError, ValueError):
+        return None
+
+
+_tick_interval_cache = {}
+
+
+def _resolve_tick_interval_ms(agent, colony, fed_dir):
+    cache_key = (agent, colony)
+    if cache_key in _tick_interval_cache:
+        return _tick_interval_cache[cache_key]
+    interval = 60000
+    script_dir = os.path.dirname(os.path.abspath(__file__))
+    helper = os.path.join(script_dir, 'resolve-tick-interval.py')
+    if colony:
+        colony_dir = os.path.join(fed_dir, colony)
+        if os.path.isfile(helper) and os.path.isdir(colony_dir):
+            try:
+                proc = subprocess.run(
+                    ['python3', helper, agent, colony_dir],
+                    capture_output=True, text=True, timeout=2,
+                )
+                if proc.returncode == 0:
+                    try:
+                        interval = int((proc.stdout or '').strip())
+                    except (TypeError, ValueError):
+                        interval = 60000
+            except (subprocess.SubprocessError, OSError):
+                interval = 60000
+    _tick_interval_cache[cache_key] = interval
+    return interval
 
 
 def _load_config(path):
@@ -524,17 +590,26 @@ def main():
         #
         # In containerized mode (#622) the host PID namespace can't see
         # container PIDs, so os.kill(pid, 0) is a false-negative every
-        # tick. Fall back to `effective_state == "running"` from the
-        # daemon-list JSON (also valid for legacy mode; the runtime
-        # tracks zombie/running/etc. with a stale-heartbeat probe).
+        # tick. #706: trust memo freshness (`<agent>:last_check` within
+        # STALENESS_TICKS * tick_interval) over `effective_state`, since
+        # the daemon registry's state field stays stale across sleep/reboot
+        # cycles even when the in-process tick loop keeps writing the memo.
+        # `effective_state == "running"` remains a valid fall-through for
+        # legacy callers; the memo path simply unblocks daemons whose
+        # registry record lies.
         if containerized:
+            last_check_epoch = _parse_last_check_epoch(_read_memo_raw(fed_dir, agent_name + ':last_check'))
+            fresh = False
+            if last_check_epoch is not None:
+                tick_ms = _resolve_tick_interval_ms(agent_name, colony, fed_dir)
+                fresh = (time.time() - last_check_epoch) < STALENESS_TICKS * (tick_ms / 1000.0)
             effective_state = d.get('effective_state') or state
-            if effective_state != 'running':
+            if not fresh and effective_state != 'running':
                 decisions.append(_attach_explorer_evidence({
                     'agent': agent_name,
                     'colony': colony,
                     'decision': 'skip',
-                    'reason': f'effective_state={effective_state!r} not running',
+                    'reason': f'memo last_check stale and effective_state={effective_state!r} not running',
                     'pid': pid,
                     'confidence': confidence,
                 }))
--- a/tools/test-auto-promote.sh
+++ b/tools/test-auto-promote.sh
@@ -644,6 +644,81 @@ else
     fail "fed-local priority" "expected entries_total=5, got <$ENTRIES_B> from <$OUT_B>"
 fi
 
+# --- Test 18: containerized liveness — fresh memo overrides stale registry (#706) ---
+# Reproduces the lifecycle-blocked-after-reboot bug: the daemon registry's
+# `effective_state` stays at "zombie" across sleep/reboot cycles, but the in-
+# process tick loop keeps writing `<agent>:last_check`. Pre-#706 the
+# containerized branch trusted `effective_state` and skipped every such agent
+# with "effective_state='zombie' not running" — promotions, evolves, and
+# (downstream) cull-replicas inputs all stalled. Post-#706 a fresh memo
+# (within STALENESS_TICKS * tick_interval) carries the agent past the
+# containerized gate so the rest of the prereq pipeline runs normally.
+FED_LIVE="$TMPDIR_TEST/fed-live"
+mkdir -p "$FED_LIVE"
+ISO_NOW=$(date -u +%Y-%m-%dT%H:%M:%SZ)
+(cd "$FED_LIVE" && agentis memo set "explorer:last_check" "$ISO_NOW" >/dev/null 2>&1)
+
+DAEMONS_LIVE='[{"pid":12345,"agent_id":"explorer-1","source":"explorer.ag","state":"zombie","effective_state":"zombie","colony":"discovery","confidence":0.4,"started_at":1}]'
+
+OUT_LIVE=$(python3 "$SCRIPT_DIR/auto-promote-decisions.py" \
+    --preview --config "$SCRIPT_DIR/auto-promote-config.yaml" --containerized \
+    "$DAEMONS_LIVE" "$FED_LIVE")
+
+LIVE_CHECK=$(python3 -c "
+import json, re, sys
+arr = json.loads(sys.argv[1])
+if len(arr) != 1:
+    print('arity:' + str(len(arr))); sys.exit(0)
+d = arr[0]
+reason = d.get('reason', '')
+if re.search(r\"effective_state=.*not running\", reason):
+    print('regression: got <' + reason + '>')
+else:
+    print('ok')
+" "$OUT_LIVE" 2>&1 || true)
+
+if [ "$LIVE_CHECK" = "ok" ]; then
+    pass "containerized liveness: fresh memo overrides stale registry (#706)"
+else
+    fail "fresh-memo liveness" "$LIVE_CHECK from <$OUT_LIVE>"
+fi
+
+# --- Test 19: containerized liveness — no memo + stale registry skips (#706) ---
+# Counterpart to test 18: with NO `<agent>:last_check` memo present and the
+# registry's `effective_state` still "zombie", the sidecar must skip with
+# the new memo-aware reason string. This guards against the inverse
+# regression — relaxing the gate too far would let truly dead daemons get
+# promoted.
+FED_DEAD="$TMPDIR_TEST/fed-dead"
+mkdir -p "$FED_DEAD"
+
+DAEMONS_DEAD='[{"pid":12345,"agent_id":"explorer-1","source":"explorer.ag","state":"zombie","effective_state":"zombie","colony":"discovery","confidence":0.4,"started_at":1}]'
+
+OUT_DEAD=$(python3 "$SCRIPT_DIR/auto-promote-decisions.py" \
+    --preview --config "$SCRIPT_DIR/auto-promote-config.yaml" --containerized \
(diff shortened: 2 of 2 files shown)


ylohnitram:
## Independent PR review (cluster 2/2)

### Findings

1. **AND gate is correct, not OR**: `if not fresh and effective_state != 'running'` is the right boolean. Truth table: fresh+running proceeds, fresh+zombie proceeds (the #706 fix), stale+running proceeds (trust registry when memo dark), stale+zombie skips. Either signal can rescue the agent. OR would over-restrict (requiring both signals) and re-introduce the bug. **Severity**: dismissed. **Recommendation**: none.

2. **Memo-key shape matches dashboard**: `agent_name = os.path.basename(source)` minus `.ag` suffix (line 494-496) produces the same key prefix the collector uses (`role = os.path.basename(src)` minus `.ag`, line 225-227 of `federation-dashboard-collector.py`). Cross-checked against `research-foundry/explorer/agents/explorer.ag:468` (`memo_write("explorer:last_check", ...)`) and `arxiv-search.ag:302`, `prior_advocate.ag:254` — dashed and underscored basenames both write to bare `<agent_name>:last_check`. **Severity**: dismissed. **Recommendation**: none.

3. **Missing per-repo fallback vs dashboard divergence**: collector at lines 329-340 falls back to `<owner>__<repo>:<agent>:last_check` when the bare key is empty for multi-repo colonies (e.g., dev-apprenticeship with `[[forge.github]]` fan-out). The PR's auto-promote helper only reads the bare key. Today this is harmless because (a) only research-foundry uses `--containerized` and it is single-repo per agent, and (b) dev-apprenticeship doesn't take the `--containerized` branch. But the "keep in sync" banner is already a half-truth — if a multi-repo containerized federation appears, decisions and dashboard liveness will disagree. **Severity**: nit. **Recommendation**: either widen the comment to "per-repo fallback intentionally omitted; bare key only" or port the fallback as a no-op for single-repo callers.

4. **Subprocess-cost rationale for refusing a shared module is wrong**: the plan dismissed extracting `tools/agentis_memo_freshness.py` as a shared module because "subprocess + JSON-parse round-trip per daemon × 60 daemons × per cycle = real cost". But `import` happens once at process startup, not per daemon — the helpers only subprocess at call time, identical in both copy-paste and shared-module shapes. The real friction is the module-level `fed_dir`/`fed_tools_dir` capture in collector vs per-call args in auto-promote, which would need a small refactor. The drift risk between two divergent copies (already 4 cross-references in the banner: #683/#697/#700/#705) is the actual cost being deferred. **Severity**: nit. **Recommendation**: keep this PR as-is, open a separate follow-up to extract once a third caller appears or once divergence bites.

5. **STALENESS_TICKS default vs research-foundry intent**: comment in `federation-dashboard-collector.py:47-48` claims "research-foundry's launcher exports `FEDERATION_DASHBOARD_STALENESS_TICKS=10`" but `grep -rn STALENESS research-foundry/` returns nothing — `run-research.sh` does not actually export it. Both dashboard and sidecar fall back to default 3. This is a pre-existing inconsistency in the dashboard's documentation, not a regression introduced here. Both consumers see the same value, which is the property that matters for cross-checks. **Severity**: nit (pre-existing). **Recommendation**: open follow-up issue against research-foundry to actually export `=10` per its listen-driven cadence, or amend the dashboard comment to drop the false claim.

6. **Subtle behavior change for state=stopped + fresh memo**: pre-PR `--containerized` with `effective_state=stopped` skipped. Post-PR, a fresh memo overrides → proceeds. If an operator intentionally `stop`s a daemon, there is a `STALENESS_TICKS × tick_interval` window where the still-fresh memo will let auto-promote treat it as alive (180s default, 600s at =10). The original bug (zombie registry stuck forever) is far worse than this 3-10 minute false-positive after intentional stop, so the trade is correct. **Severity**: dismissed. **Recommendation**: none, but worth a one-line note in the inline comment about the explicit-stop edge.

7. **`_resolve_tick_interval_ms` signature divergence is acceptable but the fixture is misleading**: PR helper takes `(agent, colony, fed_dir)` and resolves the helper script via `os.path.dirname(__file__)` — works because `resolve-tick-interval.py` sits next to `auto-promote-decisions.py`. Collector takes `(agent, colony)` and reads `fed_tools_dir` from module-scope argv. Test 18's fixture uses `colony="discovery"` (the SIDE name) instead of the actual daemon `--colony explorer` value research-foundry's launcher emits. Test still passes because `<fed_dir>/discovery` doesn't exist → falls back to 60000ms — but the fixture is wrong about real daemon-list shape. **Severity**: nit. **Recommendation**: fix Test 18's colony field to `"explorer"` for fidelity.

8. **Test 18 has no `command -v agentis` guard**: `(cd "$FED_LIVE" && agentis memo set ...)` runs unconditionally. If `agentis` is missing (CLAUDE.md notes 1 skip in CI baseline for that reason), the memo never gets written, `_read_memo_raw` returns None, fresh=False, and the new gate fires the regression-shaped reason — which the LIVE_CHECK regex `effective_state=.*not running` will MATCH (since the new reason contains `effective_state='zombie' not running` as a substring). Test 18 would false-fail on agentis-less runners. Test 19 is immune (it expects this exact path). The PR's reported 37/0 was run with `agentis` present. **Severity**: nit. **Recommendation**: gate Test 18 with `command -v agentis >/dev/null 2>&1 || { skip "..."; continue; }` matching the colony-lint.sh pattern, or scope the LIVE_CHECK regex to exclude the new "memo last_check stale and ..." prefix.

9. **Race condition during cycle is idempotent**: gate is evaluated independently per daemon, no shared mutable state between iterations, no persisted side-effects in the liveness path. Partially-fresh cycle just produces partial decisions; next cycle reflects reality. **Severity**: dismissed.

10. **Timezone-safe**: `_parse_last_check_epoch` returns UTC epoch seconds; `time.time()` returns UTC epoch seconds. Subtraction is timezone-agnostic. **Severity**: dismissed.

### Verdict

⚠️ approve-with-followups

The architectural core is sound: gate semantics correct, key-shape compatible with dashboard, fail-open behavior (stale memo + running registry still proceeds) preserves the property the original `--containerized` branch had. The two real items worth tracking are #4 (helper drift, eventual shared-module extraction) and #8 (Test 18 fragility on agentis-less CI). Neither blocks this PR; both are follow-up issues.

ylohnitram:
## QA Report (cluster 1/2)

| Helpers mirrored from collector | ✅ | none | `_read_memo_raw` + `_parse_last_check_epoch` byte-equivalent; `_resolve_tick_interval_ms` swaps `fed_tools_dir` global for `script_dir` (script-relative discovery) — reasonable adaptation, banner present |
| STALENESS_TICKS env-driven | ✅ | none | `FEDERATION_DASHBOARD_STALENESS_TICKS` env, max(1, int(...)) clamp, `TypeError/ValueError` fallback to 3 — matches collector pattern (#700) |
| Containerized branch patched | ✅ | none | decisions.py:600-616 — `fresh` boolean computed only when memo epoch parsed, skip predicate now `not fresh AND effective_state != 'running'`, reason string `'memo last_check stale and effective_state=... not running'` matches spec |
| Host-side path untouched | ✅ | none | decisions.py:617-638 byte-identical to main (no-pid + `os.kill(pid,0)` block unchanged); dev-apprenticeship unaffected |
| Test 18 (alive+fresh memo) | ✅ | none | `agentis memo set explorer:last_check $(date -u +%Y-%m-%dT%H:%M:%SZ)` + zombie-state fixture; asserts no `effective_state=.*not running` reason via regex — passes |
| Test 19 (dead+no memo) | ✅ | none | no memo + zombie-state fixture; asserts skip with literal `memo last_check stale and effective_state='zombie' not running` — passes |
| Memo key shape | ✅ | none | `agent_name = basename(source).rstrip('.ag')` → fixture `source:"explorer.ag"` yields `explorer:last_check`, matches collector precedent (federation-dashboard-collector.py:327) |
| test-auto-promote.sh | ✅ | none | 37 passed, 0 failed (35 baseline + 2 new) |
| test-run-research.sh | ✅ | none | 95 passed, 0 failed |
| test-replicate-gates.sh | ✅ | none | 29 passed, 0 failed, 0 skipped |
| test-jitter.sh | ✅ | none | 72 passed, 0 failed |
| test-last-check-early.sh | ✅ | none | 72 passed, 0 failed |
| colony-lint.sh | ✅ | none | 343 passed, 0 failed, 4 skipped — baseline preserved exactly |
| Diff scope (2 files) | ✅ | none | `tools/auto-promote-decisions.py` (+80/-5) + `tools/test-auto-promote.sh` (+75/-0); no collateral churn |
| CHANGELOG omission | ✅ | none | tools/ change, dev-apprenticeship federation uses host-side `os.kill` path (untouched); no versioned-component VERSION bump required, `check-changelog.sh` no-ops |
| AST validation | ✅ | none | `python3 -c "import ast; ast.parse(...)"` clean |
| Live container probe (bonus) | ✅ | none | Patched script run against live `research-foundry-laptop` (18 daemons, currently 0 zombie): 18 decisions returned, **zero** `effective_state` skips — happy-path preserved. Once registry goes stale (sleep/reboot), only the patched path keeps lifecycle alive. |
| PR hygiene — closing keyword | ⚠️ | follow-up (1) | PR body lacks `Closes #706` / `Fixes #706` keyword — only references `(#706)` in title and inline mentions. GitHub's `closingIssuesReferences` is empty; issue 706 will stay OPEN after merge unless edited. |
| PR hygiene — title/body | ✅ | none | Title accurate, body explains lifecycle-blocked → unblocked transition + helper-mirror rationale, references #683/#697/#700/#705 chain, lists test sweeps with counts |

### Follow-ups

1. **Add `Closes #706` to PR body before merge.** Without the keyword, issue stays OPEN and the chain-link to #683/#697/#700/#705 is broken from GitHub's UI. One-liner edit, no code change.

## Verdict: ⚠️ ship-with-followup — green tests, fix works on live container, single one-line PR-body edit blocks the merge cleanup

Issue #16, opened 2026-04-10
gitlab-api.sh silently swallows unknown flags


ylohnitram:
## Problem

`dev-apprenticeship/planning/scripts/gitlab-api.sh` parses its own flags with a `case` statement that silently swallows unknown flags:

```bash
while [ $# -gt 0 ]; do
    case "$1" in
        --since) SINCE="$2"; shift 2 ;;
        --state) STATE="$2"; shift 2 ;;
        ...
        *) shift ;;
    esac
done
```

A typo like `--sinse 2026-04-01` silently discards both the flag and its value and the caller never learns the filter was ignored.

## Impact

- Low: callers are agents, not humans. Agents don't typo.
- But: as agents are autogenerated by LLMs, the LLM **can** produce a typo under poor prompting, and the script will cheerfully run without the intended filter.

## Scope

`dev-apprenticeship/planning/scripts/gitlab-api.sh`

Fix is to replace `*) shift ;;` with `*) echo "unknown flag: $1" >&2; exit 2 ;;` in every `case` block, or to consolidate flag parsing in one place at the top.

The same pattern may exist in the code-review colony wrapper and should be checked in the same pass.

## Related

Reported by QA on #8 as NIT-9.

ylohnitram:
## Decision

Replace every `*) shift ;;` catch-all in the flag-parsing `case` statements inside colony `gitlab-api.sh` scripts with an explicit `*) echo "unknown flag: $1" >&2; exit 2 ;;` so typos fail fast.

## Scope

Three files, nine call sites total (confirmed by grep):

- `dev-apprenticeship/planning/scripts/gitlab-api.sh` lines 74, 109, 127.
- `dev-apprenticeship/code-review/scripts/gitlab-api.sh` lines 58, 95, 122.
- `dev-apprenticeship/triage/scripts/gitlab-api.sh` lines 66, 102, 137.

Search anchor in each file: `*) shift ;;`.

Ordering note: land after #10 so planning's file has already shrunk. Two of the three planning hits are inside blocks that #10 will delete.

### Scope adjustment for planning

After #10 lands, planning's `gitlab-api.sh` will only have one remaining `*) shift ;;` site, the `issues)` case (line 74 today, somewhere nearby after #10). If #16 ships before #10, include all three planning hits and let #10 rebase on top.

## Concrete fix

Replace every `*) shift ;;` with:

```bash
                *) echo "unknown flag: $1" >&2; exit 2 ;;
```

[...]

Issue #15, opened 2026-04-10
plan_reviewer.ag 24-char threshold needs explaining comment


ylohnitram:
## Problem

`dev-apprenticeship/planning/agents/plan_reviewer.ag` guards its work on a 24-character threshold:

```
let total_len = len(scope_str) + len(risks_str) + len(decomp_str);
if total_len < 24 {
    return;
};
```

The number 24 is not self-explanatory. It comes from the fact that `to_string(listen(ch, timeout))` returns the literal string `"void"` (length 4) when the listen times out. Three timed-out listens produce `4 + 4 + 4 = 12`, which is below 24. Any single non-empty struct stringifies to well over 24, so a single positive signal passes the gate. This was verified empirically during QA.

## Impact

Zero runtime impact. Logic is correct. But a reader six months from now has to reverse-engineer why the number is 24 and will probably guess wrong.

## Scope

`dev-apprenticeship/planning/agents/plan_reviewer.ag`

Fix is a two-line comment above the check that says what "void" stringifies to and why 24 is chosen.

## Related

Reported by QA on #8 as NIT-8.

ylohnitram:
## Decision

Insert a four-line comment above the `if total_len < 24 { return; };` check explaining the magic number. No behavior change.

## Scope

Single file: `dev-apprenticeship/planning/agents/plan_reviewer.ag`.
Search for `if total_len < 24`. Current location is line 40.

## Concrete fix

Replace lines 38-42:

```
    // Only proceed if at least one planner reported
    let total_len = len(scope_str) + len(risks_str) + len(decomp_str);
    if total_len < 24 {
        return;
    };
```

with:

```
    // Only proceed if at least one planner reported.
    // Threshold 24: to_string(void) returns "void" (length 4), and three
    // timed-out listens yield 4 + 4 + 4 = 12 characters. 24 gives a 2x
    // safety margin before we start running the expensive prompt().
    let total_len = len(scope_str) + len(risks_str) + len(decomp_str);
    if total_len < 24 {
        return;
    };
```

Do not change any code logic. Do not rename or factor the constant into a `let`. This is strictly a documentation change.

## Cross-colony coverage

Planning only, single file.

## Test plan

1. `./tools/colony-lint.sh` must pass including `agentis commit` syntax check of the `.ag` file.
2. Visual diff review, comment only.

## Commit + PR title

`docs: explain 24-char threshold in plan_reviewer.ag`

## SDLC steps

[...]

PR #730, opened 2026-05-22
fix(auto-promote): sidecar resilience + lock orphan detection (#728)

### LINKED ISSUE #728: auto-promote sidecar dies silently on cull tick + lock orphan blocks next runs
## Problem

Run #15 forensic surfaced a 3-layer bug that silently kills the auto-promote sidecar and prevents subsequent runs from acquiring the lock:

**Layer 1 — `set -euo pipefail` propagates into sidecar subshell.**
`research-foundry/tools/run-research.sh:244` sets `set -euo pipefail` at the top. The sidecar wrapper at line 1087 (`( ... ) &`) is a subshell that inherits this. Inside the wrapper's `while :;` loop (line 1090), any transient SIGTERM to a child process (e.g. `agentis daemon list --json` at line 1091, or any agentis call inside `auto-promote.sh`) makes bash treat the non-zero exit as a fatal error and tears down the subshell. The supervisor loop is supposed to be resilient — `|| printf '...exited %s\n'` catches *expected* failures, but `set -e` kills the loop on any *unexpected* child SIGTERM cascade (e.g. operator running `install.sh` that does `pkill agentis` mid-tick).

In Run #15, the sidecar died at tick 5 (cull tick) when the operator-side `install.sh` upgraded host `agentis` to v1.7.15 around 12:10:08+02:00. Last log line: `=== 2026-05-22T12:10:08+02:00: cull tick (#5) ===` — nothing after. Subshell vanished from process tree.

**Layer 2 — lock FD inherited by spawned daemons keeps lock alive after sidecar dies.**
`tools/auto-promote.sh:132` does `exec 200>"$LOCK_FILE"` opening FD 200 on `tools/.auto-promote.lock`. The Python helper at line 133 calls `fcntl.flock(LOCK_EX | LOCK_NB)` on that FD. Subsequent `agentis daemon stop` / `agentis daemon <file>` child processes (line 363, 375) **inherit FD 200** because there's no close-on-exec flag. When the sidecar shell dies, FD 200 closes in the shell but stays open in those agentis daemon child processes — and the flock persists as long as ANY FD references the OFD (per POSIX).

In Run #15, after the sidecar died around 12:10, the orphan daemons it had spawned in earlier ticks (formulator from `20260521T001537Z`, PID 1054445 + 1054448 watchdog+daemon-inner) kept FD 200 open for ~14 hours, blocking every subsequent sidecar run with "Another auto-promote instance is running. Exiting." Forensic: `fuser tools/.auto-promote.lock` showed `1054445 1054448` as holders even after the shell sidecar was gone.

**Layer 3 — `auto-promote-lock.py` has no stale-holder detection.**
The helper just calls `fcntl.flock(LOCK_EX | LOCK_NB)` and exits 1 on contention. It doesn't check whether the current holder PID is alive or whether any holder shell process is still running. A graveyard orphan from a previous session can block all future runs forever.

## Reproducer

```
# Run #15 timeline:
11:50:07Z — sidecar starts (Run #15 orchestrator at 09:50:07 UTC)
11:50:07Z — tick 1: "Another auto-promote instance is running. Exiting."  (orphan from prior session held lock)
11:55:07Z — tick 2: same
12:00:07Z — tick 3: same
12:05:07Z — tick 4: same
12:10:08Z — tick 5 + cull tick (#5):  "cull tick (#5) ===" — then silence
        (operator install.sh upgraded host agentis to v1.7.15, pkill cascade caught child agentis call)
        (set -e in subshell + SIGTERM'd child = subshell dies)
12:10:08Z — orphan daemons spawned by sidecar in previous-session ticks keep FD 200 open
13:23+   — operator notices, kills orphans manually via `kill 1054445 1054448`
13:23+   — lock released, but Run #15 sidecar already dead (no auto-restart)
```

## Acceptance criteria

- [ ] **Layer 1 fix**: Sidecar subshell explicitly disables `set -e` (`set +e` after opening the subshell). The existing `|| printf '...exited %s\n'` lines already handle expected failures; the supervisor loop must be resilient to any unexpected child exit.
- [ ] **Layer 2 fix**: `tools/auto-promote.sh` adds close-on-exec semantics for FD 200 — either via `exec 200>"$LOCK_FILE"` followed by `exec 200>&200` with bash 5 close-on-exec dialect, OR by explicitly closing FD 200 (`200>&-`) on every `agentis daemon` / `agentis daemon stop` invocation inside the script. Spawned daemons must NOT inherit the lock FD.
- [ ] **Layer 3 fix**: `tools/auto-promote-lock.py` adds stale-holder detection — when `flock` reports contention, read the lock file's content (PID written there) or use `fuser`/`/proc/<PID>/cmdline` to verify the holder is still a live auto-promote process. If the holder is dead or no holder remains, force-acquire by overwriting and retrying flock. Log the steal as `[lock] stole stale lock from dead PID <X>`.
- [ ] **Layer 4 (stretch)**: orchestrator-side supervisor — `run-research.sh` watches `AUTO_PROMOTE_PID` and restarts the sidecar subshell if it dies before the run completes.
- [ ] Regression test in `tools/test-auto-promote.sh` or new `test-auto-promote-resilience.sh`:
  - simulate sidecar subshell child SIGTERM → subshell survives the loop iteration
  - simulate lock holder dead → next acquire detects + steals
  - close-on-exec assertion: spawn `agentis daemon` inside lock-held context, verify FD 200 isn't visible in child's `/proc/<pid>/fd/`

## Out of scope

- Full supervisor process (`systemd --user` unit, podman healthcheck container) — bigger architectural change.
- Per-fed-dir lock file (would prevent cross-fed contention but is a separate design choice).

## Surfaced from

Run #15 forensic — same session that landed P1-P4 fixes. Run #15 burned ~2 hours of operator time + token budget producing no auto-promote signal because of this cascade. Operator memo recorded: "monitor must detect absence of expected change" → `feedback_monitor_eagerly`.

ylohnitram:
## Summary

- **Layer 1**: `set +e` + `set +o pipefail` at the top of the sidecar subshell in `run-research.sh` so a transient SIGTERM to a child `agentis` call no longer tears down the supervisor loop.
- **Layer 2**: `200>&-` appended to all 3 `agentis daemon` spawn sites in `auto-promote.sh` so spawned daemons stop inheriting the lock FD that otherwise outlives the parent shell.
- **Layer 3**: `auto-promote-lock.py` writes its PID on acquire and probes `os.kill(pid, 0)` + `/proc/<pid>/cmdline` on contention; dead-or-recycled PIDs trigger a steal instead of a false "another instance running" exit. Helper now takes 2 args (FD + lock file path).

## Why

Run #15 forensic showed the auto-promote sidecar exiting at tick 5 of 45 (a child `agentis daemon list` was caught in an operator-side `pkill` cascade, and the inherited `set -e` from the parent shell tore down the subshell). The lock file FD it left behind, inherited by previously-spawned `agentis daemon` processes via the close-on-exec gap, then held the OFD-based `flock` alive for 14 hours and silently no-op'd every subsequent auto-promote run.

The three layers are independent so any one can be reverted in isolation; together they cover prevention (L2), survival (L1), and recovery (L3).

## Test plan

- [x] `bash -n research-foundry/tools/run-research.sh` clean
- [x] `bash -n tools/auto-promote.sh` clean
- [x] `python3 -c "import ast; ast.parse(open('tools/auto-promote-lock.py').read())"` clean
- [x] `tools/test-auto-promote.sh`: 40 passed / 0 failed, including 3 new `#728` cases (dead PID steal, recycled PID steal, real contention still returns exit 1)
- [x] `tools/colony-lint.sh`: 344 passed / 0 failed / 4 skipped
- [ ] CI green
- [ ] After merge: a Run #16 spawned with the auto-promote sidecar should survive a deliberate `pkill -TERM agentis` during a tick (manual smoke check)

Closes #728

### DIFF
--- a/research-foundry/tools/run-research.sh
+++ b/research-foundry/tools/run-research.sh
@@ -1085,6 +1085,11 @@ start_auto_promote_sidecar() {
     mkdir -p "$AP_LOG_DIR"
     date +%s > "$AP_STAMP"
     (
+        # Disable set -e/pipefail inherited from parent — supervisor loop must
+        # survive transient child SIGTERM (e.g. operator-side agentis upgrade
+        # pkill cascade, #728 Layer 1).
+        set +e
+        set +o pipefail
         cd "$LAPTOP_DIR"
         tick_count=0
         while :; do
--- a/tools/auto-promote-lock.py
+++ b/tools/auto-promote-lock.py
@@ -5,40 +5,150 @@ Extracted from tools/auto-promote.sh in #245 to replace the `flock -n` binary
 call (from util-linux, not present on stock macOS).
 
 The parent shell opens the lock file with `exec 200>"$LOCK_FILE"` and invokes
-this helper with the FD number. We call fcntl.flock(LOCK_EX | LOCK_NB) on that
-FD. POSIX flock locks are associated with the open file description, so the
-lock persists as long as any FD pointing to that OFD stays open. This helper
-exits after acquiring the lock; the parent shell keeps its FD open until the
-auto-promote run completes, which keeps the lock held without a supervisor
-process.
+this helper with the FD number plus the lock-file path. We call
+fcntl.flock(LOCK_EX | LOCK_NB) on that FD. POSIX flock locks are associated
+with the open file description, so the lock persists as long as any FD
+pointing to that OFD stays open. This helper exits after acquiring the lock;
+the parent shell keeps its FD open until the auto-promote run completes,
+which keeps the lock held without a supervisor process.
 
-Works on Linux (OFD-based flock) and macOS (BSD flock — same OFD semantics per
+Works on Linux (OFD-based flock) and macOS (BSD flock - same OFD semantics per
 flock(2): "This lock is removed on close(2) of the last file descriptor that
 references the open file").
 
-Exit 0 on successful acquire, 1 on contention, 2 on bad usage.
+On acquire, the helper writes its PID into the lock file. On contention, it
+reads the prior holder's PID and checks whether the process is alive
+(`os.kill(pid, 0)`) and -- on Linux -- whether `/proc/<pid>/cmdline` still
+references an auto-promote.sh or agentis process. If the holder is dead or
+its PID has been recycled into an unrelated process, the helper steals the
+lock instead of returning contention. This recovers the "lock orphaned by a
+spawned daemon FD that survived the parent shell" failure mode from #728.
+
+Exit 0 on successful acquire (including steal), 1 on real contention, 2 on
+bad usage.
 
 Args (positional):
-    1: fd — integer FD number inherited from the parent shell
+    1: fd        - integer FD number inherited from the parent shell
+    2: lock_path - filesystem path to the lock file (same path the parent
+                   shell opened with `exec 200>"$LOCK_FILE"`)
 """
+import errno
 import fcntl
+import os
 import sys
 
 
 def main():
-    if len(sys.argv) != 2:
-        sys.stderr.write('Usage: %s <fd>\n' % sys.argv[0])
+    if len(sys.argv) != 3:
+        sys.stderr.write('Usage: %s <fd> <lock_file_path>\n' % sys.argv[0])
         return 2
     try:
         fd = int(sys.argv[1])
     except ValueError:
         sys.stderr.write('Not an integer FD: %s\n' % sys.argv[1])
         return 2
+    lock_path = sys.argv[2]
     try:
         fcntl.flock(fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
-    except OSError:
-        return 1
+    except OSError as e:
+        if e.errno not in (errno.EAGAIN, errno.EWOULDBLOCK):
+            raise
+        # Real contention from the kernel's point of view. Probe whether the
+        # PID recorded in the lock file still belongs to a live auto-promote
+        # process; if not, steal.
+        if not try_steal_stale(fd, lock_path):
+            return 1
+    # Acquired (either cleanly or after a steal). Write our PID for future
+    # staleness checks.
+    write_pid_into_lock(fd)
     return 0
+
+
+def try_steal_stale(fd, lock_path):
+    """Read the lock-file PID; if the holder is dead or recycled, steal."""
+    try:
+        with open(lock_path, 'r') as f:
+            content = f.read().strip()
+        pid = int(content)
+    except (OSError, ValueError):
+        # No PID written or unparseable -- treat as stale and try to steal.
+        # This also covers the legacy case where a pre-#728 holder never wrote
+        # its PID into the file.
+        return retry_after_truncate(fd, lock_path)
+    if is_holder_dead(pid):
+        sys.stderr.write(
+            '[lock] stale lock detected (PID %d dead/recycled); stealing\n'
+            % pid
+        )
+        return retry_after_truncate(fd, lock_path)
+    return False
+
+
+def is_holder_dead(pid):
+    """Return True if the PID is dead or has been recycled into something
+    unrelated to auto-promote.sh / agentis.
+    """
+    try:
+        os.kill(pid, 0)
+    except ProcessLookupError:
+        return True
+    except PermissionError:
+        # PID exists but owned by another user. In our deployment the lock
+        # file is per-user, so this implies a recycled PID owned by some
+        # system process -- treat as recycled (steal-safe).
+        return True
+    # PID is alive. On Linux, check /proc/<pid>/cmdline to confirm it's
+    # actually an auto-promote.sh or agentis process. If the PID has been
+    # recycled into e.g. gnome-shell, treat as dead.
+    try:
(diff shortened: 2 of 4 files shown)


ylohnitram:
## QA Report

| Layer 1: set +e + set +o pipefail in subshell | ✅ | none | `run-research.sh:1087-1093` — both flags first thing inside `( ... ) &`, comment references `#728 Layer 1` |
| Layer 2: 200>&- on agentis daemon list (L199) | ✅ | none | `auto-promote.sh:199` — single-line, fd 200 closed before `\|\| echo "[]"` |
| Layer 2: 200>&- on agentis daemon stop (L363) | ✅ | none | `auto-promote.sh:363` — single-line, inside `(cd "$FED_DIR" && ...)` subshell |
| Layer 2: 200>&- on agentis daemon spawn (L375) | ✅ | none | `auto-promote.sh:379` — last continuation line before `&`, before backgrounding |
| Layer 2: lock helper takes 2 args | ✅ | none | `auto-promote.sh:133` — `python3 ... auto-promote-lock.py 200 "$LOCK_FILE"` |
| Layer 3: helper takes (fd, lock_path) | ✅ | none | `auto-promote-lock.py:42-50` — argv length check + int parse, `lock_path = sys.argv[2]` |
| Layer 3: on acquire, writes own PID + fsync | ✅ | none | `auto-promote-lock.py:135-151` — ftruncate/lseek/write `getpid()+\n` + fsync (graceful OSError) |
| Layer 3: on contention, reads PID + os.kill(0) check | ✅ | none | `auto-promote-lock.py:67-99` — `try_steal_stale` reads file, parses PID, `os.kill(pid, 0)` |
| Layer 3: /proc cmdline staleness check | ✅ | none | `auto-promote-lock.py:103-114` — wraps `/proc/<pid>/cmdline` in try/except OSError; non-Linux falls through to "alive" |
| Layer 3: stale-steal log on stderr | ✅ | none | `auto-promote-lock.py:79-82` — `sys.stderr.write('[lock] stale lock detected ...')` |
| PID 1 test design (vs $$) is sound | ✅ | none | Test 11c gates on reading `/proc/1/cmdline` and pattern-matches `*auto-promote.sh*\|*agentis*` to skip cleanly in exotic envs; production cmdline (`/sbin/init` or systemd) cannot match those substrings |
| 3 new tests pass | ✅ | none | `test-auto-promote.sh`: 40 passed / 0 failed — 11b dead-PID steal, 11c recycled-PID steal, 11d real-contention-still-blocks all green |
| colony-lint clean | ✅ | none | Local run: 344 passed / 0 failed / 4 skipped |
| bash -n syntax | ✅ | none | `run-research.sh` + `auto-promote.sh` both clean |
| python3 ast.parse | ✅ | none | `auto-promote-lock.py` parses clean |
| PR hygiene + no internal paths | ✅ | none | Title is conventional-commits; body has `Closes #728`; Co-Authored-By trailer present; no `/home/...` or `nitramyloh` leakage in diff or PR body |
| macOS compat (no /proc fallback) | ✅ | none | `/proc/<pid>/cmdline` read wrapped in `try: ... except OSError` — falls through to `return False` (treat-as-alive) on macOS, with comment documenting the trade-off |
| **CI status: CI green** | ❌ | **CI shellcheck job fails — SC2002 "useless cat" on `test-auto-promote.sh:423` + `:464`** | Must be fixed before merge. CI runs shellcheck 0.9.0 which flags SC2002 as warning-class; local shellcheck 0.11.0 relaxed it. Trivial mechanical fix: `cat "$F" 2>/dev/null \| tr -d '[:space:]'` → `tr -d '[:space:]' < "$F" 2>/dev/null`. Existing line 443 in the same file already uses the correct form (`tr -d '\0' < "/proc/1/cmdline"`), so the pattern is established. |

## Verdict: ⚠️ ship-with-followups

Three-layer design is correct: Layer 1 (sidecar resilience), Layer 2 (close-on-exec via `200>&-`), Layer 3 (stale-holder detection + PID writeback + Linux `/proc` cmdline check) all verify clean. macOS portability handled. 3 new tests cover dead-PID steal, recycled-PID steal, and live-contention-still-blocks; PID 1 design is sound (helper's cmdline pattern guard plus test-side skip on exotic init). 40/0 local test suite, 344/0 colony-lint, no internal-path leakage.

**Single blocker before merge**: CI shellcheck red. Two SC2002 "useless cat" warnings introduced by the new test code (lines 423 and 464 of `test-auto-promote.sh`). Mechanical 2-line fix; the same file already uses the correct `tr -d ... < "$file"` form on line 443. Local 344/0 colony-lint was misleading because the local shellcheck (0.11.0) no longer flags SC2002 while CI ships shellcheck 0.9.0 which still does.

After that fix lands and CI turns green, ship-it.

ylohnitram:
## QA update — inline fix applied

Both `cat <file> | tr -d '[:space:]'` SC2002 violations at `tools/test-auto-promote.sh:423` + `:464` rewritten to `tr -d '[:space:]' < <file>` matching the SC2002-clean pattern already used at line 443. Local 40/0 retained. Verdict upgraded to **✅ ship-it**.

PR #656, opened 2026-05-18
feat(phase-7-pra): auto-evolve plumbing + ledger + dry-run stubs (PR-A of #628)

ylohnitram:
## Summary

Phase 7 PR-A — plumbing for the self-improving `.ag` mutation harness
(#628). Zero `.ag` files mutated; the mutator is stubbed (PR-B fills
it in); the A/B step is stubbed (PR-B); `dry_run` ships at `true` so
even after `mutation.enabled` is flipped on later, no file mutations
happen until PR-C explicitly toggles it.

- Three new `CFG_EVOLVE_MUTATION_*`, four new `CFG_EVOLVE_AB_*`, and
  three new `CFG_EVOLVE_{ARCHIVE_DIR,LEDGER_PATH,DRY_RUN}` knobs
  threaded through `auto-promote-config-parser.py` with legacy-safe
  defaults so dev-apprenticeship and pre-#628 federations keep their
  existing `agentis evolve` path.
- `auto-promote.sh` evolve handler routes to the new harness when
  `evolve.mutation.enabled=true`, else falls through to the legacy
  path verbatim.
- `auto-promote-decisions.py` emits three new fields on every evolve
  decision: `evolve_window_stagnant`, `parent_sha`, `generation_current`.
  Test 12 byte-identity preserved (the new helpers fire only on the
  evolve emit path, which test 12's fixture never reaches).
- `auto-evolve-ab.sh` (new) — five-step pipeline (throttle → stub
  mutator → validity gate → A/B stub → cleanup) with a ledger row
  per outcome.
- `auto-evolve-ab-ledger.py` (new) — single-row JSONL writer with
  required base schema `{ts, event, agent, colony, generation,
  parent_sha, parent_sha8, ab_ticks, dry_run}` plus arbitrary extras.
- `colony-lint.sh` adds `-maxdepth 1` to the per-agent `.ag` glob so
  in-flight candidates under `.evolve/` don't fail tier-coverage.
- `test-auto-evolve-ab.sh` (new) — 11 smoke-test assertions.

Out of scope (deferred):
- Real LLM mutator (PR-B `auto-evolve-mutate.py`)
- A/B daemon spawn + score comparison (PR-B)
- Flip `evolve.dry_run=false` (PR-C)

## Test plan

- [x] `tools/colony-lint.sh`: 320 passed, 0 failed (baseline 319 + new smoke test)
- [x] `tools/test-auto-promote.sh`: 35 passed, 0 failed (test 12 byte-identity preserved)
[...]

PR #655, opened 2026-05-18
feat(phase-4-prb): prior_advocate colony for claim-auditor synthesis (PR-B of #625)

ylohnitram:
## Summary

Implements PR-B of three critic colonies per Phase 4 plan
(https://github.com/Replikanti/agentis-colonies/issues/625#issuecomment-4480642376).

The prior_advocate colony sits alongside the four web searchers
(arxiv / oeis / groupprops / scholar) and runs an adversarial-
reviewer LLM call that argues the triggering claim is already known
(cites the closest theorem / lemma / identity it can find). The
auditor folds the verdict into its synthesis ctx as an additional
KNOWN_PRIOR signal alongside the four web-search reports.

Failure-mode default: pass-through. Empty prior_advocate memo (LLM
unavailable or colony down) lets the auditor proceed on the four
searcher reports alone, so the new critic cannot strand the audit
pipeline on infrastructure issues. PR-A (skeptic) already merged via
#652; PR-C (reviewer) follows.

Two commits:
- `c25184d` — colony scaffold (4 files: `prior_advocate.ag`,
  `colony.example.toml`, `start-colony.sh`, `README.md`). Same
  `cb 2000`, tick interval, and 4-tier branch shape as the existing
  searchers; `upstream_tick = tick_idx - 1` mirrors the searcher
  offset. ADR-0003-conformant `start-colony.sh` with `--restart-agent`
  + `--rate-limit-status` flags. Tier branches from PR-2 (#632)
  preserved (autonomous collapses to review-gated; non-terminal).
- `65be9f9` — auditor + orchestrator wiring. `auditor.ag` Verdict
  struct gains `evidence_prior_advocate`; seed prompt instructs the
  LLM to count a strong prior_advocate match as a KNOWN_PRIOR
  signal; `_publish_auditor` persists
  `claim:report_prior_advocate:tick-N` alongside the four searcher
  reports on `VERIFIED_NEW`. `run-research.sh` bootstrap colony
  count 16 → 17 (cp/seed loops + dedicated daemon spawn line).
  `tools/check-learn-tags.sh` schema extended with
  `prior_match:partial` + `prior_match:success` pairs. CHANGELOG
  + BUNDLE.manifest updated.

## Test plan

- [x] `tools/colony-lint.sh` — 326 passed, 0 failed, 3 skipped
  (was 319 post-PR-A; +7 from new colony, plan expected ~326).
- [x] `tools/test-check-learn-tags.sh` — 16/0.
[...]

Issue #17, opened 2026-04-10
verify shellcheck is enforced in CI, not silently skipped


ylohnitram:
## Problem

`tools/colony-lint.sh` skips shellcheck when it is not installed locally:

```
[SKIP] dev-apprenticeship/planning: shellcheck not installed
```

The QA report on #8 noted this and flagged it as "CI must enforce it." It is not currently clear that the CI workflow actually runs shellcheck against the colony scripts — the PR body claim "Shellcheck clean on both scripts" was never verified by CI logs during QA review.

## Impact

If CI also runs `colony-lint.sh` on a host without shellcheck, the skip is silent there too, and shellcheck regressions can land unnoticed.

## Scope

- `.github/workflows/*.yml` — verify that shellcheck is installed on the runner and actively invoked against all `dev-apprenticeship/*/scripts/*.sh` files.
- `tools/colony-lint.sh` — optionally promote the skip to a hard failure when an environment variable like `COLONY_LINT_STRICT=1` is set, so CI can opt in.

## Verification checklist

- [ ] CI workflow installs shellcheck on the runner
- [ ] CI workflow runs shellcheck against every `dev-apprenticeship/*/scripts/*.sh`
- [ ] CI fails the job if shellcheck reports any error
- [ ] `tools/colony-lint.sh` documents that the local skip is local-only

## Related

Reported by QA on #8 as NIT-10.

ylohnitram:
## Decision

CI already installs shellcheck and invokes `colony-lint.sh`, and `colony-lint.sh` already fails the run on shellcheck errors. The only gap is that this is not visible to a reader. The fix is a two-part documentation and guardrail pass: (a) add a one-line comment in `colony-lint.sh` asserting that the shellcheck block is a hard requirement in CI, and (b) add a CI step name and (optionally) a README badge so the enforcement is discoverable. No new CI logic needed.

## Scope

Verification already done:

- `.github/workflows/ci.yml` installs shellcheck via `sudo apt-get install -y shellcheck` (line 16) and then runs `./tools/colony-lint.sh` (line 19).
- `tools/colony-lint.sh` lines 138-148 run `shellcheck` over every colony `*.sh`, increment `FAIL` on error, and the script exits non-zero at line 265 when `FAIL != 0`. The `skip "prefix: shellcheck not installed"` branch only fires when the binary is absent, which cannot happen in CI because of the install step.

So the plan is **not** to change CI logic. The plan is to:

[...]

PR #734, opened 2026-05-22
refactor(auto-promote): switch from stop+spawn to daemon reload (#733)

### LINKED ISSUE #733: auto-promote daemon respawn breaks pipeline cascade continuity — architectural question on promote-as-restart
## Problem

Run #16 forensic exposed an unexpected interaction between auto-promote and pipeline-cascade federations. When `auto-promote.sh` decides to promote an agent (e.g. `noticer` 0.7 → 0.8), it does:

```
agentis memo set <agent>:confidence <new_value>
agentis daemon stop <id>
agentis daemon <agent.ag> [flags]
```

The `stop` + `daemon` sequence kills the running daemon and spawns a fresh one with a new PID/instance_id. The new daemon starts with empty experience (the experience store is per-instance), empty per-tick state, and no in-flight LLM responses.

For pipeline-cascade federations (research-foundry, dev-apprenticeship), this is structurally hostile:

| Federation | Cascade depth | Effect of mid-cascade respawn |
|---|---|---|
| research-foundry | explorer → noticer → skeptic → formulator → verifier → novelty → 4 search → prior_advocate → auditor → editor → submitter (10+ ticks) | killing noticer mid-cascade orphans all downstream pending verdicts that referenced its previous instance |
| dev-apprenticeship | code_writer → test_writer → refactorer → commit_composer (4 ticks) | less catastrophic but still loses context |
| tribes-bench | flat (5 tribes hunt independently) | no cascade — respawn is harmless |

## Reproducer — Run #16 evidence

Run #16 was the first run where the auto-promote sidecar (post-#728 resilience fix) survived 75 ticks. It produced exactly **2 PROMOTE events** — both on `noticer` (0.7 → 0.8 at 13:34Z, 0.8 → 0.95 at 13:39Z). Each promote triggered a daemon respawn.

Result vs Run #14 (zero promotes, no respawns):

| Metric | Run #14 (no promotes) | Run #16 (2 noticer promotes) |
|---|---|---|
| Ticks | 30 | 75 |
| Discovery claims | 136 | 295 |
| PDF preprints | **6** | **1** |
| Auditor verdicts written | 0 | 1 (only tick 48 — almost nothing got that far) |

Run #16 had 2.5× the tick budget yet produced 1/6 of the preprints. The 295 discovery claims are explorer-side — those don't depend on noticer continuity. But everything past noticer in the cascade is suspect: the auditor only managed ONE successful verdict in 75 ticks, vs. Run #14 where the pipeline cascade produced 6 PDFs without any respawn interruption.

The noticer respawn at tick ~35 (and again at tick ~37) appears to have broken the cascade window between explorer-tick-N and auditor-tick-N+10 for every claim in flight at that moment.

## The architectural question

This is not a bug to fix with a one-line patch. There are at least four design responses, and the "right" one depends on what we believe `promote` semantically means in a pipeline federation:

### Option A — Memo-only promote (no restart)

Sidecar updates the confidence memo and stops there. The daemon's `tier()` builtin reads `<agent>:confidence` on every call, so the next tick naturally picks up the new value. No respawn.

**Pros:** Zero downtime, zero state loss, zero cascade interruption.
**Cons:** Daemon's pre-cached state (CB pool, learning history, in-tick budgets) carries from the old tier into the new. If the old tier had been accumulating "shadow-tier learning" (observe-only), that state continues to weigh decisions in the new tier. Whether this is a feature or a bug depends on how much "tier-specific state" we believe should reset.

### Option B — Graceful drain before respawn

Sidecar writes a `<id>.stop_after_tick` signal, daemon completes its current tick (and any in-flight emit/memo writes), THEN sidecar swaps memo + respawns. Net effect: one tick interval of brief downtime (60-120s in research-foundry), but no mid-tick state loss.

**Pros:** Preserves cascade continuity for any tick that started before the promote decision. Respawn-state-reset semantic stays intact (the new tier starts fresh).
**Cons:** Adds a tick of latency per promote. Requires new `agentis daemon stop --graceful` flag (or similar primitive in core) + sidecar orchestration. Doesn't help cross-tick cascade (claims that needed N+5 ticks downstream).

### Option C — `agentis daemon reload` primitive (in-process tier refresh)

New agentis-core builtin: `agentis daemon reload <id> --memo-keys confidence`. The daemon's tick loop polls a reload flag every tick; when set, it re-reads the listed memo keys and updates its `tier()` cache without dropping any in-process state. No respawn, no stop.

**Pros:** Zero downtime, preserved continuity, AND tier-cache is explicit (no implicit pre-cache carry from Option A).
**Cons:** Requires agentis-core PR (new CLI subcommand + tick-loop hook). Larger surface than A or B. But it's the substrate that makes "promote-without-restart" a first-class operation everywhere, not a federation-specific hack.

### Option D — Promote means "fresh start" by design (current behavior, document it)

The `agentis daemon stop` + `daemon` sequence is intentional: a promotion changes the agent's tier-allowed behaviors (per ADR-0001), and a fresh daemon is the cleanest enforcement boundary. The pipeline-cascade disruption is a feature, not a bug — claims that were in flight under the old tier are no longer valid under the new tier's writing privileges.

**Pros:** Simplest semantics. The respawn IS the promotion. No new primitives.
**Cons:** Hostile to long-cascade federations. Either operators accept the cascade disruption OR they avoid mid-run promotes (only promote between distinct "research runs", not during one). Means research-foundry can never realistically promote past the initial seed without sacrificing throughput.

## Acceptance criteria

This issue is for **architecture discussion**, not immediate implementation. A "fix" here is the design decision. Concretely:

- [ ] Decide which option (A / B / C / D / hybrid) aligns with the long-term design intent of agentis + agentis-colonies.
- [ ] If the answer is A/B/C, file the implementation issue(s) — agentis-core for option C, agentis-colonies for options A/B.
- [ ] If the answer is D, update `doc/auto-promote.md` and ADR-0001 to make the cascade-disruption explicit, and define the operator workflow (e.g. "schedule promotes between runs, not during runs").
- [ ] Either way, the docstring in `tools/auto-promote.sh` around the `agentis daemon stop` + spawn block needs the architectural choice referenced inline so future contributors don't trip over the same surprise.

## Why this is bigger than "fix the respawn"

The operator wants the federation to behave well "long-term, not just for the next run." The current behavior breaks an invariant that pipeline-cascade federations depend on: **the agent identity is continuous within a run**. A respawn invalidates that invariant.

But the *reason* the code respawns is that the developer at #245 / #148 era assumed promotes are sparse and the daemon needs a clean-slate restart to honor the new tier semantics. That assumption was correct for `dev-apprenticeship` Stage 1 (sparse promotes, simple cascades) and wrong for `research-foundry` (frequent promotes desirable, deep cascades).

We can either accept the assumption mismatch and document it (Option D), or invest in a substrate change that makes the assumption obsolete (Option C). The middle options (A, B) are incremental compromises.

## Related

- #728 — sidecar resilience (sidecar didn't even survive the FIRST promote tick before this PR)
- #719 — per-explorer topic pinning (Run #16 showed pinning works — diverse subtopics — but downstream cascade is the bottleneck)
- ADR-0001 — confidence tiers (normative semantics for what each tier is allowed to do — the source of the "promote = behavior change" framing)
- agentis-core #652 — per-tier LLM backend routing (already implements memo-only re-resolution at every prompt() call — precedent for Option A/C approach)

## Surfaced from

Run #16 (75 ticks, 295 claims, 1 PDF) vs Run #14 (30 ticks, 136 claims, 6 PDFs). Same session, this conversation. The 2 promote events succeeded technically but cost an order-of-magnitude in pipeline throughput.

ylohnitram:
## Summary

- Replace `agentis daemon stop <id>` + `agentis daemon <file>` respawn with `agentis daemon reload <id>` in the auto-promote promote path
- Add startup version check requiring agentis-core >= 1.7.16 (hard cutover, per #733 decision)
- Drop the now-unused TOML `[gitlab]` parsing + tick-interval resolution + spawn-env composition from the promote branch (evolve branch unchanged)

## Why

Issue #733 — Run #16 forensic showed promote-via-respawn breaks pipeline cascade continuity in long-cascade federations. Run #14 (no promotes) produced 6 PDFs from 136 discovery claims; Run #16 (2 promotes) produced 1 PDF from 295 claims. The new `agentis daemon reload <id>` primitive (agentis-core #655, shipped in v1.7.16 via #656) keeps the daemon in-process, preserving the experience store, CB pool, in-flight LLM cache, and listen() subscriptions. The lifecycle event `agent.tier_reloaded` provides the audit trail the old `agent.respawned` event used to carry.

Hard cutover, not detect+fallback: pre-v1.7.16 binaries lack the `reload` subcommand and would fail silently mid-promote (the cascade-disruption class of bug #733 targeted). The version check exits 1 with a clear message instead of degrading.

LOC: 83 added, 111 removed (net -28).

## Test plan

- [x] `bash -n tools/auto-promote.sh` clean
- [x] `tools/test-auto-promote.sh` — 40/40 pass (heredoc invariant still holds)
- [x] `tools/test-auto-promote-install.sh` — 13/13 pass
- [x] `tools/colony-lint.sh` — 344 passed, 0 failed, 4 skipped
- [x] Manual awk version-check verification against valid (>=1.7.16) and invalid (1.7.15, 1.4.10, empty, garbage) version strings
- [ ] CI green
- [ ] After merge: `research-foundry/tools/Containerfile.research` bump `AGENTIS_VERSION` v1.7.15 -> v1.7.16, rebuild image, spawn Run #17 — expected: promotes happen WITHOUT cascade disruption, preprint count matches or exceeds Run #14's 6

Closes #733

### DIFF
--- a/tools/auto-promote.sh
+++ b/tools/auto-promote.sh
@@ -34,18 +34,49 @@
 #            --config tools/auto-promote-config.research-foundry.yaml
 #
 # --containerized opts out of the dev-apprenticeship-specific
-# `[gitlab]`-in-colony.toml respawn path; the forge-less
+# `[gitlab]`-in-colony.toml resolution path; the forge-less
 # `research-foundry/` federation needs this because its 18 colonies
-# have `forge.type = "none"` and the daemon must be respawned without
-# GITLAB_* env composition (#622, #638).
+# have `forge.type = "none"` (#622, #638). Note that since #733 the
+# promote path no longer respawns the daemon at all — it calls
+# `agentis daemon reload <id>` instead — so neither mode has to
+# reconstruct the GITLAB_* env at promote time.
 #
 # --config overrides the default tools/auto-promote-config.yaml path.
 #
-# Prerequisites: agentis, python3 (with PyYAML or a fallback parser)
+# Prerequisites: agentis-core >= 1.7.16 (daemon reload primitive, #733 /
+# agentis-core #655), python3 (with PyYAML or a fallback parser).
 #
 # Layer 1 of the auto-governance roadmap (#148). See README for layers 2+3.
 
 set -euo pipefail
+
+# --- Runtime floor: agentis-core >= 1.7.16 ---
+# Promote path uses `agentis daemon reload <id>` (agentis-core #655, shipped
+# in v1.7.16), not the legacy `agentis daemon stop` + `agentis daemon <file>`
+# respawn pattern. Hard cutover: older binaries lack the reload subcommand
+# and would fail silently mid-promote, which is exactly the cascade-disruption
+# class of bug #733 (agentis-colonies) set out to eliminate. Fail loudly and
+# early instead of degrading.
+AGENTIS_VERSION_OUTPUT="$(agentis --version 2>/dev/null || true)"
+if [ -z "$AGENTIS_VERSION_OUTPUT" ] || ! printf '%s\n' "$AGENTIS_VERSION_OUTPUT" | awk '
+    NR == 1 {
+        if (NF < 2) exit 1;
+        v = $2;
+        sub(/^v/, "", v);
+        split(v, parts, ".");
+        major = parts[1] + 0;
+        minor = parts[2] + 0;
+        patch = parts[3] + 0;
+        # Require >= 1.7.16
+        if (major < 1) exit 1;
+        if (major == 1 && minor < 7) exit 1;
+        if (major == 1 && minor == 7 && patch < 16) exit 1;
+        exit 0;
+    }
+'; then
+    echo "ERROR: auto-promote.sh requires agentis-core v1.7.16+ for the daemon reload primitive (issue #733 / agentis-core #655). Detected: ${AGENTIS_VERSION_OUTPUT:-<empty>}" >&2
+    exit 1
+fi
 
 # --- Path resolution ---
 
@@ -247,7 +278,7 @@ while IFS='|' read -r decision agent colony step_from step_to reason evidence_js
             log "PROMOTE $agent ($colony): $step_from -> $step_to"
             if [ "$DRY_RUN" = "true" ]; then
                 log "  [dry-run] Would run: agentis memo set ${agent}:confidence $step_to"
-                log "  [dry-run] Would restart daemon for $agent"
+                log "  [dry-run] Would reload daemon for $agent (agentis daemon reload, #733)"
                 journal_append "$agent" "promote" \
                     "$(python3 -c "import json,sys; e=json.loads(sys.argv[1]); e['action']='dry-run'; print(json.dumps(e))" "$evidence_json")" \
                     "$step_from" "$step_to"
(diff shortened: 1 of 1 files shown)


ylohnitram:
## QA Report

| Version check requires agentis-core >= 1.7.16 | ✅ | none | awk logic empirically verified against 11 cases — empty/`agentis`-only/garbage/v1.7.15/v0.9.0/v1.6.99 → FAIL; v1.7.16/v1.7.17/v1.8.0/v2.0.0/bare `1.7.16` → PASS. `NF<2` guards short input, `sub(/^v/,...)` normalizes prefix, three layered comparisons enforce floor |
| Reload replaces stop+spawn in PROMOTE block | ✅ | none | `agentis daemon reload "$AGENT_ID"` at line 347 inside `(cd "$FED_DIR" && ... 200>&-)` subshell — single primitive, no respawn |
| Legacy `[gitlab]` TOML parse removed | ✅ | none | zero refs in code (only one comment mention in script header explaining what `--containerized` used to do) |
| Legacy SPAWN_ENV eval removed | ✅ | none | `grep -n SPAWN_ENV tools/auto-promote.sh` returns zero matches |
| Legacy resolve-tick-interval.py invocation removed | ✅ | none | `grep -n resolve-tick-interval tools/auto-promote.sh` returns zero matches |
| Legacy respawn `(...) &` block removed | ✅ | none | only daemon CLI calls remaining are `agentis daemon list` (line 230) + `agentis daemon reload` (line 347); no `agentis daemon stop`, no backgrounded respawn subshell |
| Journal entry uses mechanism=reload | ✅ | none | line 350: `e['mechanism']='reload'` on success; line 355: `e['warning']='reload_failed'` on partial |
| Dry-run log uses "reload" wording | ✅ | none | line 281: `Would reload daemon for $agent (agentis daemon reload, #733)` |
| Hard cutover (no legacy fallback) | ✅ | none | reload failure → journal warning + `continue` to next loop iteration; zero fallback to stop+spawn |
| Zero heredocs (test 10 invariant) | ✅ | none | `grep -c "<<"` returns 1, but the sole match is on line 211 inside a `#` comment; test 10 (which excludes comments via `grep -vE '^[[:space:]]*#'`) reports `[PASS] auto-promote.sh has no heredocs (#245)` |
| bash -n syntax | ✅ | none | clean |
| test-auto-promote 40/0 | ✅ | none | 40 passed, 0 failed |
| test-auto-promote-install 13/0 | ✅ | none | 13 passed, 0 failed |
| colony-lint 344/0 | ✅ | none | 344 passed, 0 failed, 4 skipped |
| Single-file change scope | ✅ | none | `git diff main --stat` shows only `tools/auto-promote.sh` (83 +, 111 -, net -28) |
| PR hygiene + no internal paths | ✅ | none | title matches `refactor(auto-promote): switch from stop+spawn to daemon reload (#733)`, body contains `Closes #733`, no internal user/worktree paths in script or PR body |
| Containerfile bump correctly out-of-scope | ✅ | none | PR body Test plan explicitly lists the Containerfile.research bump as a post-merge follow-up step |

## Verdict: ✅ ship-it

PR #794, opened 2026-05-25
research-foundry: map federation outcome vocab to runtime enum in learn()

ylohnitram:
## Summary

Three `.ag` files in `research-foundry/` passed federation-domain vocabulary strings into the `outcome` slot of `learn()`, where the runtime expects values from the enum `{success, failure, partial, timeout, error}`. The runtime rejected each call with `[tick:error] invalid outcome 'verified': expected success/failure/partial/timeout/error` (and analogous lines for `incomplete`, `failed`, `cache_hit`, `rejected`) and counted the tick as errored.

The most visible effect: the **theorist daemon listed at 20% success rate** (`agentis daemon list`) during the post-v1.7.17 V4-Pro 75-tick #778 re-run, with `17+ [tick:error] invalid outcome 'verified'` lines per Lean-verified tick. The buyer-side bug in `explorer.ag` and `prior_advocate.ag` was latent — only fires when `knowledge_buy` returns `cognitive.cache_hit` or `cognitive.rejected`, and in healthy runs with active sellers most calls land on silent 'succeeded' / 'fallback' paths.

## Change

Move the federation-domain vocabulary into the `recommendation` slot (preserving log readability + `check-learn-tags.sh` disambiguation), keep the runtime-valid enum in the `outcome` slot:

| Site                        | Old outcome | New outcome | Recommendation now holds |
|-----------------------------|-------------|-------------|--------------------------|
| `theorist.ag:539`           | `verified`  | `success`   | `verified`               |
| `theorist.ag:542`           | `incomplete`| `partial`   | `incomplete`             |
| `theorist.ag:545`           | `failed`    | `failure`   | `failed`                 |
| `explorer.ag:793`           | `cache_hit` | `success`   | `cache_hit`              |
| `explorer.ag:797`           | `rejected`  | `failure`   | `rejected`               |
| `prior_advocate.ag:353`     | `cache_hit` | `success`   | `cache_hit`              |
| `prior_advocate.ag:357`     | `rejected`  | `failure`   | `rejected`               |

[...]

PR #778, opened 2026-05-24
fix(research-foundry): calibrate auditor + reviewer prompts — completes prompt-shaping sweep (#777)

ylohnitram:
**Draft until A/B evidence from Run #19 attempt 5 (Qwen3-Coder-30B via OpenRouter) lands.**

## Summary

Two-file prompt rewrite. Closes #777 and completes the calibration sweep for all five LLM-driven gate-bearing colonies in research-foundry (skeptic via #773, noticer + novelty via #775, auditor + reviewer here).

## Problem

Run #19 attempt 4 (post-#775) cascade analysis surfaced the next two gates that hadn't yet been calibrated:

- **Auditor**: 9 experience rows but **0 audit-ledger entries**. The audit-ledger only receives writes when the auditor produces a verdict; with the strict-anchor prompt, decisions biased toward `KNOWN_PRIOR` or `NEEDS_HUMAN` over `VERIFIED_NEW`, so no published audit verdicts.
- **Reviewer**: 7 experience rows, only 2 ctx writes, **0 actual approval outputs**. Same pattern — the strict-anchor prompt rejects on minor formatting differences between .tex claims and reproducibility stdout.

Both prompts opened with `"You are a strict X"` — identical anchor to pre-fix skeptic/noticer/novelty.

## Fix

Same template as #773 / #775 applied to both prompts:

| Element | Before | After |
|---------|--------|-------|
| Anchor | "strict X" | "CALIBRATED X — your job is to triage, not gatekeep" |
| Cost asymmetry | (implicit) | explicit: auditor — "false negatives drop the research / false positives waste reviewer compute"; reviewer — "false negatives block the submitter / false positives push hallucinations to publication" |
| Default when uncertain | restrictive label | permissive: `NEEDS_HUMAN` (auditor) / `approved` (reviewer) |
| Negative criterion | "strict" gating | "ONLY for near-EXACT direct match" (auditor `KNOWN_PRIOR`); "ONLY when central claim has no stdout support whatsoever, peripheral motivating remarks excluded" (reviewer `rejected`) |
| Positive criterion | conservative | "novel structural angle, even if touches classical territory" (auditor `VERIFIED_NEW`) |

Verdict-label vocabularies preserved unchanged:
- auditor: `{VERIFIED_NEW, VERIFIED_BY_LEAN, KNOWN_PRIOR, NEEDS_HUMAN}`
- reviewer: `{approved, rejected}`

[...]

PR #752, opened 2026-05-23
feat(research-foundry): Lean4 verifier integration in theorist.ag (#745)

ylohnitram:
## Summary
- Adds Lean 4 toolchain (elan + pinned `leanprover/lean4:v4.13.0`) to `research-foundry/tools/Containerfile.research` so theorist.ag can formally check generated proofs at tick time. Image bloat ~1.5 GB (1.24 GB -> ~2.7 GB acceptable).
- theorist.ag emits `.lean` alongside the existing `.tex` proof body, runs `lean <file>` with a 30s `timeout` cap, and writes `theorist:<pid>:lean_verdict:tick-N` memo with `{verified, incomplete, failed}`. Non-terminal -- even on lean failure the LaTeX is still published.
- auditor.ag fans in the lean_verdict via `_pick_upstream_by_confidence("theorist", "lean_verdict", ...)` and extends its verdict-label enum to `{VERIFIED_NEW, VERIFIED_BY_LEAN, KNOWN_PRIOR, NEEDS_HUMAN}`.

## Why
Epic #738 child P2. Supersedes deferred #627. Today the auditor relies on literature search alone — no formal proof check. This PR wires Lean 4 (via elan) into the container so theorist.ag emits `.lean` alongside `.tex` and verifies the proof body. Most LLM-generated proofs will FAIL the formal check — that is the desired signal, not a bug. `incomplete` (proof contains `sorry`) vs `failed` (real error) vs `verified` (closed proof accepted) gives the auditor and the AdaptiveEngine a richer signal than the current literature-only verdict.

## Scope
- Containerfile.research: +13 LOC (elan install + PATH + smoke test)
- theorist.ag: +110 LOC (`LeanDraft` type, `_lean_prompt`, `_run_lean_check`, per-tick lean call, branched `learn(lean_check, ...)`, plumbing into `_publish_theorist`)
- auditor.ag: +12 LOC (verdict-label enum extension, lean_verdict fan-in, ctx line)
- check-learn-tags.sh: +24 LOC (schema for `lean_check:verified|incomplete|failed`)
- test-run-research.sh: +61 LOC (19 new assertions covering all three .ag/Containerfile/schema layers)

Total: ~220 LOC inserted across 5 files.

## Out of scope (deferred follow-up)
[...]

PR #795, opened 2026-05-25
research-foundry: tighten Lean verification — reject True := trivial certificates

ylohnitram:
## Summary

The `VERIFIED_BY_LEAN` auditor verdict label was supposed to certify "a Lean 4 formal proof of a novel statement" — the strongest signal in the federation's quality stack. In the post-#794 V4-Pro 75-tick #778 re-run it was emitting on `theorem T : True := trivial` — Lean correctly certifying that "True is True" rather than the upstream mathematical claim.

Root cause: PR #752 (Lean integration) had given the theorist a cascade-keeping fallback — when the LLM-rendered claim referenced an undefined symbol that stock Lean (no mathlib, per Containerfile image-size constraint) couldn't translate, the seed_prompt explicitly instructed `fall back to a Prop placeholder (theorem T : True := trivial)`. Lean accepts it. lean_verdict=verified. Auditor sees verified, emits VERIFIED_BY_LEAN. Indistinguishable from a real formal verification at the label level.

The safety net was worth the verdict-label collapse only as long as the broader pipeline was breaking elsewhere (theorist's outcome-enum bug per #794, etc.). With the infrastructure now solid, the quality bar can be raised.

## Three-layer fix

| Layer | Where | What |
|-------|-------|------|
| 1. Seed prompt | `theorist.ag::_lean_prompt()` | Removes the `True := trivial` fallback instruction. When a referenced symbol cannot be translated without mathlib, the LLM is now instructed to declare it as a `Prop` axiom (`variable (P : Prop) ; theorem T : P -> P := fun h => h`) — captures the structural shape and is genuinely verifiable, not vacuous. |
| 2. Publish-path regex guard | `theorist.ag::_run_lean_check()` | Any `lean_source` matching `: *True *:=` is force-downgraded from `verified` to `incomplete` before the verdict is written. Defense in depth against an LLM that ignores layer 1. |
| 3. Auditor proof-kind gate | `auditor.ag::_seed_prompt()` + ctx-builder | Auditor now fans in `theorist:<pid>:proof_kind:tick-N` alongside the Lean verdict. Seed prompt requires BOTH `lean_verdict='verified'` AND `proof_kind='full'` for `VERIFIED_BY_LEAN`. `computational` / `empty` proof_kind degrade to `VERIFIED_NEW` (if other novelty signal strong) or `NEEDS_HUMAN`. |

## Expected empirical impact

`VERIFIED_BY_LEAN` frequency will drop sharply on the next V4-Pro run — most LLM-rendered claims will not be formalizable without mathlib, surfacing as `failed` (Lean rejects) or `incomplete` (sorry placeholder). The surviving `VERIFIED_BY_LEAN` cases will represent **real formal verification** of the upstream claim rather than trivial-true placeholders.

In the partial run that surfaced the bug:

| Tick | lean_source                                                           | lean_verdict (pre) | lean_verdict (post) |
|------|------------------------------------------------------------------------|--------------------|---------------------|
| 7    | `-- Placeholder ... [...]` `theorem T : True := trivial`               | verified           | **incomplete** (regex guard fires) |
| 9    | `theorem T : True := trivial`                                          | verified           | **incomplete** (regex guard fires) |

Both formerly labeled `VERIFIED_BY_LEAN` by the auditor; under this PR neither would reach that label.

## Caveat — mathlib is the longer-term unlock

Without `import Mathlib`, the theorist has a very small Lean library to work with for real number theory / combinatorics / algebra claims. Most upstream claims will surface as `failed` (Lean can't even parse the symbols) or `incomplete` (sorry as the best honest answer). That is the desired signal — empty / failed Lean output is far better than vacuous-verified output. Baking mathlib into `Containerfile.research` is a follow-up: it would 3x the image size (~3 GB compiled) but unlock real formal verification of actual math claims.

## Test plan

- [x] `agentis test research-foundry/theorist/agents/theorist.ag` — parses clean.
- [x] `agentis test research-foundry/auditor/agents/auditor.ag` — parses clean.
- [x] `bash tools/check-learn-tags.sh` — 0 violations.
- [ ] CI: colony-lint.
- [ ] Next V4-Pro run: VERIFIED_BY_LEAN drops to 0 on `True := trivial` patterns; any survivors are inspected for `proof_kind=full` + non-trivial Prop.


ylohnitram:
## QA report — PR #795 (fix/theorist-real-lean-verification)

Scope reviewed: `git diff main..fix/theorist-real-lean-verification` — 3 files, +28/-2.

| Diff scope matches PR description | ✅ | — | 3 files (`research-foundry/CHANGELOG.md`, `auditor.ag`, `theorist.ag`), 28 insertions / 2 deletions. Matches the three-layer change. |
| Layer 1: `_lean_prompt` placeholder removal | ✅ | — | `True := trivial` fallback removed; replaced with `variable (P : Prop) ; theorem T : P -> P := fun h => h` axiom-shell pattern. Reads as an explicit, non-vacuous instruction. |
| Layer 2: regex `: *True *:=` coverage | ✅ | — | Confirmed by direct shell probe on all canonical patterns: `theorem T : True := trivial` MATCH, `theorem T : True := by trivial` MATCH, `theorem T:True:=trivial` (no spaces) MATCH, `theorem T :   True   :=   by trivial` (extra spaces) MATCH, multi-line `-- comment\ntheorem T : True := trivial` MATCH, `theorem U : Nat := 0` NO_MATCH, `theorem W : True := by exact True.intro` MATCH. |
| Layer 2: `True -> True := id` vacuous-arrow case | ⚠️ | Out of scope per PR; flag as follow-up if it surfaces in next run. | `theorem V : True -> True := id` does NOT match (`: True ->` not `: True :=`). Equally vacuous mathematically. PR doc note acknowledges this is unhandled. Acceptable for v1 — `True -> True` is not a documented LLM failure mode in #778 run. |
| Layer 2: docstring-comment false positive | ⚠️ | Cosmetic risk only — LLM lean_prompt now bans `True := trivial` discussion in the output. Low likelihood. | A Lean source whose body is real but contains a comment like `-- True := trivial is bad` would be force-downgraded. False-positive probability negligible given the deterministic translation prompt. |
| Layer 2: `shell_escape(lean_source)` injection safety | ✅ | — | `lean_source` runs through `shell_escape()` before concat into the grep cmd; same pattern as the pre-existing `has_sorry` check four lines above. `grep -E` POSIX-portable. `colony-lint: safe-exec-concat` annotation present. |
| Layer 3: auditor `_pick_upstream_by_confidence` signature | ✅ | — | `_pick_upstream_by_confidence(\"theorist\", \"proof_kind\", \"proof_kind\", \"\", upstream_tick)` matches the existing 5-arg signature (`role, ranking_key, output_key, confidence_field, tick`). Empty `confidence_field` is the same pattern as the sibling `lean_verdict` call on the preceding line. |
| Layer 3: theorist actually publishes `proof_kind:tick-N` | ✅ | — | `theorist.ag` line 251: `memo_write(\"theorist:\" + self_pid + \":proof_kind:tick-\" + to_string(upstream_tick), draft.proof_kind)`. Wired end-to-end. |
| Layer 3: auditor seed_prompt vocabulary mismatch with theorist emit | ⚠️ | Open a follow-up issue: align seed_prompt enum to actual emitted values (`full` / `sketch` / `computational`) and decide whether `sketch` should gate VERIFIED_BY_LEAN (probably no — sketch is not a closed proof body). | The seed_prompt advertises `proof_kind ∈ {full, computational, empty}`, but the theorist's `_draft_prompt` emits `proof_kind ∈ {computational, sketch, full}`. The real `sketch` value is never described to the LLM. `empty` is described but never emitted (the fan-in sets `empty` as the fallback for missing memos, which works, but a real `sketch` value will pass through verbatim into the auditor prompt and the LLM has no instruction for how to treat it). In practice the auditor's strict gate (\"ONLY when proof_kind reads 'full'\") will degrade `sketch` to VERIFIED_NEW, which is the right call — but the vocabulary doc is wrong. |
| Layer 3: prompt length / coherence | ✅ | — | Auditor seed_prompt grew ~600 chars but the additional clause is structured ("ONLY when X AND Y AND Z", "computational or empty proof_kind means..."). The instruction is non-contradictory: VERIFIED_BY_LEAN requires `lean_verdict='verified' AND proof_kind='full'`; the VERIFIED_NEW clause explicitly handles "Lean is 'verified' but proof_kind is not 'full'". Internally consistent. |
| CHANGELOG entry | ✅ | — | Under `[Unreleased] / ### Changed`. Accurate description of three layers + the #778 background + the PR #752 quality-bar-collapse context. |
| Public-repo discipline | ✅ | — | No internal paths, worktree refs, or private-repo mentions in diff, commit body, or PR description. |
| `colony-lint.sh` | ✅ | — | 344 passed, 0 failed, 5 skipped on local run. CI run on PR also green (`colony-lint` 3m10s pass). |
| `check-learn-tags.sh` | ✅ | — | Exit 0, no violations (no new `learn()` calls in this PR; existing branches unaffected). |
| Backward compatibility | ✅ | — | The `lean_verdict` vocabulary is unchanged from a downstream-consumer perspective (`verified / incomplete / failed / empty`). The `: True :=` regex routes those cases into `incomplete` — which downstream consumers already handle. `proof_kind` memo key was already being written by the theorist and is already consumed by `editor.ag` line 552 — no new memo schema. Adding a new consumer (auditor) is purely additive. |
| Non-regression: editor.ag still reads `proof_kind` | ✅ | — | `editor.ag:552` uses identical `_pick_upstream_by_confidence(\"theorist\", \"proof_kind\", \"proof_kind\", \"\", upstream_tick)` call. No semantic change, no contract break. |
| `agentis test` on both .ag files | ✅ | — | Both parse clean (0 failed). |

### Verdict

✅ **ship-it (with one follow-up)**

The three-layer fix is sound: the prompt-level removal, the regex defense-in-depth, and the auditor proof-kind gate compose into a meaningful quality bar for the `VERIFIED_BY_LEAN` label. The regex covers every realistic placeholder pattern probed; the `_pick_upstream_by_confidence` call signature matches; the theorist actually publishes the `proof_kind` memo that the auditor now reads; CHANGELOG is accurate; CI is green; no public-repo discipline leaks.

**Follow-up (non-blocking) — open as a separate issue:** the auditor `_seed_prompt`'s proof_kind vocabulary `{full, computational, empty}` does not match the theorist's actual emit set `{computational, sketch, full}`. The strict gate ("ONLY when proof_kind='full'") still produces the correct verdict for a `sketch` value (degrade to VERIFIED_NEW), so this is a documentation bug in the prompt rather than a logic bug. Worth fixing before the next run so the auditor LLM has accurate ground truth about what values it can see.

**Note (out of scope, no action):** the regex does not catch `theorem T : True -> True := id` or similar vacuous arrows. This is documented in the PR description and not a regression — previously such a statement would also have surfaced as `verified` and reached `VERIFIED_BY_LEAN` (the strict `proof_kind='full'` gate would now degrade it correctly anyway, since an `id`-shaped proof is rarely flagged `full` by the theorist).

PR #799, opened 2026-05-26
tools: rescale auto-promote thresholds + heartbeat-file liveness fallback (substrate exercise)

ylohnitram:
## Summary

The auto-promote sidecar has been a **no-op in research-foundry runs** for two independent reasons; the autonomous-tier substrate primitives (`replicate`, `evolve_self`, `knowledge_sell` pre-#792, auditor #761 audit-guard `decide()`, `_publish_*`'s autonomous-only ledger writes) therefore sat **unused in a 75-tick window** even though the federation infrastructure otherwise worked correctly.

Concrete evidence from the auto-promote journal of the post-#798 claude run (75-tick × 60s ≈ 96 min wall-clock):

```
[2026-05-25T23:47:05Z] auto-promote: SKIP submitter: memo last_check stale and effective_state='zombie' not running
[2026-05-25T23:47:05Z] auto-promote: SKIP noticer:    memo last_check stale and effective_state='zombie' not running
…  ← 16 more, one per colony, every sidecar tick
[2026-05-25T23:47:05Z] auto-promote: Done. Promotes=0, Evolves=0, Skips=18
```

Two distinct root causes; this PR fixes both.

## Layer 1 — thresholds rescaled for finite-run windows

`tools/auto-promote-config.yaml` thresholds were sized for multi-day production deployments only:

| Knob                                  | Pre-fix | Post-fix | Notes |
|---------------------------------------|---------|----------|-------|
| `min_entries`                         | 200     | **30**   | Total experience rows. |
| `min_acting_entries`                  | 60      | **10**   | Acting rows only — formula. |
| `min_runtime_hours`                   | 48      | **0.5**  | 48h gate alone ruled out any 75-tick × 60s run. |
| `delta_slope_window`                  | 100     | **20**   | Acting-row sample window for slope. |
| step 0.8→0.95 `min_acting_entries_override` | 120 | **30** | Tighter than global (10) → rule-of-three @ 10% reject rate preserved. |

The autonomous-step override stays **strictly tighter** than the global threshold so the rule-of-three safety margin for terminal writes is preserved at a 10% target reject-rate — still tighter than the global 5% prereq.

## Layer 2 — heartbeat-file fallback for containerized liveness

[...]

PR #798, opened 2026-05-25
research-foundry: unblock downstream cascade — reviewer schema + auditor/submitter CB

ylohnitram:
## Summary

Two distinct downstream-colony bugs surfaced during the post-#795 claude flat-tariff run; the cascade tail (auditor 0% rate, reviewer 5% rate, submitter 0% rate, see live `agentis daemon list` at tick ~60/75) was effectively dead by tick ~30 and never recovered.

## Bug 1 — reviewer `hallucinated_claims` string-vs-list type error

`_publish_reviewer` crashed **71×** during the run with:

```
[tick:error] fn "_publish_reviewer":
  type error: expected compatible types for Add, got string and list
```

The `Verdict.hallucinated_claims` field is declared `string` (since #623) but the `_seed_prompt` body asked the LLM for `[{claim_text, line_or_section, why_unsupported}]`. The agentis runtime treats `prompt() -> T` as loose-typed: it accepts whatever shape the LLM returns at the call site, but the field then carries that runtime shape downstream — so `verdict.hallucinated_claims` arrives as a `list` and crashes the string concatenation in the verdict_json builder at the next line.

Pre-claude LLMs (qwen, V4-Flash) returned plain strings even when the prompt asked for lists, hiding the bug. Claude (sonnet at propose tier) follows the prompt schema faithfully and returns the list, exposing the latent crash.

**Fix:** align the prompt to the declared type. `hallucinated_claims` is now requested as a plain newline-separated string with the structured fields collapsed to `claim_text | line_or_section | why_unsupported` per line, and an explicit note in the prompt that the runtime enforces `string` and a list value is fatal.

## Bug 2 — auditor + submitter CB exhaustion

`cb_budget = 2000` in `auditor/config/colony.example.toml` + `submitter/config/colony.example.toml` (with matching `cb 2000;` in the `.ag` files) was insufficient once the daemons reached autonomous confidence. At autonomous tier the auditor runs `prompt()` (50 CB) + `decide()` (50 CB, #761 audit-guard) + ~10 `learn()` rows (50 CB) + per-tick searcher report reads + the new `proof_kind` picker fan-in (#795). The per-tick burst exceeded the 2000 ceiling, draining the budget faster than `daemon.cb_per_tick = 100000` could replenish.

Live evidence (claude run, tick ~60/75, before this fix):

| Daemon    | TICK_OK / ERR | RATE | Error type |
|-----------|---------------|------|-----------|
[...]

PR #800, opened 2026-05-26
dashboard: heartbeat-file fallback for containerized PID liveness

ylohnitram:
## Summary

The dashboard showed **"18 daemons with dead PID"** during the post-#798 claude run while every daemon was alive and ticking inside the container — false-negative every refresh. Same architectural issue I shipped a fix for in #799 (auto-promote sidecar), now applied to the dashboard collector.

## Root cause

`federation-dashboard-collector.py`'s containerized liveness check (introduced in #683 to replace the host-only `os.kill(pid, 0)` probe) trusted `<agent>:last_check` memo freshness as its only signal. But the `.ag` convention `memo_write("<agent>:last_check", now)` at end-of-tick is unreliable — many tick bodies short-circuit before reaching the write site (missing-upstream picker-empty paths, tier branches that bail early, `exec sh` failures on the `date -u +%Y-...` helper).

**Concrete reproducer** from the post-#798 claude run:

| File           | Tick log entries | `last_check` memo entries | Coverage |
|----------------|------------------|---------------------------|----------|
| `noticer.ag`   | 118              | 20                        | **~17%** |

The dashboard saw stale memo, classified the daemon as dead, even though the runtime tick loop was healthy.

## Fix

Add a heartbeat-file fallback that fires only when the memo path returns false. Reads `<fed_dir>/.agentis/daemon/<agent_id>.heartbeat` line 1 — ms-since-epoch of the last completed tick, written by `write_heartbeat_ext` in `src/daemon/runner.rs` every tick **regardless of `.ag` branch taken**. Reuses the same staleness window (`role_ticks × tick_interval_ms`) as `last_check`.

## Empirical impact

Restarted the dashboard against the live claude run with the patched collector:

| Metric             | Pre-fix | Post-fix |
|--------------------|---------|----------|
| `pid_alive`        | 0 / 18  | **18 / 18** |
| Dashboard banner   | DEGRADED | HEALTHY  |

## Test plan

- [x] Python syntax check: `python3 -c "import ast; ast.parse(...)"` clean.
- [x] Live verification on the running claude federation: dashboard correctly shows 18/18 alive after restart with patched lib.
- [ ] CI: colony-lint.

## Backward compatibility

- **dev-apprenticeship**: host-mode daemons write `last_check` reliably (no container short-circuiting), so the memo path returns true and the heartbeat fallback never fires. Behaviour unchanged.
- **tribes-bench / research-foundry**: containerized; previously false-negative every refresh, now reliable.
- Heartbeat path is opt-in: requires `agent_id` to be set and the heartbeat file to exist. If either is missing the fallback silently no-ops and we keep the pre-fix behaviour (last_check + `effective_state == 'running'` cascade).

Mirrors the auto-promote sidecar fix in #799 for byte-identical semantics. No `federation-dashboard/VERSION` bump in this PR — operator decides on the release cadence.


ylohnitram:
## QA report — PR #800 (heartbeat-file fallback for containerized PID liveness)

Scope: 2 files, +38 / -0. Mirrors PR #799 sidecar fix (merged at `442d101`) applied to `federation-dashboard/lib/federation-dashboard-collector.py`.

|---|---|---|---|
| Code-level parity with #799 | ✅ | — | Byte-equivalent semantics: `hb_epoch = int(line1) / 1000.0`, `freshness.resolve_tick_interval_ms`, `freshness.staleness_ticks_for`, `except (OSError, IOError, ValueError)`, `line1.isdigit()` gate, `window_seconds = role_ticks * (tick_interval_ms / 1000.0)`. Path expression `os.path.join(fed_dir, '.agentis', 'daemon', str(rec['agent_id']) + '.heartbeat')` matches the sidecar's `str(agent_id) + '.heartbeat'`. |
| Heartbeat path semantics | ✅ | — | Verified on a live containerized daemon heartbeat file: line 1 is digit-only ms-since-epoch (`1779822930067`), line 2 = interval (`4800000`), line 3 = health string (`degraded`). The collector reads only line 1 (`_hbf.readline().strip()`) and validates `line1.isdigit()` — robust to extra lines and to truncation between writes. |
| Backward compatibility — dev-apprenticeship (host-mode) | ✅ | — | Host-mode daemons reach the `memo_write("<agent>:last_check", now)` write site reliably, so the memo path at L292–L313 sets `rec['pid_alive']=True` and the fallback `if not rec['pid_alive']` never enters. Behaviour unchanged. |
| Backward compatibility — empty `agent_id` | ✅ | — | `rec.get('agent_id')` short-circuits the heartbeat read entirely; legacy fixtures with `agent_id=''` (collector line 248 default) skip the new path cleanly. |
| Backward compatibility — missing / malformed heartbeat | ✅ | — | `OSError`/`IOError` (file missing or unreadable) and `ValueError` (line 1 not numeric after `int()`) are swallowed. `rec['pid_alive']` retains its memo-path value (typically `False`) — matches pre-fix behaviour exactly. |
| Existing test coverage | ⚠️ | Follow-up issue: extend `tools/test-dashboard-freshness-liveness.sh` with three heartbeat-fallback cases (memo missing + fresh heartbeat → alive; memo missing + stale heartbeat → dead; memo missing + malformed heartbeat → dead). Non-blocking — manual live verification + #799 sibling test coverage in `tools/test-auto-promote.sh` provides cross-cut assurance. | The existing `tools/test-dashboard-freshness-liveness.sh` 5-case sweep only drives the memo path (fresh / stale / missing / containerized-PID / env-widened window). All five still pass conceptually because they all set `<agent>:last_check` memos that drive `rec['pid_alive']` to a definitive value before the fallback gate. No heartbeat files exist in those fixtures, so the new code path's exception handler silently no-ops. The fallback path itself is presently untested in the repo. |
| `federation-dashboard/VERSION` unchanged | ✅ | — | VERSION stays at `0.9.2`. Diff `git diff main..origin/fix/... -- federation-dashboard/VERSION` returns 0 lines. CHANGELOG entry sits under `## [Unreleased]` (correct — operator decides release cadence). |
| Public-repo discipline | ✅ | — | No internal absolute paths, no worktree references, no private-repo mentions in diff, PR body, or CHANGELOG entry. |
| `colony-lint.sh` | ✅ | — | CI green: `colony-lint pass 3m19s` on PR head. |
| `os` import | ✅ | — | `os` imported at line 31 of `federation-dashboard-collector.py` (`import sys, os, json, time, re, datetime, subprocess`). |
| Double-evaluation of staleness window | ✅ | — | The new heartbeat path re-computes `tick_interval_ms` + `role_ticks` inside its own scope. Acceptable: only fires for daemons the memo path classified dead (`if not rec['pid_alive']`), which on healthy federations is the empty set; on containerized federations bounded by daemon count (~18). Dashboard refresh cadence is ~60s — two extra `freshness.*` calls per dead daemon is sub-millisecond. |
| Live empirical evidence | ✅ | — | Live dashboard snapshot inspection: **17/18 daemons reported alive** with the patched collector (vs the PR body's claimed 18/18). The single dead row is a genuinely-stale daemon: heartbeat ms-timestamp is ~17 minutes old (file line 3 reports `degraded`), `agent=None` (no role binding — daemon record without `.ag` mapping). This is a true positive, not a false negative — the fix correctly distinguishes a stale runtime from a healthy quiet listener. PR body's "18/18" claim was likely captured immediately after restart; the row degraded over time. |

### Verdict

✅ **ship-it**

Parity with #799 is byte-equivalent. Backward compatibility is preserved on all three exception axes (empty `agent_id`, missing file, malformed line 1). Live evidence confirms the dashboard now correctly classifies daemon liveness from heartbeat files, including correctly flagging a daemon that genuinely degraded. CI is green. VERSION + CHANGELOG conventions respected. The one ⚠️ row (test coverage for the new fallback path) is non-blocking and can be a follow-up — recommend opening a single test-extension issue for `tools/test-dashboard-freshness-liveness.sh` to add three heartbeat-path cases mirroring the existing memo-path sweep.

PR #834, opened 2026-05-28
fix(test): make auto-promote Test 19 hermetic (Fixes #826)

### LINKED ISSUE #826: test-auto-promote.sh Test 19 (no-memo liveness) flakes on main
`tools/test-auto-promote.sh` Test 19 (containerized liveness — no memo + stale registry skips with new reason, #706) fails on a fresh clone of `main` (commit 23eb1ff).

## Observed

```
Results: 39 passed, 1 failed
```

The single failing test is Test 19. Reason string the assertion expects:

```
memo last_check stale, heartbeat file stale or missing, and effective_state='zombie' not running
```

Actual reason emitted:

```
prerequisites not met: entries_total=0 < 30
```

## Likely cause

The agent fixture survives the containerized liveness gate (the gate now defers when no memo + no heartbeat exists, instead of short-circuiting with the `effective_state='zombie' not running` reason). The fixture has zero experience rows, so the prereq gate skips first with the entries-total failure, which is what we see.

So either:

1. Test 19's intent — short-circuit containerized liveness — has been superseded by the heartbeat-fallback (#800) and the assertion needs updating to match the new short-circuit conditions
2. The fixture needs an experience JSONL with 30+ entries so prereqs pass and the containerized-liveness reason is the actual blocker
3. The containerized-liveness gate now short-circuits later than the prereq computation, in which case the test contract is stale

Pre-existing on main; not a regression. Surfaced during #823 implementation.

## Repro

```
git clone https://github.com/Replikanti/agentis-colonies.git
cd agentis-colonies
bash tools/test-auto-promote.sh
```

ylohnitram:
## Summary

Fixes #826. `tools/test-auto-promote.sh` Test 19 (containerized liveness — dead daemon must skip) flaked on `main` depending purely on host state.

## Root cause (path A — test hermeticity, not a production bug)

Test 19 represented a "dead" daemon by the ABSENCE of a fed-local `<agent>:last_check` memo. But `read_memo_raw` shells `agentis memo get` with `cwd=fed_dir`, and the `agentis` binary falls back to the host operator's global memo store (`~/.config/agentis`) when the fed-local key is absent. That global store carries a real `explorer:last_check` from any actual federation run on the host — so:

- recent real run on host → fresh global memo → containerized-liveness gate falls through → dead daemon NOT skipped (promoted or prereq-skipped) → **test false-fails**
- no recent run → stale/absent global memo → gate fires → test passes

Confirmed via trace: `agentis memo set explorer:test_probe` lands fed-local and is read back correctly, but `agentis memo get explorer:last_check` (never set fed-local) returned a fresh `2026-05-28T06:05:40Z` from the host global store.

**This is test-only.** In production the federation directory always carries the real per-daemon `<agent>:last_check` memo (the container's `/run-root/.agentis/memo` bind-mount), so the global fallback never matters and the liveness gate works correctly.

## Fix

Seed an explicit STALE `explorer:last_check` (`2020-01-01T00:00:00Z`) into the FED_DEAD fixture via `agentis memo set` (cwd=FED_DEAD). Fed-local keys shadow the global store (verified), so `fresh` is deterministically `False`, and the no-heartbeat + zombie-registry liveness gate fires as the test intends. Mirrors Test 18's existing fresh-memo seeding (the live counterpart).

## Test plan

- [x] `bash tools/test-auto-promote.sh` — 43 passed / 0 failed (was 42/1)
- [x] Deterministic across 3 consecutive runs (no host-state dependence)
- [x] `bash tools/colony-lint.sh` — clean

## Possible follow-up (out of scope, benign in production)

The `agentis memo get` global fallback is a latent footgun for any host-side tool that reads a fed-local memo which happens to be absent. Production auto-promote is unaffected (fed dir always has the memo). Could file a separate hardening issue if desired (e.g. an `AGENTIS_NO_GLOBAL_MEMO` env for test isolation), but not needed for this fix.

### DIFF
--- a/tools/test-auto-promote.sh
+++ b/tools/test-auto-promote.sh
@@ -943,6 +943,19 @@ fi
 # promoted.
 FED_DEAD="$TMPDIR_TEST/fed-dead"
 mkdir -p "$FED_DEAD"
+# #826: seed an explicit STALE `<agent>:last_check` memo into the fixture so
+# the test is hermetic. Without a fed-local memo, `read_memo_raw` shells
+# `agentis memo get` (cwd=FED_DEAD), which falls back to the host operator's
+# GLOBAL memo store (~/.config/agentis) when the fed-local key is absent — and
+# that store carries a real, possibly-fresh `explorer:last_check` from any
+# actual federation run on the host. That made Test 19's outcome depend on
+# host state: a recent real run left a fresh memo → the containerized-liveness
+# gate fell through → the dead daemon was NOT skipped (promoted or prereq-
+# skipped instead), false-failing the assertion. A fed-local stale memo
+# shadows the global one (fed-local keys win, verified), so `fresh` is
+# deterministically False and the no-heartbeat + zombie-registry liveness gate
+# fires as intended. Mirrors Test 18's fresh-memo seeding (the live counterpart).
+(cd "$FED_DEAD" && agentis memo set "explorer:last_check" "2020-01-01T00:00:00Z" >/dev/null 2>&1)
 
 DAEMONS_DEAD='[{"pid":12345,"agent_id":"explorer-1","source":"explorer.ag","state":"zombie","effective_state":"zombie","colony":"discovery","confidence":0.4,"started_at":1}]'
 


ylohnitram:
## QA Report — PR #834 (Fixes #826)

**Verdict: ✅ ship-it**

Test-only hermeticity fix for `tools/test-auto-promote.sh` Test 19. The diff is a single file, +13 lines (12 comment + 1 seeding line), fixture-only. Root-cause analysis verified, fix proven deterministic, production auto-promote path untouched.

| Diff is test-only | ✅ | — | `git diff main...HEAD --stat` → `tools/test-auto-promote.sh \| 13 +++`. Single file, +13/-0. No prod `.py` / `.yaml` / `.ag` / config touched. |
| Seeded memo is STALE | ✅ | — | `2020-01-01T00:00:00Z` is ~56,000× older than the largest plausible staleness window (60 ticks × 60 s). `fresh=False` is unconditional regardless of any `STALENESS_TICKS` override or per-role window. |
| `cwd=FED_DEAD` writes fed-local | ✅ | — | `(cd "$FED_DEAD" && agentis memo set "explorer:last_check" …)` mirrors Test 18's `(cd "$FED_LIVE" && agentis memo set …)` byte-for-byte (same key, same subshell pattern). Fed-local key shadows global. |
| Test passes, deterministic ×3 | ✅ | — | `bash tools/test-auto-promote.sh` → **43 passed / 0 failed** on all 3 consecutive runs (was 42/1). |
| Hermeticity (host-state independence) | ✅ | — | Under an isolated `HOME` seeded with a **fresh** global `explorer:last_check`: **fixed** version = 43/0; **pre-fix (main)** version = **42/1** (the #826 false-fail). Confirms the fix removes the host-state dependence and that the diagnosed root cause is real. Operator's real `~/.config/agentis` was not mutated (isolated tmp HOME). |
| `read_memo_raw` global-fallback claim | ✅ | — | `tools/agentis_memo_freshness.py:127` shells `agentis memo get <key>` with `cwd=fed_dir`; absent fed-local key falls back to the host global store. PR root-cause is accurate. |
| colony-lint clean | ✅ | — | `bash tools/colony-lint.sh` → **345 passed / 0 failed / 5 skipped**. #826 flake (Test 19) no longer appears. |
| Test 18 (live counterpart) unperturbed | ✅ | — | `[PASS] containerized liveness: fresh memo overrides stale registry (#706)` still passes alongside Test 19. |
| No internal absolute paths in committed text | ✅ | — | Diff + PR body grep for `/home/`, `worktrees/`, `agentis-core` → NONE. |
| CHANGELOG convention | ✅ | No entry needed | `tools/check-changelog.sh` watches only versioned-component prefixes (`dev-apprenticeship/`, `federation-dashboard/`, `tribes-bench/`, `trading-binance/`, `research-foundry/`). A `tools/`-only change matches no prefix → not even a soft WARN. Correctly ships without a CHANGELOG entry. |
| Production behaviour unchanged | ✅ | — | The change only seeds a test fixture's memo. The production auto-promote path (`auto-promote-decisions.py:612–640`) reads the real per-daemon `<agent>:last_check` from the container's bind-mounted fed-local memo, so the global fallback never engages in production. Liveness gate logic untouched. |
| CI | ✅ | — | `colony-lint: COMPLETED SUCCESS` on the PR. |

### Notes
- The fix is correct, minimal, and mirrors the established Test 18 pattern. The accompanying comment block accurately documents the non-hermeticity and is a net readability win.
- The PR's noted latent footgun (host-side tools reading an absent fed-local memo can be misled by the global fallback) is genuinely out of scope and benign for production auto-promote. A hardening follow-up (e.g. test-isolation env) is optional, not blocking.

No blockers, no required mitigations. Green across the board.

Issue #868, opened 2026-05-29
research-foundry: explorer's 'exec sh python3' pattern bypasses the agentis substrate (CB / OCap / determinism / sandbox)


ylohnitram:
## The architectural issue

The `explorer` colony delegates heavy mathematical work (polynomial factorization, sequence search, group-theory enumerations, …) by having the LLM emit a Python program and invoking it via `exec sh python3 /run-root/.agentis/sandbox/explore-*-tick-N.py`. Surfaced operationally in [#867](https://github.com/Replikanti/agentis-colonies/issues/867) (runaway compute accumulating 11 GB RSS / 96% CPU spinners), but the runaway is **a symptom** — the root architectural concern is independent.

When the explorer (or any colony) reaches for `exec sh python3 ...`:

| Substrate guarantee | `exec sh python3` | WASM-compiled `.ag` |
|---|---|---|
| CB metering | bypassed | enforced |
| OCap enforcement | `ExecForeign` granted, but script body free | enforced |
| Evaluator state / experience tracking | bypassed | tracked |
| Determinism | not guaranteed | WASM sandbox |
| Wall-time bound | inline `timeout:` only | CB-limited |
| Memory bound | none (cgroup container-level only) | runtime-tracked |
| Replay (#583) | unfaithful (no `prompt()` recorded for the script) | faithful |
| Audit trail (#584) | shell command logged; script body not | full |

In other words: every substrate feature this repo has invested in (`crystallizer_*`, `learn`/`distill`, replay, audit chain, capability gates, evolution metering) **does not see the work the explorer actually does**. The federation's most compute-heavy step is opaque to the runtime.

## Why this happens

Honest read: pragmatism beats purity. The `.ag` language has 63 builtins (math ops, `json_get`, `parse_*`, `crystallizer_*`, …) but **no NumPy / SymPy / SciPy equivalent**, and no path to ship one short of rewriting half of CPython in `.ag`. Generating Python via LLM + shelling out is the path of least resistance, and it works.

But it's exactly what [`feedback_agentis_substrate_use`](https://github.com/Replikanti/agentis-colonies) warns against: leaning on `exec sh` as the escape hatch erodes the case for the substrate.

### 1. Containment (operational; small)
- Strict whitelist of `exec sh python3` patterns the federation is willing to host (only `explore-*.py` for now).
[...]

Issue #867, opened 2026-05-29
explorer: runaway 'exec sh python3 explore-*.py' compute scripts — no useful timeout / cleanup


ylohnitram:
## Problem
The `explorer` colony's `.ag` invokes mathematical exploration scripts via `exec sh python3 /run-root/.agentis/sandbox/explore-<role-pid>-tick-<N>.py` each tick. The scripts can be CPU-intensive (polynomial factorization, sequence searches, …) and there's no enforced ceiling on per-script wall time or accumulated memory.

Net effect on the host during the [#833](https://github.com/Replikanti/agentis-colonies/issues/833) verification run:

- 7+ Python `explore-*.py` scripts simultaneously running, each at 80-96% of one CPU core.
- Two of them holding **8 GB and 11 GB RSS** respectively.
- Container CPU ~568% on a 12-core host; host load average climbed to 25+.
- Scripts from tick 2 / 4 / 12 / 14 still alive at tick 38 — i.e. **~50+ minutes of CPU/memory burn per script** for context that the cascade has long since moved past.

## Why it happens
The `.agentis/config` `exec.default_timeout_ms = 10000` (10s) cap applies to `exec sh` invocations without an inline override. The explorer almost certainly passes an inline `timeout:` for the compute steps (necessary — many real searches take minutes), but that inline timeout is per-script and the federation has no policy on **how stale** an explore script is allowed to be before it's killed:

- Multiple explorer replicas (5 by default per [#824](https://github.com/Replikanti/agentis-colonies/issues/824)) each schedule their own scripts per tick.
- A script generated for tick 2 has no relevance to the cascade at tick 38; the explorer has moved on but the subprocess wasn't reaped.
- No back-pressure: the explorer doesn't check "is my previous tick's compute still running?" before launching a new one.

## Mitigation taken
Killed the stale scripts manually:
```
podman exec research-foundry-laptop sh -c 'pkill -9 -f "explore-.*-tick-(2|3|4|5|6|7|8|9|10|11|12|13|14|15)\\.py"'
```
Freed ~18.5 GB of host RAM in one shot.

## Proposed fix (federation side)
1. **Per-tick cleanup hook** in `explorer.ag`: before spawning a new compute script, kill any `explore-<my-pid>-tick-<N>.py` for `N < my_current_tick - K` (e.g. K=3 ticks lookback).
[...]

PR #875, opened 2026-05-30
feat(#868 L2): faithful tracking of substrate-bypass compute spend

ylohnitram:
## Summary

- Wrap explorer.ag's two `python3` exec sites in `/usr/bin/time -f '%e,%x'` and append one JSONL row per invocation to `<agentis-root>/compute/<self_pid>.jsonl` (`{ts_ms, agent, colony, tick, script, wall_clock_s, exit_code, source}` where `source ∈ {llm, rule-hit}`).
- New `collect_compute_spend()` in the collector aggregates per colony (calls, wall-clock seconds, llm/rule-hit splits, exit-124 timeout count) and merges into the existing `analytics_per_colony` rows.
- Dashboard **Analytics tab Card 1** grows **Compute calls** + **Wall-clock&nbsp;s** columns. Exit-124 (timeout fired) count surfaces as a red `N to` pill on the Compute calls cell.
- Extends the `#867` `timeout 300s` containment to the rule-hit replay path for symmetry.
- Bumps `federation-dashboard` 0.10.2 → 0.10.3 with CHANGELOG entry.

Refs #868 (Level 2: faithful tracking; Level 1 = #867 containment, already merged).

## Why

`#867` contained runaway explorer compute but the dashboard had no visibility into it — operators couldn't tell how much wall-clock the substrate was actually burning in `python3` exec sandboxes versus how much was LLM cost. The LinkedIn analytics framing (\"LLM calls per side-effect produced\") needs the denominator visible too. This PR makes compute spend a first-class column right next to LLM call count + rule-hit count on the same Analytics page.

## Default-off path

Federations that don't emit `compute/*.jsonl` rows (everything except `research-foundry`'s explorer today) render `—` in both new columns. Helper returns empty dict when the dir is missing → no behaviour change for `dev-apprenticeship`, `tribes-bench`.

## Test plan

- [x] colony-lint 345/0/5 (passed/failed/skipped) on full repo
- [x] Python AST parse on collector
- [ ] CI green on all checks (will wait before merging — never-merge-red-CI rule)
- [ ] Smoke check after merge: spawn 5-tick research-foundry run, verify `<root>/compute/*.jsonl` populates and dashboard Card 1 shows non-zero `Compute calls` + `Wall-clock s`


ylohnitram:
## QA Report — `feat/868-level2-compute-spend`

### Verdict: ⚠️ ship-it (one finding resolved inline)

The dashboard side of this PR is clean and the merge logic is correct. QA caught a runtime regression on the explorer side: the new `/usr/bin/time` wrapper pointed at a binary not installed in the federation's container image (stock `ubuntu:24.04 --no-install-recommends` does not install `time` — `Priority: standard`, not `essential`). Without it, every `python3` invocation would have exited `127` with `sh: 1: /usr/bin/time: not found` before the exploration script ran, the JSONL rows would have shown `wall_clock_s=0, exit_code=0`, and the explorer would have gone silent on both LLM and rule-hit paths. **Fixed inline** in commit `2c79f92` by adding `time` to the apt install list in `tools/Containerfile.research`.

### Findings

|---|---|---|---|
| Repo hygiene — `colony-lint.sh` | ✅ | — | `345 passed, 0 failed, 5 skipped` post-fix. |
| Repo hygiene — Python AST parse on collector | ✅ | — | `ast.parse` clean. |
| Repo hygiene — CI status on PR | ✅ | — | `colony-lint pass 3m16s` on commit `4c53582`; will re-run on `2c79f92`. |
| `explorer.ag` — `/usr/bin/time` availability in container | ✅ | **fixed in `2c79f92`** | Added `time` to `apt-get install` in `tools/Containerfile.research`. On next `run-research.sh` invocation podman rebuilds the image and the wrapper finds the binary. |
| `explorer.ag` — `compute_dir` fallback when `AGENTIS_ROOT` is empty | ⚠️ | **dismissed** | `compute_dir = sandbox_dir + "/compute"` uses raw env-var result, not `sandbox_eff`'s fallback. Asymmetric with two lines earlier, but research-foundry's `run-research.sh` unconditionally sets `AGENTIS_ROOT=/run-root/.agentis` so the path can't be empty in practice. Worth a follow-up if the explorer ever runs outside this orchestrator. |
| `explorer.ag` — both `python3` sites wrapped consistently | ✅ | — | LLM path (line 379) and rule-hit path (line 785) identical: `/usr/bin/time -f '%e,%x' -o … timeout 300s python3 …`. JSONL schema identical except `source` discriminator. |
| `explorer.ag` — `source` discriminator matches collector | ✅ | — | Writer emits `"llm"` / `"rule-hit"`; collector checks `if src == 'rule-hit'` else llm — round-trips correctly. |
| `explorer.ag` — `comma_idx` handling on empty / single-line sidecar | ✅ | — | `index_of("", ",")` → `-1`; `comma_idx > 0` defaults both fields to `"0"` and JSON stays well-formed. `tail -1` of GNU `time` correctly captures the `%e,%x` line even when a `Command exited with non-zero status N` warning precedes it (verified on `time-1.9` and on signal-kill cases). |
| `explorer.ag` — `_colony_name` for rule-hit path | ✅ | — | `_publish_explorer_rule_hit` signature accepts `_colony_name: string`; callers pass it from `tick()`'s `let _colony_name = colony_name();`. |
| `explorer.ag` — `[k]` self-match avoidance preserved | ✅ | — | Cleanup line 356 unchanged; `time` wrapper introduces no new argv pkill could self-match on. |
| Collector — `collect_compute_spend` correctness | ✅ | — | Missing dir → `{}`; unreadable file → `continue`; malformed JSON → `continue`; missing colony field falls back to `id_to_colony`; rounding to 3 decimals applied uniformly. |
| Collector — merge into `analytics_per_colony` | ✅ | — | Existing rows get `compute_*` added (no drop). `compute-only` injection gated on `colony in _seen_colonies` so no duplication. Final `.sort(key=lambda r: r['colony'])` deterministic. |
| Template — `—` for default-off rows | ✅ | — | `computeCalls === 0` short-circuits both new cells to `<span class="muted">—</span>`. |
| Template — `compute-only` pill style | ✅ | — | Same `class="pill"` + inline-style shape as the other three (`crystallizing`, `generative`, `active`). |
| Template — red `N to` timeout pill HTML safety | ✅ | — | `esc(timeouts)` coerces via `String()` and escapes; no user-controlled HTML in the pill. |
| CHANGELOG / VERSION bump | ✅ | — | dashboard 0.10.2 → 0.10.3 (MINOR — new feature). `[Unreleased] > Added` entry references #868. `check-changelog.sh` lint clean. research-foundry CHANGELOG also picks up the Containerfile `time` entry. |
| Default-off / backwards compat — `dev-apprenticeship`, `tribes-bench` | ✅ | — | No other federation emits compute rows; collector returns `{}`; existing rows get zeros; renderer prints `—`. Status pill, sort order, other columns unchanged. |
| Multi-line `/usr/bin/time` warning before format line | ✅ | — | `tail -1` correctly captures the `%e,%x` line in non-zero-exit, signal-kill, and timeout-124 cases. |
| Race condition — multi-replica writes to same `<pid>.jsonl` | ✅ | — | Each replica has unique `self_pid`, separate files; `>>` is O_APPEND. |

### Conclusion

Blocker resolved in `2c79f92`. Everything else (collector aggregation, template rendering, default-off byte-identity, CHANGELOG/VERSION, dash-safe parsing) verified correct. Ready to merge once the re-pushed CI run goes green.

Issue #833, opened 2026-05-28
research-foundry: breeding-loop verification run (75-tick claude-mathlib on agentis v1.10.0)


ylohnitram:
## Purpose

End-to-end empirical verification of the complete breeding-loop substrate now that all five completeness gaps are closed:

| Gap | Mechanism | Shipped in |
|---|---|---|
| (a) persistence | DAG-persist crystallized rules + telemetry sidecar replay | agentis v1.9.0 (#667/#671) |
| (b) replicate inheritance | M106 wire-format carries parent rule index | agentis v1.10.0 (#669/#674) |
| (c) cross-daemon discovery | `crystallizer_fetch_rule` auto-resolution via bus | agentis v1.10.0 (#670/#677) |
| (d) selection pressure | auto-promote fitness reads rule-hit ratio | colonies #823/#829 |
| (e) rule compaction | `crystallizer_compact()` + periodic trigger | agentis v1.10.0 (#668/#673) |

Plus enabling work: noticer distillation pilot (#820), population scaling (#828), per-tier LLM routing (#827). Federation pinned at `agentis >= 1.10.0` (Containerfile bump #832).

## Run configuration

To exercise compaction + peer-fetch + persistence within a 75-tick window (their defaults are sized for multi-hour deployments), the verification run overrides:

- `evolution.crystallize_compact_interval = 30` (default 500 > 75 — compaction would never fire otherwise)
- `evolution.crystallize_peer_fetch_enabled = true` (default false — cross-daemon metrics would be zero otherwise)
- Verify `evolution.crystallize_interval` is low enough that first rules crystallize by ~tick 30 (otherwise no rules to cache-hit on in the back half)
- Mid-run: manually restart one noticer daemon (~tick 40) to force a `crystallizer.loaded_from_disk` rehydration event (otherwise persistence only shows an empty boot-time load)

### 1. Distillation loop (the headline)
| Metric | Source | Proves |
|---|---|---|
| `cognitive.cache_hit` events from noticer | events.jsonl | rule-driven decisions replacing LLM calls |
| Noticer rule-hit ratio = `distilled`-tag learn rows / acting ticks | experience JSONL | fraction of ticks running without LLM |
| Noticer LLM-call rate per tick | `spend/*.jsonl` | should drop below 1.0 as rules mature |
| `evolution.knowledge_crystallized` events | events.jsonl | when crystallize batch fires |

[...]

PR #876, opened 2026-05-30
fix(#840): verify cached image agentis version before reuse in run-research.sh

### LINKED ISSUE #840: research-foundry: run-research.sh reuses :latest image without verifying baked AGENTIS_VERSION
## Bug

`run-research.sh` gates the image build on `podman image exists "$IMAGE_TAG"` (default `research-foundry:latest`). If an image with that tag exists from a PRIOR run, the build is skipped — **even when the existing image was built from a different (older) `ARG AGENTIS_VERSION`**. The script never checks that the baked-in `agentis` binary version matches the `AGENTIS_VERSION` the current `Containerfile.research` pins.

## Impact (caught during #833 verification)

The #833 breeding-loop verification run launched against a stale `research-foundry:latest` (id `2ae9e65...`, 8h old) that carried **agentis v1.8.0** — even though `Containerfile.research` now pins `v1.10.0` and the run depends on v1.10.0-only features (compaction #668, peer-fetch #670, DAG-persist #667). The federation booted v1.8.0 daemons; the `evolution.crystallize_*` config knobs would have been inert/erroring. Confirmed via `podman run --rm --entrypoint /usr/local/bin/agentis  --version` → `v1.8.0`. Run killed at tick 0 and re-launched after a forced rebuild — no data loss, but the run would have produced misleading (all-negative) breeding-loop metrics had it not been caught.

Compounding factor: a stale `:latest` whose container is still running cannot be `podman rmi`'d without `-f` (or stopping the container first), so a naive `podman rmi :latest` silently no-ops.

## Proposed fix

Before the `podman image exists` skip, verify the existing image's agentis version matches the pinned `AGENTIS_VERSION`:
```sh
if podman image exists "$IMAGE_TAG"; then
    img_ver=$(podman run --rm --entrypoint /usr/local/bin/agentis "$IMAGE_TAG" --version 2>/dev/null | grep -oE 'v[0-9]+\.[0-9]+\.[0-9]+' | head -1)
    if [ "$img_ver" != "$AGENTIS_VERSION" ]; then
        echo "[run-research] image $IMAGE_TAG has agentis $img_ver but Containerfile pins $AGENTIS_VERSION — rebuilding"
        REBUILD=1
    fi
fi
```
…and rebuild (or hard-fail with a clear message) on mismatch. Alternatively, tag images per-version (`research-foundry:$AGENTIS_VERSION`) so a version change always misses the cache.

## SDLC

Plan → implement → QA → merge. Lower urgency than the #833 run itself (worked around by forced rebuild), but real — silently testing the wrong runtime version is a serious integrity gap for any future run.

ylohnitram:
## Summary

- `build_image()` in `research-foundry/tools/run-research.sh` now probes the cached `research-foundry:latest` image's baked `agentis --version` and forces a rebuild when it doesn't match the pin in `Containerfile.research`.
- New `RESEARCH_FORCE_REBUILD=1` env knob skips the probe entirely and forces \`podman rmi -f\` + rebuild (operator escape hatch).
- All build invocations now pass `--build-arg AGENTIS_VERSION=$PINNED_AGENTIS_VERSION` so a cache-miss rebuild is reproducible (no implicit Containerfile default).
- 4 new dry-run assertions in `test-run-research.sh` cover the gate, the build-arg pin, and the force-rebuild path.

Closes #840. Plan: https://github.com/Replikanti/agentis-colonies/issues/840#issuecomment-4584395794.

## Why

Caught during #833 verification: `research-foundry:latest` had been built from agentis v1.8.0 but `Containerfile.research` pinned v1.10.0 — `podman image exists "$IMAGE_TAG"` short-circuited the build, the daemons booted v1.8.0, the new `evolution.crystallize_*` config knobs would have been inert, and the breeding-loop metrics would have been silently misleading. Verified again today: `:latest` currently carries v1.10.0, Containerfile pins v1.10.1 — same regression in another spot.

## Test plan

- [x] `bash -n research-foundry/tools/run-research.sh` — syntax OK
- [x] `bash research-foundry/tools/test-run-research.sh` — 307 passed, 0 failed (4 new)
- [x] `tools/colony-lint.sh` — 345 passed, 0 failed, 5 skipped
- [ ] CI green on PR
- [ ] Empirical verification on next operator-driven `run-research.sh` invocation: stale `:latest` triggers rebuild emit_step + `rmi -f` + pinned build


ylohnitram:
## QA Report — `feat/840-image-version-coherence` (commit f4fdeaf → 2ee171b)

### Verdict: ⚠️ ship-it (both findings resolved inline)

Tests pass clean (`test-run-research.sh`: 307 passed / 0 failed; `colony-lint.sh`: 345 passed / 0 failed / 5 skipped). The `build_image()` rewrite is sound, default-off, byte-equivalent for the common path, and the dry-run / non-dry-run split correctly avoids dragging `podman run` into the transcript-only path. Two LOW-severity findings raised in initial QA; both fixed in `2ee171b`.

### Area-by-area

| # | Area | Result | Mitigation | Notes |
|---|------|--------|------------|-------|
| 1 | `test-run-research.sh` suite | ✅ | — | 307/0; the 4 new assertions (42a-d) pass. |
| 2 | `colony-lint.sh` | ✅ | — | 345/0/5 skipped; no regressions from the diff. |
| 3 | `bash -n` syntax | ✅ | — | Both `run-research.sh` and `test-run-research.sh` parse clean. |
| 4 | New 42a–d assertions genuinely new | ✅ | — | No duplication with prior tests. |
| 5 | `build_image()` dispatch order (FORCE_REBUILD → DRY_RUN → live) | ✅ | — | FORCE_REBUILD short-circuit runs FIRST, so `RESEARCH_DRY_RUN=1 RESEARCH_FORCE_REBUILD=1` correctly emits `rmi -f` + pinned build transcript without a live probe. Confirmed in 42c/42d. |
| 6 | `emit_cmd "podman run ..."` NOT in dry-run path | ✅ | — | Live probe is direct shell, fenced inside `if [ "$DRY_RUN" = "1" ]; then ... return; fi`. Operators without podman never trip it. |
| 7 | Version-mismatch branch covers all 3 probe outcomes | ✅ | — | match → silent reuse; mismatch → log + `rmi -f` + pinned rebuild; empty → log + reuse with FORCE_REBUILD hint. |
| 8 | `RESEARCH_FORCE_REBUILD=1` skips probe | ✅ | — | Returns at line 792 before `podman image exists`. |
| 9 | `RESEARCH_FORCE_REBUILD=1` when image doesn't exist | ✅ | — | `podman rmi -f nonexistent:latest` returns 0 on podman 5.x. |
| 10 | `set -euo pipefail` × probe pipe | ✅ | — | Trailing `\|\| true` after `\| head -1` neutralises pipefail. |
| 11 | `set -euo pipefail` × `podman image exists` | ✅ | — | `if podman image exists ...; then` form suppresses `set -e`. |
| 12 | `awk` extraction on current Containerfile | ✅ | — | Returns `v1.10.1`; matches the regex used to probe. Only one `^ARG AGENTIS_VERSION=` line (the other ARG lines are LEAN_TOOLCHAIN / MATHLIB_REV). |
| 13 | `awk` edge case: missing line | ✅ | **fixed in `2ee171b`** | Added `: "${PINNED_AGENTIS_VERSION:?Containerfile.research is missing the 'ARG AGENTIS_VERSION=...' line; cannot derive pinned agentis version (#840).}"` right after the assignment. Orchestrator now halts immediately with a clear error pointing at the Containerfile, instead of silently forcing a rebuild every run with a `pins ` `` (empty) message. |
| 14 | `awk` edge case: trailing comment / quoted value | ⚠️ | **dismissed** | Hypothetical only — current Containerfile uses no inline comments or quoted ARG values; defense-in-depth not warranted today. Worth revisiting if the Containerfile syntax ever changes. |
| 15 | `--build-arg AGENTIS_VERSION=` pin on all build paths | ✅ | — | All 4 build invocations (FORCE_REBUILD, dry-run gate, mismatch rebuild, cache-miss) carry `--build-arg AGENTIS_VERSION=$PINNED_AGENTIS_VERSION`. Pre-PR had zero — that was the bug. |
| 16 | `podman rmi -f` for stale running container | ✅ | — | Both rebuild paths use `-f`. |
| 17 | CHANGELOG entry placement + format | ✅ | — | `[Unreleased] > Fixed` block, references #840. |
| 18 | Default-off backwards compat (cached image matches pin) | ✅ | — | Path: FORCE_REBUILD unset → DRY_RUN unset → `podman image exists` → probe → match → silent reuse. Byte-identical orchestrator log to pre-PR for the common path. |
| 19 | Test idiom match | ✅ | — | 42a/b reuse existing `$OUT` snapshot + `assert_contains`; 42c/d spawn `FORCE_REBUILD_OUT` exactly like 40w/40x do for `OPENROUTER_API_KEY`. No new framework. |
| 20 | `RESEARCH_FORCE_REBUILD` documented in env-var header | ✅ | **fixed in `2ee171b`** | Added 8-line entry under `RESEARCH_DRY_RUN` in the env-var doc block (lines 155-161). Operators reading the header now discover the new knob. |
| 21 | Live probe — minimal `podman run` (no `-v`, no `-e`) | ✅ | — | Cannot trip on bind-mount setup or unset HOST_CLAUDE_DIR. Container exits in <1s. |
| 22 | Probe output regex robustness | ✅ | — | `grep -oE 'v[0-9]+\.[0-9]+\.[0-9]+'` handles `v1.10.0`, `v1.10.1-dirty (build N)`. `head -1` defends against multi-line output. |

### Conclusion

Both QA findings resolved inline in `2ee171b`. The substrate fix (forcing reproducible `--build-arg AGENTIS_VERSION=` propagation + cached-image mismatch detection) is sound, well-tested, and default-off. Ready to merge once CI on `2ee171b` goes green.

PR #967, opened 2026-06-09
feat(dark-factory): program-specific invariant library (V5)

ylohnitram:
## V5 — program-specific invariant library

Each detected vulnerability class now drives PoC synthesis through a **class-specific invariant** instead of a single hardcoded signer-drain story, so the generated proof asserts the *right* property for the detected class.

### Changes (`auditor/agents/auditor.ag`)
- **`invariant_for(class)`** — the control/exploit invariant the PoC must assert, per class: ownership substitution (`MissingOwnerCheck`), identity mismatch (`AccountDataMatching`), program-id redirection (`ArbitraryCPI`), arithmetic wrap (`IntegerOverflow`), non-signer authority (`MissingSignerCheck`), plus a generic access-control default for any unrecognised token.
- **`poc_instruction(class)`** — embeds `invariant_for(class)` into both the `solana-program-test` harness prompt and the std-only prompt. The two-sided marker contract (`CONTROL OK:` + `INVARIANT VIOLATED:` + `exit(101)`) is unchanged, so the validation gate still blocks rigged/always-fire harnesses for every class.
- **`is_vuln_class(class)`** — detection now routes *every* recognised class to synthesis. Previously only `MissingSignerCheck` was synthesized and `IntegerOverflow` stopped at a non-committal `DETECTED` stub.
- **Class-aware report** — `severity_for` / `summary_for` / `impact_for` / `remediation_for` replace the hardcoded signer narrative, so the standardized Immunefi-shaped report is accurate to the detected class.

### Safety
The built-in deterministic template is signer-shaped (correct only for `MissingSignerCheck`). A non-`MissingSignerCheck` class with no usable LLM-generated PoC resolves to `inconclusive` — **never** a false-VERIFIED. Mis-classification only routes to synthesis; the two-sided gate remains the source of truth.

### Verification
- **Per-class invariant selection** — a probe over all six classes confirms each `poc_instruction(class)` carries its own invariant (e.g. `MissingOwnerCheck` → ownership-substitution EXPLOIT clause; unknown → generic access-control).
- **Offline-mock matrix** (hardened sandbox, fully offline):

  | Target | Detected | Verdict |
  |---|---|---|
  | vulnerable signer vault | MissingSignerCheck | **VERIFIED** (+ standalone re-verified) |
  | + rigged PoC candidate | MissingSignerCheck | **INCONCLUSIVE** (`no-control`) |
  | unchecked-`+=` deposit | IntegerOverflow | **INCONCLUSIVE** (routed; no overflow template offline) |
  | signer-guarded vault | none | **SAFE** |

- **Real-LLM std-mode** — with the `claude` backend the colony generates a two-sided PoC that VERIFIES end-to-end under the class-aware prompt.
- **`colony-lint`** — 353 passed, 0 failed, 5 skipped; `auditor.ag syntax OK`.

### DIFF
--- a/dark-factory/CHANGELOG.md
+++ b/dark-factory/CHANGELOG.md
@@ -16,6 +16,29 @@ Every release declares its runtime floor as `**Requires:** agentis >= X.Y.Z`.
 
 ### Added
 
+- Program-specific invariant library (V5): each detected vulnerability class now
+  drives synthesis through a class-specific invariant (`invariant_for(class)`)
+  instead of a single hardcoded signer-drain story. The PoC-generation prompt
+  (`poc_instruction(class)`) embeds the right control/exploit invariant per class —
+  ownership substitution for `MissingOwnerCheck`, identity mismatch for
+  `AccountDataMatching`, program-id redirection for `ArbitraryCPI`, arithmetic wrap
+  for `IntegerOverflow`, non-signer authority for `MissingSignerCheck` — and the
+  standardized report's severity / summary / impact / remediation are class-aware
+  (`severity_for` / `summary_for` / `impact_for` / `remediation_for`). Detection now
+  routes every recognised class to synthesis (previously only `MissingSignerCheck`
+  was synthesized and `IntegerOverflow` stopped at a "DETECTED" stub). The built-in
+  deterministic template is signer-shaped, so a non-`MissingSignerCheck` class with
+  no usable LLM-generated PoC resolves to `inconclusive` — never a false-VERIFIED.
+  The two-sided gate (`CONTROL OK:` + `INVARIANT VIOLATED:`) is unchanged and still
+  blocks rigged/always-fire harnesses for every class.
+
+### Changed
+
+- Detection verdict for `IntegerOverflow` in offline / `mock` mode is now
+  `inconclusive` (routed through synthesis with the overflow invariant) rather than
+  the previous non-committal `DETECTED`, since no deterministic overflow template
+  exists; a real LLM backend generates the two-sided overflow PoC.
+
 - Generalised detection (V3): an LLM-driven classifier (`classify_llm`) reads the
   program source and returns a vulnerability class (`MissingSignerCheck` /
   `MissingOwnerCheck` / `AccountDataMatching` / `ArbitraryCPI` / `IntegerOverflow`
--- a/dark-factory/auditor/agents/auditor.ag
+++ b/dark-factory/auditor/agents/auditor.ag
@@ -263,39 +263,79 @@ fn assess(run_out: string) -> string {
     return "verified";
 }
 
+// V5 (#843): is `cls` one of the synthesizable vuln classes (any class the LLM
+// detector emits, except "Safe"/"none"/"")? Routes detection -> synthesis. SHALLOW
+// leaf (string compares only) — safe at any agent depth.
+fn is_vuln_class(cls: string) -> bool {
+    if cls == "MissingSignerCheck" { return true; }
+    if cls == "MissingOwnerCheck" { return true; }
+    if cls == "AccountDataMatching" { return true; }
+    if cls == "ArbitraryCPI" { return true; }
+    if cls == "IntegerOverflow" { return true; }
+    return false;
+}
+
+// V5 (#843): the program-specific invariant the PoC must assert for the detected class.
+// This is the heart of V5 — it tells the synthesis prompt WHAT to prove (control = the
+// legitimate case that must be accepted; exploit = the attacker case that, if it
+// succeeds, violates the named invariant), so the generated PoC asserts the RIGHT
+// property per class instead of a generic signer drain. The generic default covers any
+// unrecognised token (e.g. AccountConfusion / type-cosplay, a sibling of
+// AccountDataMatching). SHALLOW leaf — safe at any agent depth.
+fn invariant_for(cls: string) -> string {
+    if cls == "MissingSignerCheck" {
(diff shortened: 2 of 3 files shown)


ylohnitram:
## Dark Factory V5 — Adversarial QA Report

End-to-end QA of the program-specific invariant library (V5) on `dark-factory/auditor/agents/auditor.ag`. Ran offline-mock against the agentis runtime, exercised the documented matrix, sliced the invariant functions to prove per-class selection is live, and ran four additional adversarial attacks specifically aimed at minting a false VERIFIED.

| `colony-lint.sh` | ✅ | — | `Results: 353 passed, 0 failed, 5 skipped`. `[PASS] dark-factory/auditor: auditor.ag syntax OK` present (only when `agentis` is on PATH — a PATH-less invocation silently skips the agentis-gated `.ag` block). |
| Matrix: `vuln_missing_signer` → VERIFIED | ✅ | — | `guard: MissingSignerCheck fired [Critical]`; mock yields no `fn main`, falls through to deterministic signer template; two-sided gate passes → `Verdict: VERIFIED`. |
| Matrix: `+ BOUNTY_POC=rigged_harness` → INCONCLUSIVE | ✅ | — | `Verdict: INCONCLUSIVE (inconclusive:no-control)`. Committed rigged (no-control) harness is correctly blocked by clause (c) of `assess()`. |
| Matrix: `vuln_overflow` → INCONCLUSIVE (not VERIFIED) | ✅ | — | `guard: IntegerOverflow fired [Medium]` → routes to synthesis → `cls != "MissingSignerCheck"` short-circuit returns empty `llm_out` before the signer template → `INCONCLUSIVE (inconclusive:no-violation)`. Never false-VERIFIED. |
| Matrix: `safe_guarded` → SAFE | ✅ | — | `guard: no rule fired — target looks safe` → `Verdict: SAFE`. No false positive. |
| Per-class invariant selection is live (not dead code) | ✅ | — | Sliced `invariant_for`+`poc_instruction`; each of MissingSignerCheck / MissingOwnerCheck / AccountDataMatching / ArbitraryCPI / IntegerOverflow emits a DISTINCT `PROVE THIS INVARIANT (<class>)` clause; an unknown token hits the generic default. Marker contract present in every variant. |
| `assess()` gate unchanged by V5 | ✅ | — | Byte-identical to the parent commit. Standalone re-verify + snapshot replay exec lines also byte-identical — the only delta there is the intended `"MissingSignerCheck"` → `cls` argument threading. |
| Offline determinism lynchpin (mock → `""`) | ✅ | — | mock `prompt()` returns the literal `"mock"` (matches no class token, not even "safe") → `classify_llm` returns `""` → falls back to the structural heuristic. Routing is deterministic offline. |
| VERIFIED report is class-correct | ✅ | — | `report.md` for the signer case: Severity Critical, Rule MissingSignerCheck, signer-specific Summary/Impact/Remediation, real `CONTROL OK:` + `INVARIANT VIOLATED:` observed output. |
| No internal path / network leak | ✅ | — | No `/home/`, `worktrees/`, or private-repo references in `report.md`; no `http_post`/egress builtins in `auditor.ag` (only doc-comments). |
| Forged `BOUNTY_POC` can mint a false VERIFIED | ⚠️ | Pre-existing (`assess()` unchanged by V5); out-of-V5-scope and tracked as a **non-blocking follow-up** to harden `assess()`. Reachable only via the explicit human `BOUNTY_POC` override, never autonomously. | A target-agnostic operator-supplied PoC that prints both `CONTROL OK:` and `INVARIANT VIOLATED:` *without exercising the target* → `Verdict: VERIFIED` on any flagged class. The gate trusts marker substrings; it cannot detect a fabricated control. The committed rigged (no-control) harness IS blocked — the property V5 actually asserts. |
| Empty / unreadable `BOUNTY_TARGET` silently audits embedded default | ⚠️ | Pre-existing ingest fallback (not touched by V5); tracked in the same follow-up. Documented `embedded_target()` behavior. | An empty target file → `ingest_target()` falls back to the embedded vulnerable program. Misleading but not a false-VERIFIED of attacker-controlled input. |
| Guard keyword in a string literal masks a real vuln (false negative) | ⚠️ | Pre-existing in the structural detector (untouched by V5); tracked in the same follow-up. Safe-direction failure (under-reports, never false-VERIFIES) and mitigated by the V3 LLM classifier in real use. | A vulnerable `withdraw` that mentions `.is_signer`/`require!` inside a string literal → detector reports SAFE. `strip_comments` strips `//` but not string literals. |
| Control side is non-discriminating in the std-only stub template | ⚠️ | Pre-existing template property; dismissed — the discrimination guarantee comes from the rigged-harness negative test (passed), not the synthetic control. | The deterministic template's "control" withdraws the full balance, which the unguarded target accepts. It proves `Ok`, not that an unauthorized caller would be rejected. |

### Verdict: ✅ ship-it

V5 does exactly what it claims and introduces no regression. All five V5 design claims verified: per-class `invariant_for`/`poc_instruction` are live and distinct; every recognised class routes to synthesis via `is_vuln_class`; the report narrative is class-aware; the signer-shaped deterministic fallback is correctly walled off behind the `cls != "MissingSignerCheck"` short-circuit so non-signer classes resolve to `inconclusive` (never false-VERIFIED); and the two-sided `assess()` gate is byte-for-byte unchanged. The exact-verdict matrix matches the spec, lint is `0 failed`, and no internal paths leak.

The one substantive concern — a fabricated-control `BOUNTY_POC` defeating the two-sided gate — is **pre-existing** (`assess()` is unchanged by V5), reachable only through the explicit human-supplied override, and cannot be triggered autonomously or by the LLM/template paths. It is tracked as a non-blocking follow-up to harden `assess()` so the control side must demonstrably exercise the target. It does not gate this PR.

_QA performed with `agentis-qa-agent` (offline-mock, hardened sandbox)._

PR #984, opened 2026-06-11
dark-factory: DAG bug-pattern matching — seed real findings, match N-day forks (#861)

ylohnitram:
## Why

Empirical hunts proved the colony's LLM detection — **single-pass AND multi-lens ensemble** — returns SAFE on real audited code: **13/13 real contracts** (Acre 7, ENSv2 6). So it finds nothing to verify on real targets. The content-addressed DAG was only doing **self-dedup** of one run's sub-graphs — no external bug knowledge. This seeds it with **real findings from finished contests**, so the colony recognizes recurring patterns (N-day forks) the LLM misses.

## Mechanism

- **`match_seeded_evm(code)`** (auditor.ag): per `function ` sub-graph, `dag_put` → look up `bugpat:exact:<hash>`, mirroring `distill_subgraphs_evm`'s hashing exactly so a fork's byte-identical vulnerable function hits. Leaf calls only (shallow stack, like the existing distill).
- **reconn** emits `guard:patmatch` (matched class); **guard** fires it **directly** when present, beating the LLM classifier's conservative SAFE. The two-sided real-EVM gate still verifies — a stale/over-broad seed **can never mint a false VERIFIED**, worst case one inconclusive synthesis.
- **`seed-patterns.ag`** + **`run-audit.sh --seed-manifest`**: record a finding's vulnerable-function pattern (`Class|src|func-marker`) into `bugpat:exact` before the audit, in the same store.

## Proof (end-to-end)

Seed VulnToken's `mint` (AccessControl) → audit a **fork** (`contract ForkToken`, byte-identical `mint`):
```
seed-patterns: seeded bugpat:exact:f4dccf45... = AccessControl (fn ~mint)
reconn: SEEDED PATTERN MATCH -> AccessControl (a target sub-graph forks a recorded finding)
guard: AccessControl fired [Critical] (seeded known-bug pattern match — DAG)   # NOT the LLM
synthesis: VERIFIED — real-EVM two-sided PoC (revm)
Verdict: VERIFIED
```
(Mock run first confirmed the match+guard-fire without any LLM; claude run confirmed the synthesis verify.)

## Scope

Exact-hash match catches **byte-identical N-day forks** (common + monetizable on its own). Structural-variant abstraction (match variants, not just identical) + a harvest of real finished-contest findings are the next steps (agentis-core#861). Submission stays human-gated — the colony never auto-posts. colony-lint 354/0/5.

### DIFF
--- a/dark-factory/CHANGELOG.md
+++ b/dark-factory/CHANGELOG.md
@@ -33,6 +33,17 @@ Every release declares its runtime floor as `**Requires:** agentis >= X.Y.Z`.
 
 ### Added
 
+- DAG bug-pattern matching — seed the federation's content-addressed DAG with real findings so the
+  colony recognizes recurring patterns (N-day forks) on real targets (#861). `seed-patterns.ag` +
+  `run-audit.sh --seed-manifest` record a finished-contest finding's vulnerable-function sub-graph as
+  `bugpat:exact:<hash> = class`; reconn's new `match_seeded_evm` looks up each target sub-graph against
+  the seed (mirroring `distill_subgraphs_evm`'s hashing) and guard fires the matched class **directly**,
+  beating the LLM classifier's conservative SAFE on real audited code (which returned SAFE on 13/13 real
+  contracts in testing). The two-sided real-EVM gate still verifies, so a stale/over-broad seed can never
+  mint a false VERIFIED — worst case one inconclusive synthesis. Proven end-to-end: seed VulnToken's
+  `mint` (AccessControl) → a fork (renamed contract, identical `mint`) matches → guard fires via the seed
+  (no LLM) → synthesis VERIFIED. Exact-hash match catches byte-identical N-day forks; structural-variant
+  matching + a harvest of real findings are the next steps.
 - Decomposed EVM PoC synthesis (#982). The synthesis agent no longer asks the LLM for the WHOLE
   `poc.rs` in one prompt — a large OUTPUT that stalls `claude -p` on a non-trivial contract (a real
   target's one-shot never returned at a 600s timeout; a small-output fragment prompt returns in
--- a/dark-factory/auditor/agents/auditor.ag
+++ b/dark-factory/auditor/agents/auditor.ag
@@ -303,6 +303,23 @@ fn distill_subgraphs_evm(code: string) -> int {
     }, 0);
 }
 
+// #861: match the target's function sub-graphs against the SEEDED known-bug DAG. For each
+// "function " sub-graph, look up its content hash in the seeded exact-pattern memo
+// (`bugpat:exact:<hash>` = class, populated from finished-contest findings by seed-patterns.ag).
+// Returns the first matched vuln class, or "" when nothing matches. Mirrors distill_subgraphs_evm's
+// hashing EXACTLY so a FORK with a byte-identical vulnerable function hits (N-day). Leaf calls only
+// (dag_put + recall_latest), so the native call stack stays shallow like distill_subgraphs_evm.
+fn match_seeded_evm(code: string) -> string {
+    return reduce(regex_split("function ", code), |acc: string, c: string| -> string {
+        if len(acc) > 0 { return acc; }
+        if len(c) == 0 { return acc; }
+        let h = dag_put("function " + c);
+        let cls = recall_latest("bugpat:exact:" + h);
+        if is_vuln_class(cls) { return cls; }
+        return acc;
+    }, "");
+}
+
 // Pick the prior verdict for a sub-graph: federation blacklist wins, else own cache.
 fn pick_prior(cached: string, black: string) -> string {
     if len(black) > 0 {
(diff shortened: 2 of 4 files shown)


ylohnitram:
## QA — PR #984: dark-factory DAG bug-pattern matching (#861)

**Verdict: ✅ ship-it.** Every claimed behavior reproduces; the security-critical invariant (a bad/over-broad seed can never mint a false `VERIFIED`) holds empirically; hashing parity is exact both at source and at runtime; the two-sided real-EVM gate is byte-for-byte unchanged; `colony-lint` = 354 passed / 0 failed. One pre-existing, out-of-scope crash observed on a std-only Rust target (reproduces identically on the parent commit) — noted, not a regression.

| Hashing parity (seed ↔ reconn) | ✅ | — | `match_seeded_evm`, `distill_subgraphs_evm` (auditor.ag) and seed-patterns.ag all use `regex_split("function ", strip_comments(...))` → `dag_put("function " + c)`; `strip_comments` is byte-identical across both files. Empirically corroborated: VulnToken `mint` and ForkToken `mint` both hash to `f4dccf4501f532c5`. |
| `.ag` parse / commit | ✅ | — | `agentis init && agentis commit auditor.ag` rc 0 AND `agentis commit seed-patterns.ag` rc 0 on on-PATH agentis v1.18.27 (no debug fallback needed). |
| Reproduce the proof (core claim) | ✅ | — | `run-audit.sh --repo ~/.../fork-proj --in-scope src/ForkToken.sol --backend mock --sandbox none --seed-manifest <AccessControl\|VulnToken\|mint> --out /tmp/qa-dag` → log shows `seed-patterns: seeded bugpat:exact:f4dccf4501f532c5 = AccessControl`, `reconn: SEEDED PATTERN MATCH -> AccessControl`, `guard: AccessControl fired [Critical] (seeded known-bug pattern match — DAG)`. Verdict INCONCLUSIVE under mock (expected — seed fires the class before any LLM; the gate can't be satisfied without a real PoC). |
| No false-VERIFIED from a bad seed (security-critical) | ✅ | — | Seeded `AccessControl` against the SAFE `AccessControlVaultSecure.sol` (`onlyOwner`-gated `setOwner`) and audited that same SAFE contract: `guard: AccessControl fired` BUT verdict stayed `INCONCLUSIVE (inconclusive:no-violation)`, **no submission package staged**. Confirmed in code: a seeded match only sets `synth:class`; `synthesis()` prints `VERIFIED` only when `assess(run_out)=="verified"` (needs both `CONTROL OK:` AND `INVARIANT VIOLATED:` + exit 101), and `assess()` is unchanged. |
| Non-EVM doesn't block | ✅ | — | Both reconn branches emit `guard:patmatch` (EVM: matched class; non-EVM: literal `""`), so guard's `listen("guard:patmatch")` never blocks. Verified in code (auditor.ag L1370 / L1393) and empirically: a std-only Rust target run with a seed manifest present did NOT hang on the patmatch listen (run returned, rc≠124). |
| exec-sh safety + bash syntax | ✅ | — | `check-exec-sh.sh auditor.ag` rc 0; `check-exec-sh.sh seed-patterns.ag` rc 0; `bash -n run-audit.sh` rc 0. |
| `colony-lint.sh` | ✅ | — | 354 passed / 0 failed / 5 skipped. |
| Restore runtime-overwritten files | ✅ | — | Runs overwrote `evm-harness/src/bin/poc.rs` + `contracts/bin/AccessControlVaultSecure.bin`; restored via `git checkout`. Working tree clean. |
| Std-only Rust synthesis path | ⚠️ | Out of scope for this PR — **pre-existing**, reproduces identically on parent commit; recommend a follow-up issue. | Auditing a std-only Rust `.rs` target under mock crashes with `stack overflow` in the `synthesis` agent (`fatal runtime error: stack overflow, aborting`). The PR diff touches ONLY the EVM reconn/guard branches + seed plumbing; the non-EVM guard branch and the entire `synthesis()` body are unchanged. Verified pre-existing: same crash on parent `02a0f2b` via a detached worktree. NOT triggered by the seed/patmatch path (crash is in synthesis, after guard exits; happens with or without `--seed-manifest`). |

### ⚠️ detail — std-only Rust synthesis stack overflow (pre-existing, out of scope)

A std-only Rust target (no `--harness`/`--evm-harness`, a plain `pub fn withdraw(...)` shape) audited under `--backend mock --sandbox none` aborts:

```
[agent synthesis] entered, CB=1998086
thread 'main' has overflowed its stack
fatal runtime error: stack overflow, aborting
```

- Reproduces **with and without** `--seed-manifest` → not caused by this PR's seed/patmatch code.
- Reproduces on the **parent commit `02a0f2b`** (checked out in a throwaway worktree) → pre-existing, not a regression.
- The PR's diff is confined to the EVM branches + the new seed files; `synthesis()` and the non-EVM guard branch are byte-unchanged (`git diff 02a0f2b HEAD -- auditor.ag`).
- **Action:** recommend opening a follow-up issue for the std-only Rust synthesis path's native call-stack depth in the runtime. Does NOT block this PR (the PR's scope is the EVM N-day-fork path, which is clean end-to-end).

### Minor observation (no action)
`seed-patterns.ag` splits on the EVM marker `function `, so a non-Solidity source yields a single chunk; if that chunk contains the marker it still writes one `bugpat:exact:<hash>` memo. Harmless: the non-EVM reconn path never calls `match_seeded_evm` (it emits `guard:patmatch ""`), so a stray non-EVM seed is dead memo, never consumed.

PR #986, opened 2026-06-11
dark-factory: M1+ structural-variant bug-pattern matching (#861)

ylohnitram:
## M1+ — structural-variant bug-pattern matching (#861)

Builds on M1 (exact-hash seed matching, #984). M1 only fires on a **byte-identical** N-day fork of a recorded finding. M1+ adds a structural fallback that also catches a **renamed / reformatted / re-littered** fork — the common copy-paste-rename N-day variant.

### How

`evm-harness/struct-sig.js` computes a **parser-free structural signature** per Solidity function: identifier names → `_`, number/string literals → `0`, keeping Solidity keywords / types / external-call kinds (`.call` / `.delegatecall` / `.transfer` / …). Each function is extracted precisely (brace-matched), so the signature is independent of the surrounding functions. A variant collapses to the same content hash as the seed.

**Why no AST?** The seed side is always a harvested *fragment* (a function lifted out of its contract, full of undefined references), and solc suppresses the AST on the first `Undeclared identifier` — so an AST signature is unavailable on the seed side. A token normalizer is the only thing symmetric across a fragment (seed) and a full contract (target). Verified empirically before building.

- `seed-patterns.ag` seeds `bugpat:struct:<hash> = class` alongside `bugpat:exact:<hash>`, guarded by `struct_has_signal()` so a trivial getter/setter (no call-kind, no storage-write) is never seeded as a noisy pattern.
- `auditor.ag` — `match_seeded_struct_evm()` runs `struct-sig.js` on the target and looks each function's sig up in the seeded struct memo; `match_seeded_any_evm()` tries exact first (cheaper, most precise), then the structural fallback.
- `run-audit.sh` passes `EVM_HARNESS_DIR` into the seed loop so the seeder can compute struct sigs.

### Proof (end-to-end through the colony, mock backend)

A Reentrancy seed (a vulnerable `withdraw`) matched a **renamed + reformatted + re-littered** variant (`LiquidityPool.cashOut` with renamed vars, different whitespace, different strings):

```
seed-patterns: seeded bugpat:struct: 7e5e4346a88346f9 = Reentrancy (fn withdraw — variant match)
reconn: SEEDED STRUCTURAL MATCH -> Reentrancy (... names/literals differ, shape matches)
guard: Reentrancy fired [High] (seeded known-bug pattern match — DAG)
```

Exact-match returned nothing on this variant (not byte-identical) — only the structural path caught it. A **CEI-reordered SAFE** version (state updated before the external call) correctly did **NOT** match: its sig differs.

### Bounds (honest)

- **Catches:** rename / reformat / re-literal forks that preserve expression structure.
- **Does NOT catch (v1):** a structurally-edited variant — reordered statements, a different expression shape. That needs an AST / semantic signal (e.g. richer nodes for guard-presence, oracle-staleness) and is the next recall step. Shape-based classes (Reentrancy, UncheckedCall) benefit most; semantic classes (oracle staleness, missing-guard) are not a *shape* and are largely unaffected by structural matching.
- **Safety:** an over-broad match can never mint a false finding — it only sets the candidate class; the two-sided synthesis gate (`CONTROL OK:` AND `INVARIANT VIOLATED:` + exit 101) stays the only source of truth. A stale/wrong seed costs at most one inconclusive synthesis. Submission stays human-gated.

colony-lint: 354 passed, 0 failed, 5 skipped.

### DIFF
--- a/dark-factory/CHANGELOG.md
+++ b/dark-factory/CHANGELOG.md
@@ -14,6 +14,24 @@ Every release declares its runtime floor as `**Requires:** agentis >= X.Y.Z`.
 
 ## [Unreleased]
 
+### Added
+
+- M1+ structural-variant bug-pattern matching (#861) — `evm-harness/struct-sig.js` + a new reconn
+  fallback. Exact-hash seed matching (the prior M1) catches only a byte-identical N-day fork of a
+  recorded finding; this also catches a RENAMED / REFORMATTED / RE-LITTERED fork. `struct-sig.js`
+  normalizes each Solidity function to a parser-free structural signature (identifier names → `_`,
+  literals → `0`, keeping Solidity keywords / types / external-call kinds), so a variant collapses to
+  the same content hash as the seed — no solc, so it works on a bare harvested fragment too.
+  `seed-patterns.ag` now seeds `bugpat:struct:<hash> = class` alongside the exact one (guarded to sigs
+  that carry a call-kind or storage-write, so a trivial getter is never seeded), and reconn
+  (`match_seeded_any_evm`) tries the exact match first, then the structural fallback. Proven end-to-end
+  through the colony: a Reentrancy seed matched a renamed/reformatted variant (`SEEDED STRUCTURAL
+  MATCH -> Reentrancy`, guard fired `[High]`) where exact-match returned nothing, while a CEI-reordered
+  SAFE version correctly did NOT match. A structurally-edited variant (reordered statements / changed
+  expression shape) is out of scope for v1 — that needs an AST/semantic signal. An over-broad match can
+  never mint a false finding: it only sets the candidate class; the two-sided synthesis gate stays the
+  only source of truth.
+
 ### Fixed
 
 - Decomposed-synthesis EXPLOIT slot now uses `try_call`/`try_call_value` (revert-tolerant) for the
--- a/dark-factory/auditor/agents/auditor.ag
+++ b/dark-factory/auditor/agents/auditor.ag
@@ -320,6 +320,48 @@ fn match_seeded_evm(code: string) -> string {
     }, "");
 }
 
+// #861 M1+: STRUCTURAL variant match. Where match_seeded_evm needs a byte-identical fork, this
+// catches a RENAMED / REFORMATTED / RE-LITTERED fork: struct-sig.js normalizes each target
+// function (identifier names -> `_`, literals -> `0`, keeping Solidity keywords/types/call-kinds)
+// and we look the content hash up in the seeded `bugpat:struct:<hash>` memo. struct-sig.js is pure
+// text (no solc, no node_modules) so it runs anywhere and degrades to no output (no match) on any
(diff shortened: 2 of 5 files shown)


ylohnitram:
## QA — `agentis-qa-agent` (M1+ structural matcher)

Exercised `struct-sig.js` against synthetic fixtures (variant invariance, discrimination, brace-matching, normalizer, delimiter, degradation) + reviewed the `.ag` wiring. colony-lint: **354 passed / 0 failed / 5 skipped**.

| Variant invariance — renamed/reformatted/re-littered fork → same sig | ✅ | — | 3 renamings collapse to the identical sig; end-to-end the seeded `withdraw` byte-equals the renamed `cashOut` fork. |
| Discrimination — CEI reorder / `call` vs `delegatecall` vs `transfer` / extra guard / `uint256` vs `uint128` | ✅ | — | All produce distinct sigs ≠ seed (true negatives). |
| Brace-matcher robustness (nested, struct/array literal, no-body decl, multi-fn, modifiers+returns, assembly) | ✅ | — | Each function isolated correctly; `function` inside a real body skipped. |
| Normalizer (hex/sci/underscored/string-with-keyword literals → `0`; types/call-kinds kept) | ✅ | — | `.call{value: x}("")` keeps ` call ` and ` value `. |
| Delimiter `:::` safety | ✅ | — | Cannot be spoofed by target content (a string `"a:::b"` → `0`). |
| Graceful degradation (empty / no-fn / unbalanced / missing / binary) | ✅ | — | Exit 0, no stack trace, sane output. |
| Seed↔match sig symmetry (bare fragment vs full contract) | ✅ | — | Identical sig ⇒ seed hash and match hash collide — the load-bearing property holds. |
| Wiring — exact-then-structural, "" on no match, degrades on unset `EVM_HARNESS_DIR`, no regression to exact-match (#984) | ✅ | — | `""` falls through to the normal detector; structural is purely additive. |
| **`struct_has_signal` `] -= ` / `] += ` needles dead vs tokenized sig** | ✅ | **Fixed `ec0617f`** — needles now match the split form `] - = ` / `] + = `. | struct-sig.js space-joins tokens so `-=`→`- =`; the old glued needles never matched → a debit-/credit-only seed was silently not seeded. Was masked for the reentrancy class (` call ` trips the gate independently). Verified post-fix: slash/credit/store trip the gate, a getter does not. |
| `function`-keyword inside a string BETWEEN functions → spurious entry | ✅ | acceptable — symmetric on seed+match, malformed sig (no body), gate-safe; astronomically unlikely to collide. | Over-match only routes a candidate to the two-sided gate, never mints a finding. |
| `] = ` over-fires on `==` (read-only comparator seeded) | ✅ | acceptable — over-seed is gate-safe by design (wastes a synthesis cycle at worst; the two-sided gate is truth). | Safe direction (false-positive, not false-negative). |

### Verdict

✅ **ship-it** — the headline feature (structural variant matching) is correct and robust across all axes; the one substantive finding (7b, dead index-write needles) is fixed in `ec0617f` and re-verified; the two over-match cases are gate-safe by design. colony-lint green.

PR #993, opened 2026-06-13
fix(dark-factory): hunter was blind — reason-first prompt (returned SAFE on textbook bugs)

ylohnitram:
## The bug (critical)

The discovery hunter's prompt **suppressed all findings**, returning `SAFE` even on blatant High-severity bugs:
1. It forced an immediate single-line verdict — *"output EXACTLY one line, nothing else — no preamble, no markdown"* — leaving the model no room to reason through a bug (LLMs need chain-of-thought to find vulns), so it defaulted to `SAFE`.
2. It primed dismissal — *"this code is likely already audited, report ONLY a bug you are confident survives expert judging"* — telling the model to assume bugs were already caught.

**Consequence:** every prior hunter `SAFE` is an **unvalidated false negative**, not a true negative. A bug-finding system that finds nothing is not "clean targets" — it may be a blind detector. It was.

## The fix

- Let the model **reason step-by-step first**, then emit `CANDIDATE|…` line(s) or a final `SAFE`. `run-discovery.sh` greps `CANDIDATE|` anywhere, so free-form analysis before the verdict is parsed correctly.
- Drop the "already-audited / dismiss" framing → *"an obvious-looking bug IS a real bug unless a guard provably stops it; do NOT assume prior auditors caught it."*
- `outcome_of` matches `CANDIDATE|` anywhere (`>= 0`), not only at offset 0, so `learn()` fitness is right with reasoning-first output.

## Validation (positive control)

A `VulnVault` with 3 known textbook High bugs (reentrancy, first-depositor inflation, missing access control):

| | reentrancy (C8) | first-depositor (C11/C6) | missing-auth (C5) |
|---|---|---|---|
| **before** (blind) | SAFE | SAFE | SAFE |
| **after** (fixed) | ✅ CANDIDATE | ✅ CANDIDATE | ✅ CANDIDATE |

After the fix: **4/4 cells CANDIDATE**, each with the correct exploit path + a concrete Foundry PoC sketch. A plain `claude -p` on the same source finds the same bugs — confirming the model was capable and the *prompt* was the blindfold.

> Follow-up: re-screening the previously-"SAFE" real targets (dreUSD, CapyFi, Alchemix, Twyne) with the fixed hunter, since those negatives were produced by the blind version.

### DIFF
--- a/dark-factory/auditor/agents/hunter.ag
+++ b/dark-factory/auditor/agents/hunter.ag
@@ -51,7 +51,7 @@ fn class_section(taxo: string, cls: string) -> string {
 // A surfaced candidate = success; a rigorous SAFE = failure (so taxonomy fitness rewards the
 // classes that actually produce leads across targets, per #861's evolve loop). No reassignment.
 fn outcome_of(verdict: string) -> string {
-    if index_of(verdict, "CANDIDATE|") == 0 { return "success"; }
+    if index_of(verdict, "CANDIDATE|") >= 0 { return "success"; }
     return "failure";
 }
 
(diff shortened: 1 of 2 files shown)


ylohnitram:
Retroactive QA (merged on green CI before the QA cluster). Consolidated report: https://github.com/Replikanti/agentis-colonies/pull/1004#issuecomment-4699363262 — hunter reason-first fix ✅ (the CANDIDATE-anywhere heuristic noted as pre-existing/low).

Issue #285, opened 2026-04-24
start-colony.sh --restart-agent leaves the old daemon process running (duplicates accumulate)


ylohnitram:
`<colony>/scripts/start-colony.sh --restart-agent <name>` spawns a new daemon for `<name>` but **does not kill the previous instance**. Each invocation adds another live process for the same agent. Repeated restarts (e.g. dashboard `/restart`, dashboard `/confidence` triggering a respawn after a tier change, or any operator-driven manual restart) silently grow the federation.

**Reproducer**

```
$ pgrep -cf 'agentis daemon-inner.*agents/plan_reviewer'
1
$ planning/scripts/start-colony.sh --restart-agent plan_reviewer
started plan_reviewer pid=42310 tick=60000
$ pgrep -cf 'agentis daemon-inner.*agents/plan_reviewer'
2          # ← old PID still alive, new PID alongside it
$ planning/scripts/start-colony.sh --restart-agent plan_reviewer
started plan_reviewer pid=52561 tick=60000
$ pgrep -cf 'agentis daemon-inner.*agents/plan_reviewer'
3          # ← three live daemons for the same agent now
```

Same was reproduced in this federation across 5 agents that I restarted as part of a confidence-tier promote flow:

```
$ ps -eo pid,command | awk '/agentis daemon-inner.*agents\// { …grouping… }'
  security_reviewer       3 copies
  labeler                 3 copies
  test_reviewer           3 copies
  logic_reviewer          3 copies
  issue_creator           3 copies
total daemon-inner processes: 31     # expected 21 for this federation
```

**Why it bites**

1. **Daemon registry collapses to one entry per agent_id.** The `agentis daemon list` view shows just one entry per agent (correctly indexed by agent_id), so the duplicates are invisible to the operator unless they hit `pgrep` or `ps` directly.
2. **Log-file lock contention.** All N copies of the daemon append to the same `<agent_id>.log` file. In my federation, this manifested as a daemon that appeared `state: running, no_ops: 0, skip_no_input: 0` while the log mtime stayed pinned for 4 hours — the lock contention or IO interleaving wedged the writer thread.
3. **Resource leak.** With 5 agents at 3 copies each I had 50% more `agentis daemon-inner` processes than intended, all chewing the same memo store, the same colony bus port, and the same forge API rate-limit budget.
[...]

PR #1017, opened 2026-06-13
fix(tests): deterministic + concurrency-safe daemon-restart test tooling (#1008)

### LINKED ISSUE #1008: Flaky: tools/test-llm-per-colony.sh t5 --restart-agent races on host load
The `t5: --restart-agent succeeds ...` subcase of `tools/test-llm-per-colony.sh` flakes under host CPU contention (e.g. several `colony-lint.sh` runs executing concurrently across worktrees on one machine).

## Symptom

```
[FAIL] t5: --restart-agent succeeds with both override file AND [llm] block: expected 'started <agent>' line in stdout:  / stderr:
```

Both stdout AND stderr are empty — the test expected `start-colony.sh --restart-agent` to print `started <agent> pid=<n> tick=<ms>` but read nothing.

## Root cause (timing fixture, not product code)

The test installs a **stub** `agentis` (no real binary needed) whose `daemon` branch `sleep`s "just long enough for start-colony.sh's `sleep 0.5`" before the shim is considered up (see the `SHIM` heredoc, ~line 81). It is a hardcoded sleep-vs-sleep race: under load the 0.5s pre-flight in `start-colony.sh` elapses before the stub writes its `started` line, so the harness captures empty output and fails the subcase. The same lint invocation's failure-path *re-run* of the test then passes (`Results: 16 passed, 0 failed`), and a standalone loop reproduces it intermittently (observed ~5/8 on a loaded host, 0/5 → 1/5 on a quiet one).

## Evidence it is pre-existing + environmental

- Reproduces on `main` (commit 86c6484) with no working-tree changes: 5/8 standalone runs failed on the same subcase.
- Correlates with concurrent `colony-lint.sh` / `test-*.sh` daemon-spawn load from sibling worktrees on the same host.
- Passes deterministically when run alone on an idle machine and on CI's isolated `ubuntu-latest` runner.

## Fix direction

Replace the sleep-based readiness handshake with a deterministic wait: have the stub write a sentinel file and have the test (or `start-colony.sh`'s pre-flight) poll for `started`/the sentinel with a bounded `until` loop instead of a fixed `sleep 0.5`. That removes the race regardless of host load.

Surfaced while implementing #1000 (dark-factory `gen-agent.sh`); that PR touches none of `start-colony.sh`, daemon spawning, or `test-llm-per-colony.sh`, so this is tracked separately rather than fixed inline.

ylohnitram:
Closes #1008

## Root cause

Both the `t5: --restart-agent` subcase of `tools/test-llm-per-colony.sh` and the `#285` count-invariant subcase of `tools/test-start-colony-restart.sh` flaked under host CPU contention (several `colony-lint.sh` runs executing concurrently across worktrees on one machine). The cause is **host-global shared state in the test fixtures**, not product code.

**`tools/test-llm-per-colony.sh` (the t5 flake in the issue):** the test used a fixed fixture root `/tmp/qa-319-pr1-fed` with a startup `rm -rf` + an EXIT-trap `rm -rf`. When several copies run concurrently, one copy's cleanup wipes another copy's `start-colony.stdout` mid-flight, so `assert_restart_ok` reads an empty file and fails with **empty stdout AND empty stderr** — the exact symptom reported.

I instrumented this to separate the two suspected causes:

| condition | t5 result |
|---|---|
| single isolated run, host load avg 26+ | passes (the `sleep 0.5` + `kill -0 $!` is robust: `$!` is the shim's PID, alive for its whole sleep) |
| 8 runs **sequential** under load | **0/8 fail** |
| 24 runs **concurrent** under load | **16/24 fail** |

So the shared-fixture collision is the dominant cause, not the sleep itself. The issue also asks for a deterministic readiness handshake to remove any residual sleep-vs-sleep race, which is done below.

**`tools/test-start-colony-restart.sh` (the QA-confirmed sibling flake):** the fake-daemon marker was a fixed string (`fake-daemon-inner-for-test-285`). Under concurrency, `pgrep -cf` counted **another** run's fakes (assertion saw `2`) and the cleanup `pkill -f` reaped another run's fakes (`0`). Demonstrated against the pristine `main` version run as 3 concurrent copies under load: the `pgrep` output literally showed a copy in fixture `tmp.ccmldn02R6` counting a fake from a different mktemp dir because the marker string was identical.

## Fix (test-only, no product change)

1. **Readiness race (`tools/test-llm-per-colony.sh` t5):** replaced the fixed `sleep`-based handshake with a deterministic one. The shim touches an `AGENTIS_READY_FILE` sentinel the instant the simulated daemon is up, and `assert_restart_ok` polls that sentinel (plus the `started` line) with a bounded loop and a timeout ceiling instead of relying on a fixed sleep. Removes the race regardless of host load. The printed contract `started <agent> pid=<n> tick=<ms>` is unchanged.

2. **Per-run-unique fixture root (`tools/test-llm-per-colony.sh`):** `FIXTURE_ROOT` is now `mktemp -d` instead of `/tmp/qa-319-pr1-fed`, so two concurrent runs never touch each other's fixtures. EXIT cleanup also `pkill`s this run's shim daemons, scoped to the unique fixture-root token.

3. **Per-run-unique fake-daemon marker (`tools/test-start-colony-restart.sh`):** the marker is now `<stem>-$$-<mktemp-basename>` and threaded through the fake binary name, the `daemon list` `pgrep`, the `exec`, and the count `pgrep`. Two concurrent runs can no longer see each other's fakes.

4. **Same marker-collision fix applied to the two siblings using the identical pattern:** `tools/test-start-colony-multi-repo.sh` (`316m2`) and `tools/test-start-colony-per-repo-labels.sh` (`316m5a`) both used a fixed `pkill -f` marker that reaps a concurrent run's fakes; both now use a per-run-unique marker.

**Product `dev-apprenticeship/triage/scripts/start-colony.sh` is NOT touched.** The instrumentation above proved the `sleep 0.5` pre-flight is not the racing element (an isolated run never failed even under heavy load), so the fix stays entirely in the test tooling — the preferred outcome.

## Under-load before/after proof

Method: 3 concurrent copies per test, 6-8 iterations, with a CPU-burn background load saturating all cores (the issue's "several concurrent colony-lint runs" scenario).

| subcase | before (pristine `main`) | after |
|---|---|---|
| `test-llm-per-colony.sh` t5 (24 concurrent runs) | **16/24 FAIL** (empty stdout AND stderr) | **0/24 FAIL** |
| `test-start-colony-restart.sh` #285 count invariant (18 concurrent runs) | **13/18 FAIL** (`got 2` / `got 0` cross-run contamination) | **0/18 FAIL** |
| combined (24 concurrent copies each, moderate load) | — | t5 0, #285-count 0 |

Sample pre-fix failure showing the cross-run contamination (pristine `main`):
```
[FAIL] triage: iter2 expected exactly 1 live fake, got 2: pgrep=2550199 .../tmp.ccmldn02R6/shim_bg_triage/fake-daemon-i...
[FAIL] triage: iter1 expected exactly 1 live fake, got 0
```

Validation:
- `./tools/colony-lint.sh` from repo root: **363 passed, 0 failed, 5 skipped**.
- `bash -n` clean on all four edited scripts.
- `shellcheck` clean on all four edited scripts.

## Note on a separate, out-of-scope flake

At an artificially extreme load (avg ~32, well beyond the "several concurrent lint runs" scenario), a different subcase of `test-start-colony-restart.sh` — the `pipe-regression` check that runs `python subprocess.run(..., timeout=3)` against a `sleep 5` shim — can hit its hardcoded 3s wall-clock purely from scheduler starvation (`elapsed=3.01`). This subcase is not part of #1008, is unmodified here, and is clean at realistic concurrency (0 failures in the combined run above). Filed #1018 to harden that hardcoded timeout separately rather than expand this PR's scope.

### DIFF
--- a/tools/test-llm-per-colony.sh
+++ b/tools/test-llm-per-colony.sh
@@ -50,10 +50,25 @@ FAIL=0
 pass() { echo "[PASS] $1"; PASS=$((PASS + 1)); }
 fail() { echo "[FAIL] $1${2:+: $2}"; FAIL=$((FAIL + 1)); }
 
-FIXTURE_ROOT="/tmp/qa-319-pr1-fed"
-rm -rf "$FIXTURE_ROOT"
-mkdir -p "$FIXTURE_ROOT"
-trap 'rm -rf "$FIXTURE_ROOT"' EXIT
+# Per-run unique fixture root (#1008). A fixed path like /tmp/qa-319-pr1-fed
+# collides when several copies of this test run concurrently (sibling
+# worktrees on one host under a shared colony-lint pipeline): one copy's
+# startup `rm -rf` or EXIT-trap cleanup wipes another copy's
+# start-colony.stdout mid-flight, so assert_restart_ok reads an empty file
+# and the t5 subcase fails with empty stdout AND stderr. mktemp -d gives
+# each run its own tree, so two concurrent runs never touch each other's
+# fixtures. TMPDIR is honoured when set.
+FIXTURE_ROOT="$(mktemp -d "${TMPDIR:-/tmp}/qa-319-pr1-fed.XXXXXX")"
+# Cleanup reaps both the fixture tree AND any still-sleeping shim daemons
+# this run backgrounded. The shims' argv embeds the unique $FIXTURE_ROOT
+# path, so `pkill -f` matches ONLY this run's processes — a concurrent copy
+# (different mktemp dir) is never hit. This is the #1008 host-global-marker
+# rule applied here: scope every pgrep/pkill to a per-run unique token.
+cleanup() {
+    pkill -f "$FIXTURE_ROOT" 2>/dev/null || true
+    rm -rf "$FIXTURE_ROOT"
+}
+trap cleanup EXIT
 
 # The shim records every `agentis daemon ...` invocation's argv to
 # AGENTIS_RECORD_FILE so the test can grep it for --config-override pairs.
@@ -78,9 +93,19 @@ case "${1:-}" in
         if [ -n "${AGENTIS_RECORD_FILE:-}" ]; then
             printf '%s\n' "$*" >> "$AGENTIS_RECORD_FILE"
         fi
-        # Sleep just long enough for start-colony.sh's `sleep 0.5` +
-        # `kill -0` liveness probe to see us as alive.
-        sleep 1
+        # #1008: deterministic readiness, not a sleep-vs-sleep race. The
+        # OLD shim slept a fixed 1s "just long enough for start-colony.sh's
+        # `sleep 0.5`"; under host CPU contention either side's scheduling
+        # could slip and start-colony.sh's `kill -0` probe could miss us, so
+        # no `started` line was printed and the harness read empty stdout.
+        # Instead, drop a sentinel the instant we're up (so the harness can
+        # poll for genuine readiness) and then stay alive long enough for the
+        # liveness probe to observe us regardless of load. AGENTIS_READY_FILE
+        # is exported by run_for_fixture.
+        if [ -n "${AGENTIS_READY_FILE:-}" ]; then
+            : > "$AGENTIS_READY_FILE"
+        fi
+        sleep 5
         exit 0
         ;;
     memo)
@@ -149,12 +174,17 @@ run_for_fixture() {
     local tag="$1"
     local shim_dir="$FIXTURE_ROOT/$tag/shim"
     local record_file="$FIXTURE_ROOT/$tag/agentis-argv.log"
+    local ready_file="$FIXTURE_ROOT/$tag/daemon-ready"
     local fed="$FIXTURE_ROOT/$tag/dev-apprenticeship"
     build_shim "$shim_dir" "$record_file"
     : > "$record_file"
+    rm -f "$ready_file"
     # Restrict PATH to shim + minimum so the real agentis cannot leak in.
+    # AGENTIS_READY_FILE (#1008) is the deterministic readiness sentinel the
+    # shim touches the instant the simulated daemon is up.
     PATH="$shim_dir:/usr/bin:/bin" \
         AGENTIS_RECORD_FILE="$record_file" \
+        AGENTIS_READY_FILE="$ready_file" \
         bash "$fed/triage/scripts/start-colony.sh" \
             --restart-agent labeler \
             "$fed/triage/config/colony.toml" \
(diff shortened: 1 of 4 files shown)


ylohnitram:
## QA Report — PR #1017 (test-isolation fix, Closes #1008)

Verified in a worktree on a 12-core host under a 10× `yes` CPU-burn (load 9-12); no git stash, main untouched.

| headline t5 fix under load | ✅ | — | `test-llm-per-colony.sh` t5: pristine main **16/24 FAIL** (empty stdout+stderr symptom) → branch **0/24**. Matches PR's claim. |
| #285 count-invariant fix under load | ✅ | — | `test-start-colony-restart.sh`: main flaked **3/48** at 6 copies (cross-run `got 0`/`got 2`) → branch **0/48**. |
| combined under-load (both tests) | ✅ | — | 18 t5 + 18 restart runs under burn → **0 failures** on branch |
| test-ONLY scope | ✅ | — | diff = exactly the 4 `tools/*.sh`; no product/colony code |
| product `start-colony.sh` + `started … pid=… tick=…` contract | ✅ | — | byte-identical to main (not in diff) |
| mktemp FIXTURE_ROOT (no shared `/tmp/qa-319-pr1-fed`) | ✅ | — | per-run unique, TMPDIR-honoring; fixed path gone |
| bounded sentinel-poll readiness | ✅ | — | 10s ceiling, dual signal (ready-file + `^started`); replaces fixed sleep |
| per-run-unique markers threaded | ✅ | — | `<stem>-$$-<mktemp>` through fake-name + pgrep + pkill + exec across all 4 scripts; no real-daemon collision; no process leaks |
| `./tools/colony-lint.sh` + shellcheck | ✅ | — | 363 passed / 0 failed / 5 skipped; bash -n + shellcheck clean on all 4 |
| #1018 out of scope | ✅ | — | pipe-regression subcase untouched (only adjacent context lines); 0 pipe failures on branch; tracked separately in #1018 |
| #285 needed 6 copies (not 3) to flake on main | ⚠️ | **non-blocker — load-profile note** | the PR's stated 13/18 main-flake rate is host-load-specific; the fixed-marker collision is unambiguous on inspection + reproduces, and the after-state is clean (0) — which is what matters |

### Verdict: ✅ ship-it
Root cause = host-global shared fixture state (fixed FIXTURE_ROOT + fixed fake-daemon marker); the fix eliminates it — main flakes under concurrent load, the branch is **0** across every stress config. Strictly test-only (product script + contract byte-identical). Mechanism sound (mktemp isolation + bounded dual-signal poll + per-run-unique markers). colony-lint 363/0/5. #1018 correctly carved out. The one ⚠️ is a load-profile observation, not a defect. Merging.

PR #1021, opened 2026-06-14
fix(tools): de-flake pipe-regression subcase under extreme host load (#1018)

### LINKED ISSUE #1018: Flaky: tools/test-start-colony-restart.sh pipe-regression subcase hits hardcoded subprocess.run(timeout=3) under extreme host load
## Symptom

The `pipe-regression` subcase of `tools/test-start-colony-restart.sh` (one per colony: `restart returns promptly under subprocess.run(capture_output=True)`) can fail under **extreme** host CPU contention:

```
[FAIL] implementation: subprocess.run(capture_output=True) hangs or fails — daemon stdio likely inherited: body=TIMEOUT elapsed=3.01 stdout=''
```

## Root cause (test fixture, not product code)

The subcase launches `start-colony.sh --restart-agent` via a Python `subprocess.run(..., timeout=3)` against a shim whose daemon arm `sleep 5`s. A correctly-detached daemon lets the call return in ~0.6s; a pipe-inheriting one hangs to the 3s timeout. Under a saturated host (observed at load avg ~32 on a 12-core box) the OS cannot schedule the fork/exec/print of the *correctly-detached* path within 3s either, so the wall-clock crosses the hardcoded `timeout=3` and the assertion reports a spurious `TIMEOUT` even though stdio detachment is fine (`elapsed=3.01`).

## Evidence it is environmental + out of scope for #1008

- Surfaced while fixing #1008 (PR #1017). That PR fixes the marker-collision and readiness-race flakes; this subcase is untouched and orthogonal.
- Clean at realistic concurrency: in PR #1017's combined under-load proof (3 concurrent copies, moderate load avg ~9-12) this subcase had **0 failures**. It only flaked at the artificially extreme load avg ~32.

## Fix direction

Widen the gap between the "detached = fast" and "inherited = hangs" signals so scheduler jitter cannot cross the boundary: raise the shim sleep well above any plausible scheduling delay and the Python `timeout` to a value that load-induced jitter never reaches while staying comfortably below the shim sleep. Keep the assertion semantics (fast return vs hang) intact.

ylohnitram:
## What

Fixes the `pipe-regression` flake in `tools/test-start-colony-restart.sh` (#1018). Test-fixture-only change — no product code, no other subcase touched.

## Root cause

The `pipe-regression` subcase stands up a shim `agentis` whose daemon arm `sleep`s, then calls `start-colony.sh --restart-agent` through `subprocess.run(capture_output=True, timeout=N)`. A correctly-**detached** daemon returns in ~0.6s; a pipe-**inheriting** one blocks Python on the captured pipes until the shim exits, so the timeout fires first → `TimeoutExpired`. With `sleep 5` / `timeout=3` the detached-vs-inherited gap was only ~2.4s, so under extreme host load (~32 on a 12-core box, per the issue) scheduling jitter pushed the *detached* path's fork/exec/print past 3s too → spurious `TIMEOUT elapsed=3.01`.

## Fix

Widen the gap so jitter can't cross it:

| Knob | Before | After |
|------|--------|-------|
| pipe-regression shim `sleep` | `5` | `30` |
| Python `subprocess.run(timeout=)` | `3` | `15` |

`timeout (15) < shim_sleep (30)` is preserved, so a genuinely pipe-inheriting daemon still blocks the full 30s and trips the 15s timeout — assertion semantics (`OK rc=0` + `started …` → pass; `TimeoutExpired` → fail) unchanged. The detached path now has ~14s of jitter headroom (≈6× the observed ~3s jitter). No happy-path slowdown: the call returns as soon as `start-colony.sh` does; `timeout` is only the ceiling. A comment documents the margin so it isn't tightened back.

## Verification

- `tools/test-start-colony-restart.sh` → **26 passed / 0 failed** at normal load (all 5 pipe-regression subcases green).
- **Under-load smoke:** 24 CPU burners on a 12-core box (load avg ~6.8, ~25 runnable) → still **26/0**, every pipe-regression subcase PASS, **no `TIMEOUT`** (all detached returns well under 15s).
- `bash -n` clean; `colony-lint.sh` → 0 failures.
- No CHANGELOG entry — `tools/test-*.sh` is contributor tooling, not a versioned federation component.

Closes #1018.

### DIFF
--- a/tools/test-start-colony-restart.sh
+++ b/tools/test-start-colony-restart.sh
@@ -171,6 +171,11 @@ test_one_colony() {
     # every /restart. This test uses a shim that lives LONGER than the
     # subprocess.run timeout; if stdio is detached from the inherited
     # pipes the script returns in ~0.6s, otherwise Python hits TimeoutExpired.
+    # The shim sleep (30s) and the timeout (15s) are deliberately far apart:
+    # the detached path returns in ~0.6s, so 15s leaves ~14s of headroom for
+    # scheduling jitter under extreme host load (#1018 saw a 0.6s op pushed past
+    # 3s at load avg ~32), while an inherited pipe still blocks the full 30s and
+    # trips the 15s timeout. Do NOT tighten these toward each other.
     out="$TMPDIR_TEST/$colony.pipe.log"
     SHIM_PIPE_DIR="$TMPDIR_TEST/shim_pipe_$colony"
     mkdir -p "$SHIM_PIPE_DIR"
@@ -184,7 +189,7 @@ if [ "${1:-}" = "daemon" ] && [ "${2:-}" = "list" ]; then
     printf '[]\n'
     exit 0
 fi
-sleep 5
+sleep 30
 exit 0
 SHIM
     chmod +x "$SHIM_PIPE_DIR/agentis"
@@ -194,7 +199,7 @@ start = time.time()
 try:
     r = subprocess.run(
         ['bash', sys.argv[1], '--restart-agent', sys.argv[2], sys.argv[3]],
-        capture_output=True, text=True, timeout=3,
+        capture_output=True, text=True, timeout=15,
     )
     elapsed = time.time() - start
     print(f'OK rc={r.returncode} elapsed={elapsed:.2f} stdout={r.stdout.strip()!r}')


ylohnitram:
# QA — PR #1021: de-flake pipe-regression subcase (#1018)

Changed file: `tools/test-start-colony-restart.sh` (+7 / −2). Two numbers + a rationale comment.

| Diff scope = exactly the 2 numbers + comment | ✅ pass | — | `git diff` shows only the pipe-regression shim `sleep 5→30` (line 192), the `subprocess.run` `timeout=3→15` (line 202), and a 5-line rationale comment. Nothing else touched. |
| Numbers changed in the RIGHT places | ✅ pass | — | Happy-path shim `sleep 1` (line 84) UNTOUCHED. `test_no_duplicate_on_restart` fake-daemon `sleep 60` (line 259) and `sleep 0.3` settle (line 362) UNTOUCHED. Only `timeout=` is the one inside the pipe-regression `subprocess.run`. |
| Detection invariant `timeout(15) < shim_sleep(30)` | ✅ pass | — | Gap preserved (15 < 30). A pipe-inheriting daemon would block Python on the captured pipe for the full 30s, tripping the 15s timeout → `TimeoutExpired` → `fail` branch. The assertion is NOT neutered into always-passing; it still requires `OK rc=0` + the `started <agent> pid=<n> tick=<ms>` line. |
| No happy-path slowdown / no semantic change | ✅ pass | — | Pass condition (`^OK rc=0` + `started ... pid=... tick=...`) and fail condition (`TimeoutExpired`) unchanged. `timeout` is a ceiling only; detached path returns in ~0.6–0.85s. |
| Full test run | ✅ pass | — | `bash tools/test-start-colony-restart.sh` → **26 passed, 0 failed**, exit 0, ~21s wall. All 5 colonies' pipe-regression subcases pass. |
| `bash -n` | ✅ pass | — | Clean. |
| `colony-lint.sh` | ✅ pass | — | **364 passed, 0 failed, 5 skipped** (shellcheck-absent skips). `test-start-colony-restart.sh` unit-test entry passes. |
| Under-load re-run | ✅ pass | — | 8 self-terminating CPU burners (12-core box, loadavg ~5). All 5 pipe-regression subcases still pass; measured detached elapsed = **0.85s** (≈14s headroom under the 15s ceiling). All burners killed afterward; host confirmed clean. |
| Long-sleeping shim cleanup (line 220 `pkill`) | ⚠️ caveat | Pre-existing defect, **not introduced** by this PR; cleanup is ineffective but harmless (self-terminating orphans). **Tracked as follow-up #1022** (make the shim reap its own sleep on signal). Non-blocking for this PR. | After a run, `pgrep -af 'sleep 30'` shows up to ~3 orphaned `sleep 30` children reparented to `systemd --user`. The cleanup `pkill -f "$SHIM_PIPE_DIR/agentis"` never matches because (a) the shim is invoked as bare PATH-resolved `agentis` so its argv is `agentis daemon …`, not the full shim path, and (b) the actual long-lived process is a separate `sleep 30` child with argv `sleep 30`. The EXIT-trap `cleanup()` only `pkill`s `$RUN_MARKER`, which the pipe shim does not carry. Orphans self-terminate when their own `sleep` expires. **This PR widens the linger window from ~5s to ~30s** (6×) — strictly worse, though still transient and process-count-bounded (≤5, one per colony). |
| Cross-colony interference from longer-lived orphans | ✅ pass | — | The pipe-regression assertion reads only its own `py_out`; `test_no_duplicate_on_restart` counts only `$RUN_MARKER` argv patterns. A stale `sleep 30` from an earlier colony cannot be miscounted or cause a false pass/fail in a later subcase. |

## Detail — orphaned-sleep mechanism (the one ⚠️)
- start-colony.sh `--restart-agent` backgrounds `agentis daemon … </dev/null >/dev/null 2>&1 &` (line 413–418). The shim's `agentis` runs `sleep 30; exit 0`; bash waits on the sleep. When the parent returns (~0.6s) and Python's `subprocess.run` completes, the shim+`sleep 30` are orphaned to init/systemd.
- `pkill -f "$SHIM_PIPE_DIR/agentis"` (line 220) matches neither the shim process (argv = `agentis daemon …`) nor the `sleep 30` child (argv = `sleep 30`). Verified: after a run `pgrep -af 'shim_pipe'` = NONE while `pgrep -af 'sleep 30'` = several.
- Impact bounded: ≤5 orphans (one per colony), each self-terminating within 30s of its own spawn. No permanent leak, no PID exhaustion, no test-correctness effect. Same `pkill` line + ineffectiveness exists on `origin/main` (only the duration grows 5→30 here).

**Verdict:** ✅ ship-it — the fix is correctly placed, preserves the `timeout<sleep` detection invariant, passes all 26 assertions + 364 lint checks clean (also under simulated load with ~14s headroom); the only caveat is a pre-existing, harmless orphaned-`sleep` cleanup gap that this PR lengthens but does not create — track it as a non-blocking follow-up.

PR #1023, opened 2026-06-14
fix(tools): pipe-regression shim reaps its own sleep child (#1022)

### LINKED ISSUE #1022: pipe-regression test shim leaks orphaned sleep children (cleanup doesn't reap)
## Problem

The `pipe-regression` subcase of `tools/test-start-colony-restart.sh` launches a shim `agentis` whose daemon arm `sleep`s longer than the `subprocess.run` timeout (the detached-vs-inherited signal). The per-colony cleanup `pkill -f "$SHIM_PIPE_DIR/agentis"` (and the EXIT-trap `cleanup()`, which only pkills `$RUN_MARKER`) does **not** reliably reap the long-lived sleep:

- the detached daemon is invoked as a bare PATH-resolved `agentis`, so its cmdline may not contain `$SHIM_PIPE_DIR`; and
- the actually-long-lived process is a separate `sleep` **child** (argv `sleep <N>`), which `pkill -f "$SHIM_PIPE_DIR/agentis"` never matches.

So after a run, up to ~5 `sleep` children (one per colony) are orphaned to `systemd --user` until they self-terminate.

## Impact

Harmless today: the orphans are count-bounded (≤ one per colony), self-terminate when the sleep elapses, and don't interfere across colonies (each colony's assertion has already completed). Pre-existing — not introduced by #1018/#1021, which only lengthened the per-orphan linger window (the shim sleep went 5s → 30s while widening the de-flake margin).

## Fix direction

Make the shim reap its own sleep on signal (e.g. `sleep "$N" & trap 'kill "$!" 2>/dev/null' TERM INT; wait`) so the existing `pkill` is effective, **and/or** confirm what cmdline the detached daemon actually exposes and pkill on a pattern that matches it. Verify no orphaned `sleep` survives a full run (`pgrep`-assert after the suite).

ylohnitram:
## What

Fixes #1022 — the `pipe-regression` subcase of `tools/test-start-colony-restart.sh` leaked orphaned `sleep` children after each run. Test-fixture-only change.

## Root cause (corrected from a /proc probe)

The cleanup `pkill -f "$SHIM_PIPE_DIR/agentis"` **does** reach the shim: a `#!/bin/bash` shim's detached daemon has cmdline `/bin/bash <SHIM_PIPE_DIR>/agentis daemon …`, which carries the full path. But the long-lived process is the shim's **`sleep` child** (argv `sleep 30`, no shim path). Killing the bash parent left the `sleep` reparented to `systemd --user`, running out its remaining seconds — up to one orphan per colony per run. (Made more visible after #1021 widened the shim sleep 5→30.)

## Fix

Background the sleep and trap `TERM`/`INT` to kill it, so the existing cleanup reaps the whole process tree:

```bash
sleep 30 &
_sleep_pid=$!
trap 'kill "$_sleep_pid" 2>/dev/null' TERM INT
wait "$_sleep_pid"
exit 0
```

No change to the cleanup line — it already matches the bash shim; the trap propagates the kill to the child. The shim still `wait`s the full 30s, so the detached-vs-inherited detection (an inherited pipe blocks past the 15s `subprocess.run` timeout) is unchanged.

## Verification

- **/proc probe:** after the test's exact `pkill -f "$SHIM_PIPE_DIR/agentis"`, **neither** the bash shim nor its `sleep` child survives (`ps` → none). Without the trap the `sleep 30` lingers.
- `tools/test-start-colony-restart.sh` → **26 passed / 0 failed**; `pgrep -af 'sleep 30'` shows **no orphan** after the run completes.
- `bash -n` clean; `colony-lint.sh` → **364 / 0 / 5**.
- No CHANGELOG entry — contributor test tooling, not a versioned federation component.

Closes #1022.

### DIFF
--- a/tools/test-start-colony-restart.sh
+++ b/tools/test-start-colony-restart.sh
@@ -189,7 +189,17 @@ if [ "${1:-}" = "daemon" ] && [ "${2:-}" = "list" ]; then
     printf '[]\n'
     exit 0
 fi
-sleep 30
+# Background the long sleep and reap it on signal. The cleanup `pkill -f
+# "$SHIM_PIPE_DIR/agentis"` (below) reaches THIS bash process (its cmdline
+# carries the full shim path), but a foreground `sleep 30` would be a separate
+# child (argv `sleep 30`, no shim path) that pkill can't match and that would
+# linger, orphaned to systemd, for its remaining seconds (#1022). The trap
+# propagates the kill to the child. We still `wait` the full duration, so a
+# genuinely pipe-inheriting daemon still trips the subprocess.run timeout.
+sleep 30 &
+_sleep_pid=$!
+trap 'kill "$_sleep_pid" 2>/dev/null' TERM INT
+wait "$_sleep_pid"
 exit 0
 SHIM
     chmod +x "$SHIM_PIPE_DIR/agentis"


ylohnitram:
# QA Report — PR #1023: pipe-regression shim reaps its own sleep child (#1022)

Scope: single file changed — `tools/test-start-colony-restart.sh` (+11/-1). The
`pipe-regression` shim's foreground `sleep 30` is backgrounded with a
`trap ... TERM INT` that kills the child, so the per-colony cleanup `pkill`
no longer leaves an orphaned sleep reparented to systemd.

| Orphan reaping (PR) | ✅ | — | Full suite run, then exact-argv sweep: **0** `sleep 30`, **0** `shim_pipe*/agentis` survivors. |
| Orphan leak reproduced on main | ✅ | — | Ran `origin/main` script in-place: leaked **5** `sleep 30` orphans (one per colony, PPID=reparented). PR fixes it to 0. Confirms the change is a real fix, not a no-op. |
| Detection invariant intact | ✅ | — | Isolated shim blocks the **full** duration under `wait` and exits 0 when unsignalled (measured 3.01s for a 3s stand-in). A genuinely pipe-inheriting daemon still blocks 30s > 15s `subprocess.run` timeout, so the assertion still catches a real regression. |
| Trap correctness | ✅ | — | `_sleep_pid=$!` captures the bg child; child argv is bare `sleep N` (the exact process `pkill -f .../agentis` cannot match). SIGTERM to the bash shim fires the trap → `kill "$_sleep_pid"` → **0** child survivors, **0** shim survivors. `kill` is guarded `2>/dev/null`. |
| `set -e` / nonzero-`wait` safety | ✅ | — | Pipe shim has **no** `set -e`; a SIGTERM-interrupted `wait` cannot abort it. Caller (`start-colony.sh`) launches the daemon backgrounded `</dev/null >/dev/null 2>&1 &`, then only checks `kill -0` liveness and prints `started ...` — it never reads the shim's exit code, so a 143 exit is irrelevant. |
| Full suite | ✅ | — | `bash tools/test-start-colony-restart.sh` → **26 passed, 0 failed**, rc=0, ~29s wall. |
| `bash -n` | ✅ | — | Clean. |
| `colony-lint.sh` | ✅ | — | **364 passed, 0 failed, 5 skipped** (rc=0). Includes `test-start-colony-restart.sh` as a sub-suite (PASS). |
| Suite self-cleanup guard | ✅ | — | Per-colony `pkill -f "$SHIM_PIPE_DIR/agentis" 2>/dev/null \|\| true` (line 230) unchanged; EXIT trap cleanup unaffected. |
| Diff hygiene | ✅ | — | Diff is exactly the documented change; no other behavior touched. Working tree restored pristine after probes (`git status` clean). |
| Trap-install race (theoretical) | ⚠️ | None needed — not reachable in this test. Worth a one-line comment if ever revisited. | Microsecond window between `sleep 30 &` and the `trap` line where a SIGTERM would orphan the child. In the actual flow the `pkill` fires ~0.6s+ after launch, so the trap is always installed first. Not a user-facing leak. |

## Observed numbers
- Suite: **26 passed / 0 failed** (rc=0, ~29s).
- Post-run orphans (PR): **0** sleep30, **0** shim_pipe.
- Same metric on `origin/main`: **5** orphaned `sleep 30`.
- `colony-lint.sh`: **364 / 0 / 5**.

## Counts
✅ 10  ⚠️ 1  ❌ 0

**Verdict:** ✅ ship-it — the fix demonstrably eliminates the 5-per-run orphan leak (main → 0) while preserving the 30s-block detection invariant and introducing no `set -e`/exit-code regression.

Issue #46, opened 2026-04-11
docs: federation status table marks implementation and release colonies as Active but agents/ dirs are empty


ylohnitram:
## Problem

The federation README at `dev-apprenticeship/README.md` has a colony status table that lists all 5 colonies as **Active**:

\`\`\`
| [code-review](./code-review/) | ... (5 agents) | Active |
| [planning](./planning/) | ... (4 agents) | Active |
| [implementation](./implementation/) | ... (4 agents) | Active |
| [triage](./triage/) | ... (4 agents) | Active |
| [release](./release/) | ... (4 agents) | Active |
\`\`\`

But only three of those colonies actually have `.ag` files today. `implementation/agents/` and `release/agents/` contain only `.gitkeep`.

This is not a functional bug and it is already tracked at a higher level by issue #5 (overall implementation plan), but the status table is a user-facing signal, so a fresh operator reading the federation README would expect 21 active agents and find 13.

## Surfaced during

PR #45 QA. That PR intentionally scopes the new Bootstrap section to only the three implemented colonies (triage, code-review, planning), which is the correct behavior, but exposes the inconsistency in the status table above.

## Proposal

Two reasonable fixes, either is fine:

1. **Flip status to Scaffolded or Planned** for implementation and release until their agents land. Matches reality.
2. **Add a status note** under the table: "implementation and release have scaffolding (config, scripts) but no agent files yet, see #5 for roadmap".

Option 1 is cleaner. Option 2 keeps the table optimistic for roadmap purposes.

## Acceptance

- Federation README status table accurately reflects which colonies have `.ag` files today
- No functional changes, docs only

PR #1050, opened 2026-06-15
dark-factory: reject stub harness + fix flaky LLM delivery at the source (flat-cyborg v0.10.0)

ylohnitram:
Hardens autonomous harness generation (`run-autoharness.sh`) after a live end-to-end run on a real ERC4626 vault (sDAI) exposed two gaps, and moves the delivery fix from a shell workaround into the flat-cyborg tool itself.

### 1. Reject a vacuous stub harness (commit 1)
The flat-cyborg LLM backend occasionally returned a degenerate ~27-line stub (a `test_sanity()` asserting `1+1==2`, no fork, no fuzz, no invariant) that compiled and so was mistaken for a ready harness — hunting nothing. `real_harness()` now gates on three structural requirements: a real fork (`createSelectFork`), a `testFuzz_*(uint256 ...)` fuzz entrypoint, and a `require()` invariant. A file missing any of them is sent back to the compile-repair loop with the full structural requirement re-stated, so the LLM repairs toward a real fork-fuzz harness instead of degenerating into a constant-asserting stub.

### 2. Fix flaky delivery at the source, drop the shell workaround (commit 2)
Driving the interactive claude CLI over a PTY surfaced two delivery/timing failures that the script had been papering over with a fold-the-prompt + retry-on-empty workaround:
- an ultra-long single-line instruction block overflowed claude's editor input;
- a slow first reply (model emits startup chrome / thinks before answering) was captured as empty.

Both are now fixed in flat-cyborg v0.10.0 (`--wrap-input` folds long lines; `--extract` is gated on the reply sentinel). `flat-cyborg-claude.sh` passes `--wrap-input 72` and the floor is bumped to v0.10.0, so `gen()` collapses back to a single backend call. A rare empty extraction is still handled by the compile/`real_harness` repair loop.

## Validation
- `colony-lint.sh`: 367 passed, 0 failed, 5 skipped.
- Both scripts pass `sh -n` (POSIX/dash-safe).
- CHANGELOG updated (`### Changed`, flat-cyborg floor → v0.10.0).

Upstream tool fix: flat-cyborg #46 / #47, released as v0.10.0.

### DIFF
--- a/dark-factory/CHANGELOG.md
+++ b/dark-factory/CHANGELOG.md
@@ -33,6 +33,14 @@ Every release declares its runtime floor as `**Requires:** agentis >= X.Y.Z`.
   `flat-cyborg` backend branch previously set only the timeout and left `llm.command`
   unset. `--backend claude` (metered `claude -p`) stays an explicit opt-in. Requires
   flat-cyborg >= v0.9.1 (`--extract` implies the screen grid).
+- `run-autoharness.sh`: **rejects a vacuous stub harness**. A generated file must carry a real fork
+  (`createSelectFork`), a `testFuzz_*(uint256 ...)` fuzz entrypoint, and a `require()` invariant, or it is
+  sent back to the compile-repair loop with the full structural requirement re-stated — a live sDAI run
+  exposed the LLM occasionally returning a degenerate `1+1==2` sanity test that compiled but hunted nothing.
+  The shell-side prompt-fold + empty-retry workaround is removed now that delivery is fixed at the source:
+  `flat-cyborg-claude.sh` passes `--wrap-input 72` (folds the long instruction block so it no longer
+  overflows claude's editor) and flat-cyborg >= v0.10.0 gates `--extract` on the reply sentinel (a slow
+  first reply is no longer captured as empty). Bumps the flat-cyborg floor to **v0.10.0**.
 
 
 ### Added
--- a/dark-factory/flat-cyborg-claude.sh
+++ b/dark-factory/flat-cyborg-claude.sh
@@ -12,7 +12,10 @@
 set -eu
 PROMPT="${1:-}"
 if [ -z "$PROMPT" ]; then PROMPT="$(cat)"; fi
-exec flat-cyborg --extract --no-jitter --auto-approve \
+# --wrap-input 72: fold the (often single-line, ~700-char) instruction block so it
+# does not overflow claude's editor input. flat-cyborg >=0.10.0 also gates --extract
+# on the reply sentinel, so a slow first reply is no longer captured as empty.
+exec flat-cyborg --extract --no-jitter --auto-approve --wrap-input 72 \
   --idle-ms "${FLAT_CYBORG_IDLE_MS:-8000}" \
   --timeout-ms "${FLAT_CYBORG_TIMEOUT_MS:-180000}" \
   --cmd "$PROMPT" -- claude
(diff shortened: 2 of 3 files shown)


ylohnitram:
## QA — PR #1050 (`fix/autoharness-reject-stub`)

Scope: `dark-factory/run-autoharness.sh`, `dark-factory/flat-cyborg-claude.sh`, `dark-factory/CHANGELOG.md`. All checks below were run against the branch head.

| `sh -n` (dash path) + `bash -n` — both scripts | ✅ | — | Both parse clean under `sh` and `bash`. |
| `./tools/colony-lint.sh` | ✅ | — | 367 passed, 0 failed, 5 skipped. The single `[WARN]` is the pre-existing check-secrets warning-mode banner, not a finding in this diff. |
| `real_harness()` gate — accepts a real fork-fuzz harness | ✅ | — | Fixture with `createSelectFork` + `function testFuzz_hunt(uint256 seed)` + `require(` is ACCEPTED. |
| `real_harness()` gate — rejects a `1+1==2` sanity stub | ✅ | — | A bare `test_sanity()` asserting `1+1==2` (no fork, no fuzz fn, no require) is REJECTED, as is a near-miss with `require` but no fork/fuzz fn. |
| Repair loop re-states full structural requirement + bounds attempts | ✅ | — | Reject prompt lists all 3 markers and an explicit "a test that only asserts constants (e.g. 1+1==2) … is WRONG"; loop is bounded by `REPAIRS` (default 6), `exit 1` once `ATT > REPAIRS`. |
| `gen()` both call sites pass prompt via stdin | ✅ | — | `gen "$PROMPT"` and `gen "$WORK/repair.txt"` both feed the file on stdin (`"$WRAP" < "$1"`); pipeline is byte-identical to the prior inline command. |
| Removing the dedicated retry is safe | ✅ | — | Empty/bad output is re-generated by the compile/`real_harness` repair loop; no behavioral gap. |
| `set -u` safety — no new unbound vars; old `_g`/`_i`/`GEN_RETRIES` gone | ✅ | — | `gen()` references only `$1` (always passed) and `$WRAP` (set at top). The removed locals appear nowhere; `--help`, no-args, and trailing-value-flag paths all exit cleanly (2/0) with no unbound-var crash. |
| `flat-cyborg-claude.sh` — `--wrap-input 72` placement + stdin fallback + exec | ✅ | — | Flag sits before the value-taking opts and `--cmd … -- claude`; full invocation parses (verified live). Prompt still read from `$1`-or-stdin, then `exec flat-cyborg`. |
| flat-cyborg floor: `version` reports 0.10.0 and `--help` shows `--wrap-input` | ✅ | — | `flat-cyborg 0.10.0` on PATH; `--wrap-input <COLS>` present in help. |
| CHANGELOG `### Changed` bullet | ✅ | — | New bullet documents the stub rejection, the workaround removal, `--wrap-input 72`, and the floor bump to v0.10.0. (The duplicate `### Added` heading in `[Unreleased]` pre-exists on main; untouched by this PR.) |
| Secrets / internal-path / private-repo scan of the diff | ✅ | — | No keys/tokens, no internal absolute paths, no private-repo refs in added lines. `0x7109709E…` is the standard Foundry VM cheatcode address. |
| `testFuzz…(uint256 …)` regex matches only same-line signatures | ⚠️ | Dismissed — acceptable | The grep requires `function testFuzz`, `(`, and `uint256` on one line. A multi-line signature, or a seed typed `uint128`, would be rejected and re-repaired. The generation prompt explicitly dictates `testFuzz_hunt(uint256 seed)` on one line and Foundry fuzz entrypoints conventionally take `uint256`, so this gates correctly in practice. |
| `gen()` awk keeps trailing prose after a closing fence | ⚠️ | Dismissed — not a regression | After the closing ``` ``` ```, residual prose (e.g. "Done!") is kept. This is byte-identical to the prior inline pipeline (not introduced here), and stray trailing tokens are caught by the `forge build` compile gate, which triggers a re-generate. |

### Verdict: ✅ ship-it

The core change is sound: the `real_harness()` gate provably accepts a genuine fork-fuzz harness and rejects a vacuous `1+1==2` stub, the repair loop re-states the full structural contract and is bounded, the `gen()` simplification preserves the exact extraction pipeline while removing the now-redundant shell-side retry (delivery is fixed at the source via flat-cyborg v0.10.0 + `--wrap-input 72`), and both scripts pass `sh`/`bash` syntax and full colony-lint (0 failures). The two ⚠️ rows are pre-existing/benign edge cases dismissed with rationale, not blockers.

PR #1061, opened 2026-06-15
fix(dashboard): pin collector to a single injected now — kill 00:00 UTC date-boundary flake (#1043)

### LINKED ISSUE #1043: test-sse-stream.sh + test-timeline-rendering.sh flake at the 00:00 UTC date boundary (block every PR that straddles midnight)
## Symptom

On PR #1042's CI run (`27516183561`), `colony-lint` failed with **two** test failures, both **unrelated to the PR's diff** (which was 100% under `dark-factory/`):

- `[FAIL] tools: test-sse-stream.sh unit tests`
- `[FAIL] 8: test-timeline-rendering.sh failed under the refactor … Results: 58 passed, 1 failed` (test 64, the #369 Promote-Candidates cap)

A plain **re-run of the same commit passed green** with no code change. The failing run's log timestamps cluster at **`2026-06-15T00:00:33Z` … `00:00:55Z`** — i.e. the tests executed across the **00:00 UTC date boundary**. `test-timeline-rendering.sh` renders timeline entries with relative/absolute timestamps, and `test-sse-stream.sh` is timing-sensitive; a midnight/date-rollover assumption (a "today" vs "yesterday" bucket, a same-day relative-time string, or a clock-tick race) is the likely cause.

## Why it matters

A test that flips red purely because it ran at midnight UTC blocks every PR whose CI happens to straddle 00:00, and trains reviewers to "just re-run" — which erodes the red-CI gate. This is a latent determinism bug, not a one-off.

## Ask

- Make `test-sse-stream.sh` and `test-timeline-rendering.sh` **deterministic across the date boundary**: pin the clock (inject a fixed `now`/epoch the renderer + tests share), or assert on stable relative buckets rather than wall-clock-derived day strings.
- Add a CI guard or a test that runs the suite with a **simulated 23:59:59 → 00:00:01 rollover** so the regression can't return silently.

Found while merging #1041 FM1 (the diff did not touch `tools/` or `federation-dashboard/`). Not a blocker for that PR (a re-run was clean), filed so the flake is fixed at the root rather than re-run away.

ylohnitram:
Closes #1043.

## Root cause
`federation-dashboard-collector.py` already gets the caller's "now" as `epoch` (argv[4]) and derives the cost day-bucket from it — but **5 sites still call `time.time()` again** (startup-grace, two orphan checks, sidecar health, sidecar-now). That second wall-clock sample can disagree with `epoch` across the 00:00 UTC boundary, flipping a startup-grace/orphan flag → the timeline tests assert on that JSON and flake. `test-sse-stream.sh` test 8 runs `test-timeline-rendering.sh`, so it goes red in the same midnight run (one cause, not two).

## Fix
- `NOW_TS = epoch` — the single injected now; all 5 now-reads route through it. No second wall-clock sample, so the date boundary can't desync them. The cost day-bucket already used the injected epoch; this finishes the job.
- **Source-guard test 65** in `test-timeline-rendering.sh`: asserts the collector contains no bare `time.time()` (all now-derived from the injected epoch). The "CI guard so the regression can't return silently" the issue asks for — deterministic, no simulated rollover.
- `federation-dashboard/CHANGELOG.md` `[Unreleased]` entry (no VERSION bump).

Out of scope (checked): `auto-promote-decisions.py`'s `time.time()` reads are **durations** (freshness/runtime elapsed seconds), monotonic across midnight — not date-boundary-sensitive.

## Validation
- `tools/test-timeline-rendering.sh`: 60 passed, 0 failed (incl. new test 65).
- `tools/test-sse-stream.sh`: 9 passed, 0 failed (incl. test 8 running the timeline suite).
- `tools/colony-lint.sh`: 367 passed, 0 failed, 5 skipped.


ylohnitram:
## QA Report — PR #1061 (`fix/1043-pin-collector-now`)

Closes #1043 (00:00 UTC date-boundary flake). Verified against `federation-dashboard/lib/federation-dashboard-collector.py`, `tools/test-timeline-rendering.sh`, and `tools/auto-promote-decisions.py`.

| Zero `time.time()` in collector | ✅ | — | `grep -n 'time\.time()' federation-dashboard/lib/federation-dashboard-collector.py` → no match (exit 1). Diff is exactly 5 `int(time.time())` → `NOW_TS` swaps + the comment block; no other logic touched. |
| `NOW_TS = epoch`, `epoch = int(sys.argv[4])` | ✅ | — | collector L74 `epoch = int(sys.argv[4])`, L81 `NOW_TS = epoch`, defined module-level before all 5 uses. |
| 5 former now-reads all use `NOW_TS` | ✅ | — | startup-grace L968, orphan L977, cost-cap startup-grace L1866, cost-cap orphan L1873, sidecar-now L1921 — all `= NOW_TS`. |
| Semantic correctness (now-vs-stored, not in-run duration) | ✅ | — | All 5 are `now − stored_ts < threshold`: L969-970 `(now_ts − started_at_ts) < (interval_s+120)`; L978 `(now_ts_orphan − last_tick_ts) < ORPHAN_INFER_S`; L1867-1868 + L1874 mirror for cost-cap; L1921 feeds `_sidecar_status()` `age = now_ts − last_tick_ts` (L1914). None measures elapsed wall-time **within** the collector run, so the injected epoch is the correct sample. |
| AST parse / no NameError | ✅ | — | `python3 -c "import ast; ast.parse(...)"` → PARSE OK. `import time` still present (L31) — harmless. Live smoke-run with injected `epoch=1718409600`: exit 0, empty stderr, `now_epoch` in output echoes argv[4] exactly. |
| `test-timeline-rendering.sh` (incl. new test 65) | ✅ | — | `60 passed, 0 failed`. Test 65 PASS. |
| `test-sse-stream.sh` (incl. test 8 → timeline suite) | ✅ | — | `9 passed, 0 failed`. Test 8 PASS. |
| `colony-lint.sh` | ✅ | — | `367 passed, 0 failed, 5 skipped`. |
| Source-guard test 65 actually catches reintroduction | ✅ | — | Appending `x = time.time()` to a copy → guard's `grep -Eq 'time\.time\(\)'` matches → would `fail`. `COLLECTOR_PY` at test 65 resolves to the real collector (set L859). Does NOT self-match: the `time\.time\(\)` literal lives only in the test script; the grep targets `$COLLECTOR_PY`, never the test file. |
| CHANGELOG `[Unreleased]` accuracy + VERSION not bumped | ✅ | — | CHANGELOG `Fixed` entry describes 5 sites, `NOW_TS = epoch`, source-guard test 65, links #1043 — accurate. The cost day-bucket (`today_start_ms`, L1003) already derived from `epoch_s` pre-fix, so the narrative ("bucket was right, the 5 re-samples were the bug") is exact. `VERSION` = 0.10.3, unchanged (correct — release PRs bump). |
| Out-of-scope claim: `auto-promote-decisions.py` `time.time()` | ✅ | — | 5 reads, none date-boundary-sensitive: L339 `now` → L707 `runtime_hours = (now − started_at)/3600` (elapsed duration); L659 / L669 freshness windows `(time.time() − last_check_epoch) < N·tick_s` (elapsed deltas); L1178 `run_end_ts` + L1250 `generated_at` are raw timestamp stamps, not now-vs-stored comparisons. No calendar-day bucketing like the collector's `today_start_ms`. Correctly out of scope. |

### Verdict: ✅ ship-it

Clean, minimal, surgical fix. All 5 wall-clock re-samples replaced with the single injected `NOW_TS = epoch`; each is provably a now-vs-stored comparison (no in-run duration measurement loses precision). Source-guard test 65 is correct, catches reintroduction, and does not self-match. Timeline (60/0), sse-stream (9/0), and colony-lint (367/0/5) all green. CHANGELOG accurate; VERSION correctly left unbumped. Out-of-scope deferral of `auto-promote-decisions.py` is sound.

PR #1105, opened 2026-06-17
feat(dark-factory): monitor governance + liquidity/flow/pause watchers (#1095, #1096)

ylohnitram:
Four new read-only, **non-custodial** tier-gated watchers feeding the monitor coordinator blackboard (same pattern as invariant/oracle watchers; `cast call` only):

- **governance-watcher.ag** (#1095) — EIP-1967 implementation slot / admin / owner / role-grant change vs baseline (the highest-value pre-exploit early warning) → `monitor:signal:governance`.
- **liquidity-watcher.ag** (#1096) — reserve/TVL drop beyond a learned band → `monitor:signal:liquidity`.
- **flow-watcher.ag** (#1096) — abnormal net out/in over a window vs baseline → `monitor:signal:flow`.
- **pause-state-watcher.ag** (#1096) — `paused()` / circuit-breaker state transition → `monitor:signal:pause`.

Coordinator fuses the new signal kinds into severity + dossier; `config/colony.example.toml` + `start-colony.sh` register all four + their `MONITOR_*` env; README + CHANGELOG updated. Each: one `tier()`/tick, `cb`==budget, `last_check` start+end, baseline via durable memo, degrade-safe, all `exec sh` dynamics `shell_escape()`d.

## Validation
`./tools/colony-lint.sh` → **409 passed, 0 failed, 5 skipped**.

Closes #1095. Closes #1096.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1108, opened 2026-06-17
feat(dark-factory): monitor RPC failover + read consensus + watch-spec drift detector (#1097, #1098)

### LINKED ISSUE #1097: monitor: watch-spec drift detector + periodic re-derivation
Part of EPIC Replikanti/agentis-core#872 (Path C). Builds on the #1086 live-watch runtime.

## Problem
`run-live-watch.sh` derives the invariant set (`watch-spec.json`) **once**. When the target upgrades (new implementation, changed params, new method selectors) the static spec silently stops matching the deployed contract — the monitor keeps "watching" stale invariants and goes effectively blind without saying so.

## What to build
- **Drift detector**: each spec carries a fingerprint of the deployed bytecode / impl slot at derivation time; the watcher (or a periodic job) re-reads it and flags `spec-drift` when it changes. Tie into the governance/upgrade watcher (C4 / #—) since an impl-slot change is exactly the drift trigger.
- **Re-derivation hook**: on drift, raise a `monitor:alert` (kind=`spec-stale`) **and** optionally kick `run-live-watch.sh` to re-derive + hot-swap `MONITOR_INV_SPEC` (operator-gated; re-derivation needs the LLM/forge path).

## Acceptance
- Upgrading the target on a fork raises `spec-stale`; re-derivation produces an updated watch-spec the watcher picks up.
- No drift → no behaviour change. Read-only / non-custodial. `colony-lint.sh` clean.

### LINKED ISSUE #1098: monitor: RPC failover — multiple endpoints + read consensus
Part of EPIC Replikanti/agentis-core#872 (Path C). Builds on #1085.

## Problem
Watchers read from a single `MONITOR_RPC_URL`. A 24/7 service can't rely on one endpoint — and a flaky RPC must not be confused with an invariant verdict.

## What to build
- Accept a **list** of RPC endpoints (e.g. `MONITOR_RPC_URLS`, comma-separated); on read failure fall over to the next.
- Optional **read consensus** for the hot-path fact: agree across ≥2 endpoints before flagging `violated` (cuts false pages from a single lying/lagging node).
- Cleanly distinguish `no-read` (all endpoints failed → feeds the C2 blind meta-alert) from a real verdict.

## Acceptance
- Killing the primary endpoint transparently fails over; killing all → `no-read` → blind path (#1093).
- Consensus mode suppresses a single-node false `violated`. Read-only. `colony-lint.sh` clean.

ylohnitram:
Read robustness for the 24/7 monitor service. **Read-only / non-custodial** throughout.

- **#1098 `monitor/scripts/cast-read.sh`** — the ONE place chain reads happen: read-only wrapper with a hard read-verb whitelist (`call|storage|balance|code`; writes rejected even on caller mis-wiring), **RPC failover** over comma-list `MONITOR_RPC_URLS` (fallback `MONITOR_RPC_URL`), and optional **read consensus** (≥N endpoints must agree before flagging `violated`; disagreement/all-fail → no-read sentinel, distinct from a verdict). All 6 watchers route through it (`MONITOR_CAST_READ`); single endpoint behaves as before.
- **#1097 `monitor/scripts/check-drift.sh` + `run-live-watch.sh` fingerprint** — spec records deployed-code hash + EIP-1967 impl slot at derivation; drift check re-reads and raises `monitor:alert` (kind `drift`, high) when deployed code/impl no longer matches — monitor SAYS it's gone blind vs silently watching a stale spec. Quiet on no-drift/blind/empty. Optional `--rederive`.

Watchers keep big-number-safe handling (#1105) + `shell_escape()` + tier-gating; new env documented + exported; README + CHANGELOG updated.

## Validation
`./tools/colony-lint.sh` → **409 passed, 0 failed, 5 skipped** (stable across re-runs + post-merge). Scripts shellcheck clean. Non-custodial verified.

Closes #1097. Closes #1098.


ylohnitram:
## QA — ✅ ship-it (one finding fixed inline)

Adversarial QA (30+ cast-stub cases): #1098 failover / read-consensus (single lying/lagging node can never push a value past quorum) / write-rejection allowlist / all-fail→no-read sentinel, and #1097 drift detect/quiet/degrade-safe — all pass. Non-custodial verified (zero key/sign/write-RPC tokens). #1105 big-number fix + `shell_escape()` intact. colony-lint 409/0/5.

- **⚠️→✅ drift code-hash casing asymmetry (fixed, 4622249):** `run-live-watch.sh` hashed non-lowercased `cast code` for the baseline while `check-drift.sh` re-reads via `cast-read.sh` (which lowercases) → `sha256(mixed)≠sha256(lower)` → a false `spec-stale` HIGH page on an unchanged target if cast ever emits mixed-case bytecode (dormant on stock foundry). Lowercased the `_code` line symmetric with `_impl`.
- **⚠️ dismissed — consensus value-count tie:** a 2-honest-vs-2-faulty split (quorum 2) returns the `sort -rn`-first value rather than no-read. Accepted: that split is already at/over the Byzantine fault bound (consensus assumes <half faulty); optional hardening, not a defect.

**Verdict: ✅ ship-it.**

Issue #1117, opened 2026-06-18
dev-apprenticeship: first unthrottled end-to-end run on the task it previously died on (data, not vibes)


ylohnitram:
## Problem
Everything "known" about this federation (learns forever, never finishes, under-utilizes Agentis) was observed through a throttle wall. None of it is trustworthy. There has never been a single unthrottled end-to-end run.

## Goal
Run the federation on the exact task it previously died on, with #1111–#1115 in place, and record what actually happens.

## Scope
- Same task, same federation, but now: shared state (#1111), compressed payloads (#1112), caching (#1113), instrumented (#1114), guarded (#1115).
- Observe ONE question: does it reach the completion criterion from #1116, or does it get stuck on something OTHER than I/O?
- Do not optimize Agentis-utilization. Do not refactor for elegance mid-run. Just let it run and watch.

## Done (hard pass/fail)
- A complete run log exists (consumes #1114).
- Outcome is recorded as exactly one of: (a) hit the #1116 criterion → first completion in project history, or (b) stuck on a specific, named, non-I/O failure mode.
- The run was NOT cut early; it ran until completion or a recorded hard stop.

## Blocked by
#1111, #1112, #1114, #1116 (#1113 and #1115 strongly recommended).

ylohnitram:
## Pre-run scaffolding is now in place — this run is operator-driven

Everything needed to make this run objective (not vibes) is merged on `main`:

- **Instrumentation (#1114, PR #1128)** — `tools/cost-rate-report.sh <fed>` emits per-tick / per-agent prompts, prompts/hour, chars in/out, cost, and a **throttle-vs-task-error** split; `--baseline` records the pre-fix number so improvement is provable. A `start-federation.sh` sidecar logs it live to `.agentis/logs/cost-rate.log`.
- **Guardrails (#1115, PR #1128)** — `--cb-per-tick` cap on every daemon + jittered forge-429 backoff with an observable `<colony>:rate-limited` state, so throttling backs off instead of masquerading as task failure.
- **Completion criterion + gate (#1116, PR #1130)** — the single binary "done" condition + `tools/completion-gate.sh` + the human-intervention boundary, in [`doc/dev-apprenticeship-first-task.md`](../blob/main/doc/dev-apprenticeship-first-task.md).
[...]

PR #1127, opened 2026-06-18
feat(dev-apprenticeship): shared GitLab snapshot + payload compression (#1111, #1112)

### LINKED ISSUE #1111: dev-apprenticeship: share one GitLab state snapshot per colony per tick (kill per-agent raw fetch)
## Problem
Federation died from I/O economics, not intelligence. Per #119, every agent `curl`s the full GitLab endpoint every tick and passes raw JSON into `prompt()`. With 21 agents that was ~74 KB/agent and ~2 GB/day through the Claude CLI, and the same `/issues` payload was fetched 3× per tick (labeler, router, prioritizer) and `/merge_requests` 4×. Result: constant throttling, which agents see as transient errors → retries → more throttling. #119 was closed when the federation was cut, NOT when the problem was fixed. This issue rebuilds the fix from scratch.

## Goal
One GitLab state snapshot per colony per tick, fetched once and shared across all agents. No agent fetches raw GitLab data directly anymore.

## Scope
- A single per-colony fetcher pulls each needed endpoint (`/work_items` — GitLab's current name for what used to be `/issues` — `/merge_requests`, …) once per tick.
- The snapshot is exposed to agents through colony-local shared state, not re-fetched per agent.
- Agents read from the snapshot; direct `curl`-into-`prompt()` paths are removed.

## Done (hard pass/fail)
- Number of GitLab HTTP calls per tick == number of distinct endpoints (not number of agents × endpoints).
- Zero agents call GitLab directly; grep of `.ag` sources shows no raw-fetch-into-prompt pattern.
- Measured payload duplication within a tick == 0 for identical endpoints.

## Notes
This is the prerequisite for everything else. Until this lands, throttling masks every other signal, so no other issue can be evaluated.

**Endpoint rename:** GitLab recently renamed the `/issues` REST endpoint to `/work_items`. The `#119` history and the historical fetch counts in the Problem section above refer to the old `/issues` name; any new fetcher must target `/work_items`. The same migration affects the `scripts/gitlab-api.sh` read helpers (`gl_get`/`gl_get_q`) — audit those for hardcoded `/issues` paths while implementing this.

### LINKED ISSUE #1112: dev-apprenticeship: compress GitLab payloads via AST/DAG structural chunking before they reach prompt()
## Problem
Even one shared fetch per tick still pushes large raw JSON into the model. #119 is explicit: caching only helps *after* dev-apprenticeship stops wasting a factor 5–10× on raw-JSON-as-prompt-input. The shared snapshot must be compressed into a compact, structured representation before it reaches `prompt()`.

## Goal
Transform raw GitLab JSON into a structurally-chunked, deduplicated representation using the AST/DAG compression approach (normalized subtree hashing / merkleized structural chunks) before any agent consumes it.

## Scope
- Normalize the snapshot into a structural DAG; hash subtrees so unchanged structure across ticks is referenced, not re-serialized.
- Agents receive the compressed view + only the delta relevant to their role, not the full blob.
- Keep the transform deterministic so the same snapshot always yields the same compressed form (caching depends on this).

## Done (hard pass/fail)
- Median chars-into-`prompt()` per agent drops by ≥5× vs. the raw-JSON baseline on the same real project run.
- Compressed form is byte-stable for identical input (verified by hashing two runs).
- No role loses information it actually needs (each role's required fields are present in its view).

## Notes
This is the first sharp test of the AST/DAG context-compression work in live operation. If it works here it transfers directly to the_factory.

---
_Pairs with #1111 (shared per-colony snapshot) — compression consumes the snapshot that issue produces._

ylohnitram:
21 agents each `curl`'d the full GitLab endpoint per tick + pushed raw JSON into `prompt()` (`/issues` fetched 3×/tick) → throttling death spiral. This builds the **shared-snapshot + compression mechanism** and applies it to the **triage/issues** path (the canonical 3×→1× duplicate-fetch case the issues call the prerequisite for everything else).

**#1111 shared snapshot:** `start-colony` `publish_snapshot()` fetches issues ONCE → memo `gitlab:snapshot:issues` (+`:ts`); the 4 triage agents read via `recall_latest()` + `forge-api.sh issues --from-snapshot` instead of each curl'ing. `gitlab/github-api.sh` gain `snapshot`/`--from-snapshot` verbs (backend-symmetric). `--snapshot-refresh` sidecar mode. **Backward-safe:** missing/empty/stale(>600s)/malformed snapshot → agents fall back to legacy direct fetch (no crash).

**#1112 compression:** `snapshot-compress.py` normalizes items to role-relevant fields, content-addresses + interns repeated structures (normalized-subtree hashing, reusing the dark-factory `struct-sig` concept), refs by index → **~11× smaller**, byte-stable, `--self-test` + `--rehydrate`, total-on-failure.

## Honest scope
Mechanism + triage/issues path done. The `merge_requests` collections (planning/impl/code-review/release — label-event-FILTERED reads, not full-collection) are deferred to the live-run-driven #1117; the mechanism is portable + documented. Measured I/O reduction is only provable on #1117 (not runnable here) — DoD here = mechanism-correct + lint-clean + backward-safe.

## Validation
`./tools/colony-lint.sh` (agentis on PATH) → **416 passed, 0 failed, 5 skipped**; `snapshot-compress.py --self-test` passes (byte-stable round-trip); check-prompt-gate/check-exec-sh/check-forge-dispatch clean.

Closes #1111. Closes #1112.

### DIFF
--- a/dev-apprenticeship/CHANGELOG.md
+++ b/dev-apprenticeship/CHANGELOG.md
@@ -17,6 +17,32 @@ is asserted until multi-version CI is in place.
 
 ### Added
 
+- Shared GitLab snapshot + payload compression for the Triage colony
+  (#1111 / #1112). The `issues` collection is now fetched **once per
+  colony per tick** and shared via a memo instead of each of the four
+  triage agents (labeler / router / prioritizer / issue_creator)
+  curling it independently (the former 3–4× duplicate fetch per tick).
+  `triage/scripts/start-colony.sh` publishes the snapshot to
+  `gitlab:snapshot:issues` (+ epoch freshness key
+  `gitlab:snapshot:issues:ts`) via the new `forge-api.sh snapshot issues`
+  verb on bootstrap, and standalone via `--snapshot-refresh`. The
+  snapshot is **compressed before it reaches `prompt()`**: the new
+  `triage/scripts/snapshot-compress.py` (normalized-subtree-hashing,
+  reusing the `dark-factory/evm-harness/struct-sig.js` concept)
+  normalizes each item to role-relevant fields, content-addresses each
+  item's structure, interns repeated structures once, and references
+  them by index — deterministic + byte-stable, ~11× smaller than raw
+  on a realistic 20-issue payload. Agents read the memo via a new
+  `snapshot_issues_cmd()` helper and render their role view with
+  `forge-api.sh issues --from-snapshot --view <role>` (zero HTTP).
+  Backward-safe: a missing / empty / stale (> 600 s) / malformed
+  snapshot transparently degrades to the legacy direct fetch, and the
+  shared snapshot is used only on the single-repo path (multi-repo
+  fan-out keeps its per-repo fetch). Both `gitlab-api.sh` and
+  `github-api.sh` gained the symmetric `snapshot` verb +
+  `--from-snapshot` flag. The `merge_requests` collection in the other
+  four colonies uses label-event-filtered reads and is deferred to the
+  live run (#1117).
 - Cross-repo reference detection in PR review prompts (#317): the four
   code-review agents (logic / style / security / test) scan PR body
   for `<owner>/<repo>#<N>` references and splice resolved issue context
--- a/dev-apprenticeship/CHANGELOG.md
+++ b/dev-apprenticeship/CHANGELOG.md
@@ -42,7 +42,34 @@ is asserted until multi-version CI is in place.
   `github-api.sh` gained the symmetric `snapshot` verb +
   `--from-snapshot` flag. The `merge_requests` collection in the other
   four colonies uses label-event-filtered reads and is deferred to the
-  live run (#1117).
+  live run (#1117). Three follow-up fixes harden the snapshot path:
+  - **Snapshot-refresh sidecar (#1111).** `start-federation.sh` now
+    runs a background loop that re-publishes the shared snapshot every
+    300 s (override via `SNAPSHOT_REFRESH_INTERVAL_S`) — shorter than
+    the 600 s freshness window — so the snapshot never goes stale and
+    the agents never permanently fall back to per-agent direct fetches
+    (the exact I/O problem #1111 fixes). It mirrors the auto-promote /
+    cost-cap sidecars (self-terminates on zero running daemons,
+    EXIT/TERM/INT trap) and is backward-safe (refresh failures are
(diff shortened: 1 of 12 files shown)


ylohnitram:
## QA report — PR-B shared snapshot + compression (#1111/#1112)

colony-lint (agentis on PATH) **416 passed, 0 failed, 5 skipped**. Compression byte-faithful for labeler/router/prioritizer projected fields (iid/title/labels/state/assignees/author) — verified through the real script across edge cases (empty/1-item/unicode/quotes/backslash/emoji/interning); degrade total-on-failure (missing/empty/stale/malformed → legacy direct fetch, no crash); empty-overwrite guard holds; `GITLAB_SNAPSHOT` shell_escape()'d (injection test with malicious title → no artifacts); single fetch path (check-forge-dispatch clean); multi-repo fan-out unregressed.

### Findings — all resolved inline (commit `5337c69`)

| #8 `--snapshot-refresh` unwired | ✅ resolved | Refresh sidecar wired into `start-federation.sh` (300s, `SNAPSHOT_REFRESH_INTERVAL_S` override, tick-first, self-terminates on zero daemons, killed via EXIT/TERM/INT traps) | Snapshot now stays inside the 600s freshness window for the federation lifetime; `test-auto-promote-install.sh` (13/0) + `test-cost-cap-install.sh` (7/0) still green |
| #2b issue_creator `description` truncated to 200 chars | ✅ resolved | `normalize_issue()` stores full `description`; `rehydrate()` passes it through; only the issue_creator view selects it (no bloat elsewhere) | issue_creator view now **byte-identical** to legacy direct fetch (377-char desc verified); `--self-test` asserts full-description fidelity |
| #9 router/prioritizer prompt() every tick on quiet projects | ✅ resolved | SHA-256 snapshot-content-hash gate added to router/prioritizer/labeler: skip `prompt()` when the view is unchanged since `<agent>:snapshot_hash`; `last_check` still advanced | Empty hash never matches → a hash-compute failure never starves an agent; one `tier()`/tick + existing prompt-gate preserved; check-prompt-gate clean for all 4 agents |

**Verdict:** ✅ ship-it — all 3 findings fixed inline, colony-lint 0-failed, self-test green, `.ag` conventions intact, backward-safe throughout.

Informational (not PR-B): GitLab single-block reachability is pre-existing (#316/#980); MR collections deferred to #1117 per scope.

Issue #1132, opened 2026-06-18
epic: switch all colony federations to flat-cyborg as default LLM backend


ylohnitram:
## Goal

Switch **all colony federations** to the **flat-cyborg** LLM backend (agentis-core v1.19.0 `llm.backend = flat-cyborg`) as the default CLI, so every colony's `prompt()` bills against the flat-rate Claude subscription via the flat-cyborg PTY wrapper instead of the metered `claude -p` (`cli` / `CliBackend`) path or any per-token API.

flat-cyborg (`Replikanti/flat-cyborg`, MIT/Apache) drives the interactive `claude` TUI: per-call `flat-cyborg --tui --auto-approve --extract --no-jitter --idle-ms <N> --cmd <prompt> --timeout-ms <T> -- claude [--model M]`. Config keys: `llm.flat_cyborg.{command,target,idle_ms,no_jitter}` (defaults `flat-cyborg` / `claude` / `4000` / `true`); shares `llm.model`, `llm.cli_timeout_ms`, `llm.max_retries`. Cost is flat-rate (`usage = None`).

**Fidelity caveat (from agentis-core v1.19.0 notes):** `--extract` is a TUI screen-scrape, not a structured envelope like `claude -p --output-format json`. Happy path is reliable; keep the metered `claude` (`cli`) backend available as an opt-in for fidelity-critical work.

## Scope (37 colony configs, all currently `backend = "cli"`)

| Federation | Colonies | Run model | Container? |
|---|---|---|---|
| trading-binance | 6 tribes | replay orchestrator | `tools/Containerfile.replay` |
| tribes-bench | 5 tribes | stage2/3 runners | `tools/Containerfile.stage3` (already has claude-backend + `~/.claude` mount precedent) |
| research-foundry | 18 colonies | `tools/run-research.sh` | `tools/Containerfile.research` |
| dev-apprenticeship | 5 colonies | host daemons (`start-federation.sh`) | none (host has flat-cyborg + claude + creds) |
| dark-factory | 3 colonies | host `run-*.sh` | none |

Plus: scaffolding (`tools/new-colony.sh`, `tools/new-federation.sh` generate `backend = "cli"` → should generate `flat-cyborg`) and `CLAUDE.md` docs.

## Per-federation sub-issues (each its own PR + QA via standard SDLC)

[...]

Issue #1131, opened 2026-06-18
dev-apprenticeship: default CLI backend = flat-cyborg (PTY wrapper over Claude Code)


ylohnitram:
## Problem
The dev-apprenticeship colonies, when run on the CLI LLM backend, drive `claude` directly (`llm.backend = cli` / `llm.command = claude`). We are standardising on **flat-cyborg** — a PTY wrapper over Claude Code — as the default CLI for all colonies. flat-cyborg drives the *interactive* Claude Code session (subscription, not the metered `claude -p` API path) and returns only the model's reply (`--extract`), which is the same integration dark-factory already uses (`dark-factory/flat-cyborg-claude.sh`).

## Goal
Make the flat-cyborg wrapper the **default CLI command** for the dev-apprenticeship federation (all 5 colonies), so choosing the CLI backend drives Claude Code through flat-cyborg instead of bare `claude`.

## Scope
- New shared `tools/flat-cyborg-claude.sh` — the agentis `llm.command` wrapper (prompt from `$1`, stdin fallback), generalised from the proven `dark-factory/flat-cyborg-claude.sh`: `exec flat-cyborg --extract --extract-structural --no-jitter --auto-approve --wrap-input 72 --idle-ms <env> --timeout-ms <env> --cmd "$PROMPT" -- claude`. Resolves `flat-cyborg` + `claude` from PATH (no absolute paths). `FLAT_CYBORG_IDLE_MS` / `FLAT_CYBORG_TIMEOUT_MS` overridable.
- `install.sh` §6 (LLM backend): make the flat-cyborg wrapper the **headline / default** CLI recommendation, and offer to write `llm.backend = cli` + `llm.command = <resolved tools/flat-cyborg-claude.sh>` into `<fed>/.agentis/config` (default Yes; decline → stays mock with the docs). Check `flat-cyborg` is on PATH and warn + point at its install if absent. (The daemon reads `.agentis/config`; per #351 the colony `[llm]` block is inert, so this is the real switch point.)
- All 5 `*/config/colony.example.toml` `[llm]` blocks: update the commented CLI example to the flat-cyborg wrapper (doc-only / forward-compat per #351, kept consistent with the real default).
- `BUNDLE.manifest`: add `tools/flat-cyborg-claude.sh`. README + CHANGELOG.

## Out of scope
[...]

Issue #351, opened 2026-04-26
start-colony.sh splices unsupported --config-override flag (breaks all agent respawns post-#348)


ylohnitram:
## Bug

`agentis daemon` does NOT accept the `--config-override` flag (`agentis daemon --help` only lists `--tick-interval / --cb-per-tick / --deadline / --priority / --colony / --enable-* / --skip-prompt-* / --max-skip-no-input-streak`). PR #348 (#319 PR 1) shipped per-colony `[llm]` config splicing in `start-colony.sh` that emits `--config-override llm.backend=<value>` based on either the cost-cap override file or a colony's `[llm]` block. On any agent restart post-#348, `agentis daemon` errors out with `unknown flag: --config-override` and the spawn dies within 0.5s — `start-colony.sh` reports `agentis daemon failed to launch <name>` and exits 4.

This breaks both:

- **Per-colony LLM override** (#319 PR 1, just landed) — never worked end-to-end on the actual binary.
- **Cost-cap downgrade** (#318) — `tools/cost-cap.sh`'s breach path writes `<fed>/.agentis/llm-backend-override`, then `start-colony.sh --restart-agent` reads it and splices the same nonexistent flag. Cost-cap downgrade was documented as functional but, given the flag is missing, the downgrade likely never re-spawned agents on `mock`.

## Reproduction

Live `dev-apprenticeship` federation, post-#348 + `agentis 1.4.7`:

```
$ bash dev-apprenticeship/planning/scripts/start-colony.sh --restart-agent risk_assessor
start-colony.sh: agentis daemon failed to launch risk_assessor
exit 4

$ bash -x dev-apprenticeship/planning/scripts/start-colony.sh --restart-agent risk_assessor 2>&1 | tail -3
+ agentis daemon /…/risk_assessor.ag --colony planning --enable-exec --enable-messaging --tick-interval 60000 --config-override llm.backend=cli
+ agent_pid=…
+ kill -0 … → fails (process gone)

$ agentis daemon /tmp/nonexistent.ag --config-override llm.backend=cli
agentis daemon: unknown flag: --config-override
```

## Impact

**HIGH**. Any operator-triggered respawn (dashboard `/restart`, `kill-federation && start-federation`, watchdog max-restart respawn, manual `start-colony.sh --restart-agent`) on a federation that has either:

- A colony.toml `[llm]` block (created by an operator copying from `colony.example.toml`), OR
[...]

PR #1144, opened 2026-06-18
fix(dev-apprenticeship): colony.example.toml [llm] backend back to "cli" (inherit, per CLAUDE.md)

ylohnitram:
Follow-up correction to #1131.

The CLAUDE.md **"LLM backend"** convention (added in #1132) is explicit: a colony's `[llm] backend` in `colony.example.toml` is `"cli"` — meaning *"use the agentis daemon default"*, which **inherits** the federation-level backend from `.agentis/config`; a specific backend must **not** be hardcoded there (it would override the federation default and break the host wrapper path).

#1131 flipped all 5 `colony.example.toml` blocks to `backend = "claude"` to silence a deprecation warning — but that warning comes from `.agentis/config` (where `install.sh` correctly writes `claude`), **not** from the colony block, which is **inert per #351** (the daemon reads `.agentis/config`). So the change was unnecessary and contradicted the documented convention.

## Change
- Revert `backend` → `"cli"` in all 5 `dev-apprenticeship/*/config/colony.example.toml`.
- Add a comment documenting the `cli` = inherit / don't-hardcode / inert-per-#351 semantics, pointing at the CLAUDE.md "LLM backend" section.
- Doc-consistency only — the real backend (`llm.backend = claude` + the flat-cyborg wrapper) is still written to `.agentis/config` by `install.sh`; runtime behaviour unchanged.

## Verification
- colony-lint: **417 passed, 0 failed, 5 skipped**
- All 5 `[llm]` blocks byte-identical; no `backend = "claude"` left in any colony.example.toml.

### DIFF
--- a/dev-apprenticeship/CHANGELOG.md
+++ b/dev-apprenticeship/CHANGELOG.md
@@ -15,6 +15,18 @@ is asserted until multi-version CI is in place.
 
 ## [Unreleased]
 
+### Fixed
+
+- Reverted the `[llm] backend` in all 5 `*/config/colony.example.toml`
+  blocks from `"claude"` back to `"cli"` (a follow-up correction to #1131).
+  Per the CLAUDE.md "LLM backend" convention, `"cli"` means "use the
+  agentis daemon default" and **inherits** the federation-level backend
+  from `.agentis/config` (flat-cyborg); a specific backend must **not** be
+  hardcoded in the colony block (it would override the federation default
+  and break the host wrapper path). The block is inert per #351 either way,
+  so this is doc-consistency only — the real backend (`llm.backend = claude`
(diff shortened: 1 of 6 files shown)


ylohnitram:
CI run [27796699573](https://github.com/Replikanti/agentis-colonies/actions/runs/27796699573) failed on **unrelated federation-dashboard tests** (`test-sse-stream.sh`, `test-timeline-rendering.sh`), not this diff (which only touches `colony.example.toml` `[llm]` blocks). Root cause = a **midnight-UTC-boundary flake**: the run straddled 23:58→00:01, tripping the epoch/wall-clock-sensitive timeline assertions. Both suites pass locally (60/0, 9/0) and every other recent run (main + siblings, pre-midnight) is green. Re-ran the failed jobs; tracked the flake as #1145.

Issue #1216, opened 2026-06-20
code-edit-in-checkout: commit the diff even when flat-cyborg exits non-zero (idle/timeout)


ylohnitram:
## Problem

`tools/code-edit-in-checkout.sh` checks the flat-cyborg editing-agent exit code **before** staging the diff:

```sh
FC_RC=$?
if [ "$FC_RC" -ne 0 ]; then exit "$FC_RC"; fi   # <-- discards a completed edit
git add -A
...
```

flat-cyborg drives an **interactive Claude Code TUI** that frequently never settles to its idle prompt, so flat-cyborg rides to its `--timeout-ms` and returns non-zero **even when claude finished editing correctly**. The orchestrator then `exit $FC_RC` and throws away a perfectly good edit. The autonomous `code_writer` retries, hits the same wall, and loops — never opening the PR. Observed live on a doc-update task: claude made the correct surgical multi-file edit (`git diff --stat` confirmed), but every run rode to the 600s timeout and discarded it.

## Fix

The **artifact is the edit, not the editing session's exit code**. Stage first, then decide on the diff:

| State | Action |
|-------|--------|
| staged change present | commit + push + open PR, **regardless of FC_RC** (a timeout that still produced edits is a success) |
| no change + FC_RC == 0 | `NO_EDITS` → exit 3 (retry, not error) — unchanged |
| no change + FC_RC != 0 | session genuinely failed → exit FC_RC |

## Tests

`tools/test-code-edit-in-checkout.sh` gains a `FC_STUB_RC` knob and two runs: run 3 (edit + non-zero exit → commit + PR + exit 0) and run 4 (no-edit + non-zero exit → exit FC_RC, no PR). 23 passed, 0 failed; colony-lint 428/0/5.

PR #1220, opened 2026-06-21
fix(tools): flat-cyborg-claude reads reply from a result file, not TUI scrape (#1219)

### LINKED ISSUE #1219: flat-cyborg-claude backend: read reply from a result file, not the TUI screen-scrape
## Problem

`tools/flat-cyborg-claude.sh` (the `llm.command` backend that drives the interactive Claude Code session) extracts the model's reply by **screen-scraping the TUI** (`--extract` / `--extract-structural`). This is fundamentally lossy for any reply taller than the screen: the start sentinel scrolls out of view, `--extract` returns nothing, and `--extract-structural` falls back to whatever chrome/prose is on screen. The agentis caller then decodes **invalid JSON**.

Observed live on the dev-apprenticeship federation: `code_writer` drafting prompts intermittently failed with `LLM invalid response: CLI returned invalid JSON` (`expected ','` / `unexpected end of input`), and TUI chrome (`❯ ⏵⏵ auto mode on …`) leaked into a downstream task. The first attempt at a task would succeed on a clean scrape; a retry would get a corrupted scrape and fail. This is the root cause of unreliable autonomous runs.

## Fix

Stop trusting the screen. Append an output-channel directive to the prompt telling claude to write its **complete reply** to a temp file with its file-write tool, and read **that file** instead of the scraped screen. claude's Write tool emits exact bytes — no scroll, no line-wrap, no chrome.

- Result file non-empty → authoritative reply (even if flat-cyborg exited non-zero on an idle/timeout that still produced the answer — Write is atomic, so a non-empty file is complete; same spirit as commit-on-diff in #1216).
- Result file empty → fall back to the existing screen-scrape path, so the change is **strictly >= old behaviour**.

Backend-generic (JSON and prose consumers alike); no flat-cyborg binary change required.

## Tests

`tools/test-flat-cyborg-claude.sh` gains a stubbed-`flat-cyborg` section (CI-runnable, no real flat-cyborg/claude): result-file preferred (4), screen fallback when no file (5), non-empty file honoured on non-zero exit (6), exit-code propagates when nothing produced (7). 7 passed; colony-lint 428/0/5. Validated live: the real wrapper returns valid JSON via the file channel where the screen-extract returned prose.

ylohnitram:
Fixes #1219.

## Root cause
`tools/flat-cyborg-claude.sh` extracted claude's reply by **screen-scraping the TUI** (`--extract`/`--extract-structural`). For any reply taller than the screen the start sentinel scrolls off → `--extract` returns nothing, `--extract-structural` returns on-screen chrome/prose → the agentis caller decodes **invalid JSON**. This is the root cause of the dev-apprenticeship federation's intermittent `CLI returned invalid JSON` drafting failures and the `❯ ⏵⏵ auto mode on …` chrome leaking into a task — clean scrape succeeded, corrupted scrape failed, so runs were nondeterministically unreliable.

## Fix
Stop trusting the screen: append an output-channel directive telling claude to write its **complete reply** to a temp file with its file-write tool, and read **that file**. claude's Write emits exact bytes — no scroll, no line-wrap, no chrome.

| State | Behaviour |
|-------|-----------|
| result file non-empty | authoritative reply, **even on a flat-cyborg non-zero exit** (idle/timeout that still produced the answer; Write is atomic → non-empty = complete). Same spirit as commit-on-diff in #1216. |
| result file empty | fall back to the existing screen-scrape path → **strictly ≥ old behaviour** |

Backend-generic (JSON + prose); no flat-cyborg binary change.

## Tests
`tools/test-flat-cyborg-claude.sh` gains a stubbed-`flat-cyborg` section (CI-runnable, no real flat-cyborg/claude): result-file preferred (4), screen fallback when no file (5), non-empty file honoured on non-zero exit (6), exit-code propagates when nothing produced (7). `7 passed, 0 failed`; `colony-lint 428 passed, 0 failed, 5 skipped`. Validated live: the real wrapper returns valid JSON via the file channel on a prompt where the screen-extract returned prose.

### DIFF
--- a/tools/flat-cyborg-claude.sh
+++ b/tools/flat-cyborg-claude.sh
@@ -5,7 +5,8 @@
 # metered `claude -p` API path). agentis invokes the configured llm.command with
 # the prompt as a trailing positional arg; this wrapper takes the prompt from
 # "$1" (falling back to stdin) and returns only the model's reply on stdout
-# (flat-cyborg --extract).
+# (read from a result file claude writes; see the RESULT-FILE channel note
+# below, with a flat-cyborg --extract screen-scrape as fallback).
 #
 # This generalizes the proven dark-factory/flat-cyborg-claude.sh into shared,
 # path-independent tooling so every dev-apprenticeship colony can use the same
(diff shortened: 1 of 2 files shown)


ylohnitram:
## QA Report — `fix/fc-file-result` (RESULT-FILE channel)

Two-file diff (`tools/flat-cyborg-claude.sh`, `tools/test-flat-cyborg-claude.sh`).

| RESULT_FILE via mktemp + cleanup trap | ✅ | — | Created up front; trap removes all three temp files. |
| Directive appended with correct path | ✅ | — | mktemp path is alnum-only `/tmp/tmp.XXXX`; last token on the line; no wrap/quote-breaking chars. |
| Selection logic; no double-output; exit code intact | ✅ | — | File branch (`[ -s ] && grep -q '[^[:space:]]'`) → unwrap → `exit 0`; empty → `FC_RC` propagates → screen unwrap. Single output per path. |
| Fallback byte-identical to old behaviour when file empty | ✅ | — | Final two lines identical to origin/main; verified empirically (screen-only + none scenarios match a reverted old wrapper). |
| Whitespace-only file falls back (grep beats `[ -s ]`) | ✅ | — | Confirmed empirically the second predicate does real work. |
| `set -eu` not broken by new predicates | ✅ | — | Both sit in `if`/`&&` conditions (explicit `set -e` exemption). |
| No token/secret regression | ✅ | — | Wrapper handles no tokens; nothing from the reply is executed (only piped through unwrap.py). |
| POSIX/dash safety | ⚠️ | Dismissed — `shellcheck -s sh` clean (POSIX mode, flags bashisms) + manual audit (no `[[`, arrays, `local`, `function`, procsub) + CI dash gate green. Local `dash -n` unavailable (dash absent, `/bin/sh`→bash). | New constructs (`[ -s ]`, `grep -q`, multi-line `PROMPT=`, `printf '%s'`) all POSIX. |
| Tests 4–7 discriminate new vs old | ✅ | — | Against a reverted old wrapper, test-4/test-6 scenarios yield `rc=1 out=[]` → would fail the new assertions; 5/7 lock the unchanged fallback. |
| Test suite | ✅ | — | `7 passed, 0 failed`. |
| colony-lint | ✅ | — | `428 passed, 0 failed, 5 skipped`; new test auto-discovered. |
| Directive on EVERY prompt (all backends) | ⚠️ | Dismissed — file channel returns exact bytes; prose passes through unwrap.py byte-for-byte; if a model ignores the directive it falls back to today's path. Strictly safer than the lossy scrape. Watch-item. | — |
| "non-empty = complete" vs torn write | ⚠️ | Watch (no code change) — claude's Write writes the full content in one call; realistic failure is "no file" (handled by fallback), not "half a file". | Holds for Write-as-used. |

### Verdict
✅ **ship-it** — correct and strictly ≥ old behaviour (empty file → byte-identical legacy path, verified against a reverted copy; new branch only adds a recovery path). No double-output, no exit-code loss, `set -eu` intact, mktemp path injection-safe, tests 4–7 demonstrably discriminate the new branches. Non-green rows are dismissed/watch-only; the POSIX gap is closed by `shellcheck -s sh` clean + the CI dash gate.

Issue #1192, opened 2026-06-20
docs: update README 'How work enters the system' (assignment-based pickup, forge-agnostic) + stale CI baseline


ylohnitram:
## Problem
Several behavior changes this session were not reflected in the operator docs (CHANGELOG was updated, but `dev-apprenticeship/README.md` and `CLAUDE.md` were not). The most harmful is the "How work enters the system" section, which still tells operators that cross-colony routing is automatic and reliable — the opposite of what is now true.

### 1. `dev-apprenticeship/README.md` — "How work enters the system"
- Source 1 currently says "GitLab polling". Make it **forge-agnostic**: "GitHub or GitLab polling" (the GitHub forge is the proven path).
- Source 2 currently says cross-colony handoffs are automatic and "You do not need to trigger these handoffs manually." This is no longer reliable. Replace with: work is picked up **by assignment** — an issue **assigned to the federation's bot account** (and carrying the colony's trigger label, default `implementation` / `needs-planning`) is picked up on the next tick (#1181). Cross-colony bus events are a best-effort fast path, not a guarantee; assigning the issue is the reliable trigger.
- Add a short note that the implementation colony's `code_writer` now **edits existing files** (fetches current content first), not just creates new ones (#1172).

### 2. `dev-apprenticeship/README.md` — operator setup
- Note that `exec.env_passthrough` must include `IMPLEMENTATION_TRIGGER_LABEL,PLANNING_TRIGGER_LABEL` for a `colony.toml` `trigger_label` override to take effect (install.sh sets this as of #1185); otherwise the default trigger label is used.

### 3. `CLAUDE.md` — stale CI baseline
- The line "Current CI baseline: 42 passed, 0 failed, 1 skipped" is stale. The current CI colony-lint baseline is **289 passed, 0 failed, 6 skipped**. Update the number.

## Done (checkable)
- README "How work enters the system" describes **assignment-based** pickup + forge-agnostic polling, and no longer claims cross-colony routing is automatic/reliable.
- README documents code_writer editing existing files + the trigger-label env-passthrough requirement.
- CLAUDE.md CI baseline reads 289 passed / 0 failed / 6 skipped.
- `./tools/colony-lint.sh` 0 failed (markdown links resolve).

ylohnitram:
[...]

PR #1222, opened 2026-06-21
fix(tools): code-edit editing agent uses --extract to reliably drive Claude (#1221)

### LINKED ISSUE #1221: code-edit-in-checkout: editing agent must use --extract to reliably drive Claude (no --extract = no edit)
## Problem

`tools/code-edit-in-checkout.sh` runs the editing Claude session via flat-cyborg **without** `--extract` / `--wrap-input`, on the theory that an editing agent makes file changes with its own tools and never needs a scraped reply. But that flag set does not reliably **drive** the interactive session: the prompt is typed but never submitted/executed, so claude sits idle at its input bar (`❯ ⏵⏵ auto mode on …`) until `--timeout-ms` and produces **no edit**.

Reproduced live and in isolation: the exact orchestrator invocation on a **one-line** task ("append a line to CLAUDE.md") → rc=124, no edit. The same invocation **with** `--extract --extract-structural --wrap-input 72` → rc=0, edit applied ("Done. Appended …"). This is why the federation's #1192 re-run kept failing at the edit stage while drafting succeeded.

## Fix

Mirror `flat-cyborg-claude.sh`'s full driving flag set in the editing invocation: add `--extract --extract-structural --wrap-input 72`. `--extract` is what makes flat-cyborg wait for readiness, submit the prompt, and let claude run to a settled screen. We do **not** consume the scraped reply (the artifact is the `git diff`), and we redirect flat-cyborg's stdout to stderr so the scraped screen text can never contaminate this script's stdout, which must stay the PR URL alone (also fixes TUI chrome leaking into the job result).

## Tests

`tools/test-code-edit-in-checkout.sh` run-1 flag assertion flips from "must NOT be --extract" to "must drive with --extract + --wrap-input". 23 passed; colony-lint 428/0/5. Validated live: editing probe edits with the new flags, sits idle without them.

ylohnitram:
Fixes #1221.

## Root cause
`tools/code-edit-in-checkout.sh` drove the editing Claude session via flat-cyborg **without** `--extract` / `--wrap-input` (the comment claimed an editing agent needs no scraped reply). But that flag set does not reliably **submit-and-run** the interactive session: the prompt is typed but never executed → claude sits idle at `❯ ⏵⏵ auto mode on …` until `--timeout-ms` → **no edit**. This is why the federation's #1192 re-run kept failing at the edit stage while drafting succeeded.

Reproduced in isolation:
| Invocation | Result |
|------------|--------|
| orchestrator's flags (no `--extract`), one-line edit task | rc=124, **no edit** |
| same + `--extract --extract-structural --wrap-input 72` | rc=0, **edit applied** ("Done. Appended …") |

## Fix
Add `--extract --extract-structural --wrap-input 72` to the editing invocation — purely to **drive** the session (wait for readiness, submit, run to a settled screen). The scraped reply is **not** consumed (the artifact is the `git diff`), and flat-cyborg's stdout is redirected to stderr so screen text never contaminates this script's stdout, which must stay the PR URL alone (also stops TUI chrome leaking into the job result — observed earlier).

Pairs with #1216 (commit-on-diff) and #1219 (result-file channel for the drafting prompt) to make the autonomous edit path reliable end to end.

## Tests
`tools/test-code-edit-in-checkout.sh` run-1 flag assertion flips from "must NOT be `--extract`" to "must drive with `--extract` + `--wrap-input`". `23 passed, 0 failed`; `colony-lint 428 passed, 0 failed, 5 skipped`. Validated live: editing probe edits with the new flags, sits idle without them.

### DIFF
--- a/tools/code-edit-in-checkout.sh
+++ b/tools/code-edit-in-checkout.sh
@@ -198,21 +198,31 @@ run_git -C "$WS" checkout -B "$BRANCH" "origin/$DEFAULT_BRANCH"
 
 # ---------------------------------------------------------------------------
 # 3. Edit: run claude (via flat-cyborg) as an editing agent inside the
-#    checkout. We mirror flat-cyborg-claude.sh's flags MINUS --extract /
-#    --extract-structural: here claude EDITS files with its own tools, so we do
-#    not need to scrape a reply off the screen. --cwd points it at the checkout;
-#    --auto-approve lets it write without an interactive confirm; --no-jitter +
-#    the idle/timeout knobs keep behaviour identical to the wrapper.
+#    checkout. We mirror flat-cyborg-claude.sh's FULL flag set, including
+#    --extract --extract-structural --wrap-input 72. Earlier this path dropped
+#    those flags on the theory that an editing agent makes file changes with its
+#    own tools and never needs a scraped reply — but #1221 showed that without
+#    --extract flat-cyborg does NOT reliably drive the interactive Claude session
+#    to submit-and-act: the prompt is typed but never executed, so claude sits
+#    idle at the input bar until --timeout-ms and produces NO edit (reproduced
+#    live, even for a one-line edit). --extract is what makes flat-cyborg wait
+#    for the session to be ready, submit the prompt, and let claude run to a
+#    settled screen. We do NOT consume its scraped reply here (the artifact is
+#    the git diff), and we redirect flat-cyborg's stdout to stderr so the
+#    scraped screen text can never contaminate this script's stdout, which must
+#    stay the PR URL alone. --wrap-input 72 folds the (large) task prompt so it
+#    does not overflow claude's input editor; --cwd points claude at the
+#    checkout; --auto-approve writes without an interactive confirm.
 # ---------------------------------------------------------------------------
 printf 'Implement issue #%s in this repository by editing the files directly with your tools. %s\n\n%s\n\nMake the change and stop; do not run git or open a PR.' \
     "$ISSUE" "$TITLE" "$TASK" > "$TASKFILE"
 
 echo "[code-edit] running flat-cyborg editing agent in $WS" >&2
 set +e
-flat-cyborg --no-jitter --auto-approve --cwd "$WS" \
+flat-cyborg --extract --extract-structural --no-jitter --auto-approve --wrap-input 72 --cwd "$WS" \
     --idle-ms "${FLAT_CYBORG_IDLE_MS:-8000}" \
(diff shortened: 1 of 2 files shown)


ylohnitram:
## QA Report — `fix/fc-edit-driving`

`code-edit-in-checkout.sh` now drives the editing Claude session with the full flat-cyborg flag set (`--extract --extract-structural --wrap-input 72`) and redirects flat-cyborg's stdout to stderr so screen text never contaminates the PR-URL-only stdout.

| New flags added correctly | ✅ | — | `--extract --extract-structural --no-jitter --auto-approve --wrap-input 72 --cwd` — matches the wrapper's driving set. |
| `>&2` on flat-cyborg line; FC_RC captures its exit | ✅ | — | Redirect is on the flat-cyborg command; `cmd >&2; echo $?` reflects cmd's status (verified). |
| stdout = PR URL only (success) | ✅ | — | Reproduced with simulated screen noise → routed to stderr; stdout carried only the PR URL. `run_git` → stderr. |
| `NO_EDITS` still on stdout (retry contract) | ✅ | — | `echo "NO_EDITS"` unredirected; run 2 asserts exit 3 + `^NO_EDITS$`. |
| Commit-on-diff (#1216) unchanged | ✅ | — | Diff touches only the invocation line + comment; stage-then-decide block byte-identical to main. |
| Token safety | ✅ | — | No token line touched; GIT_ASKPASS path intact; `set -x` + stdout/stderr leak tests pass. |
| Test suite | ✅ | — | `23 passed, 0 failed`, deterministic. |
| Flipped run-1 assertion non-vacuous | ✅ | — | Now requires `--extract` AND `--wrap-input` (verified against real flat-cyborg 0.11.0). |
| colony-lint | ✅ | — | `428 passed, 0 failed, 5 skipped`. |
| shellcheck | ✅ | — | Clean under bash + `-s sh`; no new warnings from changed lines. |
| `--extract` may ride to `--timeout-ms` | ✅ | — | An editing session may not emit the sentinel → can ride to `CODE_EDIT_TIMEOUT_MS`; harmless by design — #1216 commit-on-diff treats non-zero-exit + staged diff as success (run 3 proves it). |
| Stale stub-header comment | ✅ | Fixed inline (follow-up commit) — line 127 now reads the full driving set incl. --extract. | Was the only ⚠️; resolved. |

### Verdict
✅ **ship-it** — correct and well-tested. `>&2` does not mask flat-cyborg's exit code; stdout-is-PR-URL-only holds (verified with simulated noise); NO_EDITS retry + #1216 commit-on-diff intact; token discipline unchanged; 23 tests + colony-lint 428/0 + shellcheck all green. The one ⚠️ (stale comment) is fixed inline.

Issue #1227, opened 2026-06-21
dashboard: sidecar health monitors only auto-promote — also surface snapshot-refresh + cost-rate sidecars


ylohnitram:
## Problem

The federation runs **three** background sidecars, each with its own log under `<fed>/.agentis/logs/`:

| Sidecar | Log | Interval | Started by |
|---------|-----|----------|------------|
| auto-promote | `auto-promote.log` | 1800s | `start-federation.sh` (#216), gated by `.auto-promote-install.toml` |
| snapshot-refresh | `snapshot-refresh.log` | 300s | colony start (always-on) |
| cost-rate | `cost-rate.log` | 60s | colony start (always-on) |

But the dashboard collector (`federation-dashboard/lib/federation-dashboard-collector.py`) only models the **auto-promote** sidecar — the `sidecar = {...}` block reads `.auto-promote-install.toml` + `auto-promote.log` + `auto-promote.sidecar_started_at` and nothing else. `grep -c 'snapshot-refresh\|cost-rate'` on the collector returns 0. So snapshot-refresh and cost-rate liveness is **invisible** on the dashboard — historically there was only the auto-promote sidecar and the other two were never wired into the status panel when they shipped.

## Required

Generalize sidecar health so the dashboard surfaces **all three** sidecars in the existing sidecar status area (the `_sidecar_status()` / "Sidecar status rows" path). For each sidecar: detect its `<fed>/.agentis/logs/<name>.log`, derive last-tick from the log mtime, and mark it `running` when the last tick is within a staleness ceiling of `4 × interval` (matching the existing auto-promote `down` cutoff), else `stale`/`down`.

Note the asymmetry — do **not** assume the other two behave like auto-promote:
- Only **auto-promote** has an install toggle (`.auto-promote-install.toml`) and writes a `*.sidecar_started_at` stamp + startup-grace. Keep its existing installed/enabled/startup-grace semantics **unchanged**.
- **snapshot-refresh** and **cost-rate** are always-on (no enable toggle, no `started_at` stamp). For those, liveness = fresh log tick within `4 × interval` (240s for cost-rate, 1200s for snapshot-refresh); there is no "not installed" state for them.

The known per-sidecar intervals (1800 / 300 / 60s) can be a small table in the collector.

## Done (checkable)
- The collector emits liveness for all three sidecars (auto-promote, snapshot-refresh, cost-rate).
[...]

Issue #216, opened 2026-04-19
install.sh: offer auto-promote cron install (complete the system)


ylohnitram:
## Problem

The auto-promote / auto-evolve system (#148) ships as `tools/auto-promote.sh` with a docstring saying "Intended to be invoked by cron (e.g. `*/30 * * * *`)". The README (`dev-apprenticeship/README.md` section "Auto-promote and auto-evolve") shows a cron-entry example as a commented-out line.

But nothing actually installs it. Operators who run `./install.sh` get a fully working federation that never self-promotes — because they're expected to figure out cron by themselves. That's a half-finished system.

## Why not cron

A first-draft of this issue proposed splicing a cron entry into the operator's crontab during `install.sh`. That was discarded because:

1. **Stale entries survive teardown.** `kill-federation.sh` tears down daemons + dashboard + sidecars, but a crontab entry persists system-wide. Next `./install.sh` of a new federation in a different dir, or no federation at all, still has the cron ticking against a possibly-removed path.
2. **Cron lives outside the federation.** Scheduling belongs *to* the federation's lifecycle — start when federation starts, die when federation dies. Cron couples it to the user account instead.
3. **Discoverability.** A line hidden in `crontab -e` is less visible than a process visible to `pgrep` and controllable via the dashboard / kill script.

## Proposal

Schedule auto-promote **inside the federation's own lifetime**, so it dies naturally when the federation is torn down. Two reasonable hosts for the scheduler exist:

**Option A — Background thread in `federation-dashboard-server.py`.** The server runs while the operator has the dashboard open. A `threading.Thread(..., daemon=True)` ticks every 30 min and shells out to `tools/auto-promote.sh <fed-name>`. When the server exits (Ctrl-C, `/kill`, `kill-federation.sh`), the daemon thread dies with it. Pros: already has threading precedent; natural surface for a "last auto-promote run" indicator in the dashboard UI. Cons: only runs while dashboard is open — federations where the operator does not keep a dashboard tab up get no auto-promote.

[...]

PR #1228, opened 2026-06-21
The collector already emits a `sidecars` array consumed by `renderSidecarStatus()`, but it only models auto-promote (and cost-cap), so snapshot-refresh and cost-rate liveness is invisible. The fix adds a per-sidecar interval table plus a log-mtime liveness helper and extends that array with always-on records for snapshot-refresh and cost-rate, reusing the existing `4 × interval` staleness ceiling while leaving auto-promote's install/started_at/startup-grace semantics unchanged. Template and CHANGELOG updates then render all three rows and document the change, verified via `./tools/colony-lint.sh`.

ylohnitram:
Implements #1227.

Issue:
{"iid": 1227, "title": "dashboard: sidecar health monitors only auto-promote \u2014 also surface snapshot-refresh + cost-rate sidecars", "description": "## Problem\n\nThe federation runs **three** background sidecars, each with its own log under `<fed>/.agentis/logs/`:\n\n| Sidecar | Log | Interval | Started by |\n|---------|-----|----------|------------|\n| auto-promote | `auto-promote.log` | 1800s | `start-federation.sh` (#216), gated by `.auto-promote-install.toml` |\n| snapshot-refresh | `snapshot-refresh.log` | 300s | colony start (always-on) |\n| cost-rate | `cost-rate.log` | 60s | colony start (always-on) |\n\nBut the dashboard collector (`federation-dashboard/lib/federation-dashboard-collector.py`) only models the **auto-promote** sidecar \u2014 the `sidecar = {...}` block reads `.auto-promote-install.toml` + `auto-promote.log` + `auto-promote.sidecar_started_at` and nothing else. `grep -c 'snapshot-refresh\\|cost-rate'` on the collector returns 0. So snapshot-refresh and cost-rate liveness is **invisible** on the dashboard \u2014 historically there was only the auto-promote sidecar and the other two were never wired into the status panel when they shipped.\n\n## Required\n\nGeneralize sidecar health so the dashboard surfaces **all three** sidecars in the existing sidecar status area (the `_sidecar_status()` / \"Sidecar status rows\" path). For each sidecar: detect its `<fed>/.agentis/logs/<name>.log`, derive last-tick from the log mtime, and mark it `running` when the last tick is within a staleness ceiling of `4 \u00d7 interval` (matching the existing auto-promote `down` cutoff), else `stale`/`down`.\n\nNote the asymmetry \u2014 do **not** assume the other two behave like auto-promote:\n- Only **auto-promote** has an install toggle (`.auto-promote-install.toml`) and writes a `*.sidecar_started_at` stamp + startup-grace. Keep its existing installed/enabled/startup-grace semantics **unchanged**.\n- **snapshot-refresh** and **cost-rate** are always-on (no enable toggle, no `started_at` stamp). For those, liveness = fresh log tick within `4 \u00d7 interval` (240s for cost-rate, 1200s for snapshot-refresh); there is no \"not installed\" state for them.\n\nThe known per-sidecar intervals (1800 / 300 / 60s) can be a small table in the collector.\n\n## Done (checkable)\n- The collector emits liveness for all three sidecars (auto-promote, snapshot-refresh, cost-rate).\n- `federation-dashboard/lib/federation-dashboard.html.template` renders all three in the sidecar status area.\n- `federation-dashboard/CHANGELOG.md` `[Unreleased]` notes the change.\n- `./tools/colony-lint.sh` reports 0 failed (dashboard tests + markdown links pass).\n", "state": "opened", "labels": ["dev-apprenticeship", "implementation"], "assignees": [{"username": "ylohnitram"}], "author": {"username": "ylohnitram"}, "created_at": "2026-06-21T10:02:53Z", "updated_at": "2026-06-21T10:03:45Z", "user_notes_count": 0, "priority": null}

Plan:
- `federation-dashboard/lib/federation-dashboard-collector.py` — Add a small per-sidecar interval table (`auto-promote` 1800s / `snapshot-refresh` 300s / `cost-rate` 60s) as the single source of truth. Add a helper that derives always-on liveness for a sidecar from the mtime of `<fed>/.agentis/logs/<name>.log` (last-tick = mtime; `last_tick_ts=None` when the log is absent). Extend the existing `sidecars = [...]` array (currently only `auto-promote` + `cost-cap`) with `snapshot-refresh` and `cost-rate` records, each `installed=True/enabled=True` with no `started_at`/startup-grace, so `_sidecar_status()` applies its existing `4 × interval` `down` ceiling (240s cost-rate, 1200s snapshot-refresh) → `running`/`stale`/`down`. Leave the `auto-promote` record, its `.auto-promote-install.toml` install toggle, `*.sidecar_started_at` stamp, and startup-grace semantics byte-identical.
- `federation-dashboard/lib/federation-dashboard.html.template` — Update the default `sidecars` fallback and the `renderSidecarStatus()` comments (currently say "two known sidecars") so all three render; ensure always-on rows (no install file) don't show a spurious `not installed` age or an incorrectly-disabled restart button — treat `installed/enabled=true` for them.
- `federation-dashboard/CHANGELOG.md` — Add an `[Unreleased] ### Added` entry noting the dashboard now surfaces liveness for all three federation sidecars (auto-promote, snapshot-refresh, cost-rate), referencing #1227.

The collector already emits a `sidecars` array consumed by `renderSidecarStatus()`, but it only models auto-promote (and cost-cap), so snapshot-refresh and cost-rate liveness is invisible. The fix adds a per-sidecar interval table plus a log-mtime liveness helper and extends that array with always-on records for snapshot-refresh and cost-rate, reusing the existing `4 × interval` staleness ceiling while leaving auto-promote's install/started_at/startup-grace semantics unchanged. Template and CHANGELOG updates then render all three rows and document the change, verified via `./tools/colony-lint.sh`.


ylohnitram:

This MR adds real production logic (the _always_on_sidecar_record helper, SIDECAR_INTERVALS_S table, and two new sidecar records) but only updates one render-level test (test 40), which asserts presence + installed/enabled flags. The feature's actual value -- per-sidecar liveness status derived from log mtime against per-interval down/silent ceilings -- is entirely untested, as are the missing-log and OSError edge paths and the interval_s cadence contract. Tests are inadequate: add a direct unit test for _always_on_sidecar_record covering fresh/silent/down/absent/error cases, pin interval_s values, and assert the full always-on record shape (started_at_ts=None, in_startup_grace=False, running_orphan=False).

ylohnitram:

The diff is well-organized and follows existing conventions closely: new identifiers (`SIDECAR_INTERVALS_S` constant, `_always_on_sidecar_record` helper) match the surrounding snake_case / UPPER_SNAKE naming and the leading-underscore private convention, dict key ordering mirrors the existing sidecar records, and the test file reuses the established `%r`-style string formatting rather than introducing a new style. The only real style nits are a PEP 8 two-blank-line deviation around the new top-level function and a trivial spacing inconsistency in the `×` glyph between the docstring and the CHANGELOG. No naming, import-ordering, or whitespace problems otherwise.

ylohnitram:
## QA Report — `fix/issue-1227` (dashboard always-on sidecars)

| `fed_dir` closure + `now_ts_for_sidecars` scope | ✅ | — | `fed_dir` module-global (set from argv before the helper def + call sites); `now_ts_for_sidecars=NOW_TS` immediately above the `sidecars=[...]` literal. No NameError at runtime. |
| Records APPENDED (index stability) | ✅ | — | Runtime: `sidecars[0]=auto-promote, [1]=cost-cap, [2]=snapshot-refresh, [3]=cost-rate`. Positional orphan test 5/5. |
| `_sidecar_status()` signature | ✅ | — | Call `(True,True,False,last_tick_ts,interval_s,now_ts)` matches the auto-promote call site order. |
| Template renders always-on records | ✅ | — | `installed=true,enabled=true,started_at=null` + fresh tick → live age + enabled `[restart]`, NOT "not installed"; absent log → "never"/silent. |
| CHANGELOG under `[Unreleased]` | ✅ | — | First item under `[Unreleased] → Added`, refs #1227. |
| `test-timeline-rendering.sh` | ✅ | — | 62 passed, 0 failed (test 40 extended with always-on assertions). |
| `colony-lint.sh` | ✅ | — | 428 passed, 0 failed, 5 skipped. |
| Runtime end-to-end collector | ✅ | — | Synthetic feds: snapshot-refresh (10s)→healthy; cost-rate (200s)→silent; cost-rate (1000s)→down; absent log→silent/last_tick=None. No crash. Ceilings 240/120 (cost-rate) + 1200/600 (snapshot-refresh) match. |
| py_compile / no path leak | ✅ | — | Compiles clean; diff has no internal paths / private-repo refs. |

### Verdict
✅ **ship-it** — NameError risk cleared (both names in scope, confirmed at runtime), index stability green (orphan test), signature + template rendering correct, all gating tests pass, CHANGELOG correct. The federation also sharpened the framing (collector already had a `sidecars[]` array modelling auto-promote + cost-cap; the two always-on ones were the gap) and added test coverage.

ylohnitram:

The diff adds two always-on sidecar records (snapshot-refresh, cost-rate) to the collector, with interval_s sourced from a new SIDECAR_INTERVALS_S table and last_tick derived from the per-sidecar log mtime. No crash-class bugs found: fed_dir is a defined global (sys.argv[3]), the isfile/getmtime TOCTOU is caught by except OSError, and _sidecar_status() safely handles last_tick_ts=None (->silent) and interval_s=None (->degraded). The remaining concerns are correctness/edge-case issues with using log mtime as a liveness proxy: quiet ticks can yield false 'down' status (medium), a never-started always-on sidecar renders 'silent' rather than 'down' (low), and unknown sidecar names degrade silently (low).

Issue #440, opened 2026-05-05
tribes-bench: tools/kill-federation.sh selectivity — cascade kills dashboard during pilot orchestration


ylohnitram:
## Background

`tools/kill-federation.sh` is called automatically at end-of-pilot by
`tribes-bench/tools/run-stage2.sh` and `tribes-bench/tools/run-baseline.sh`
(and therefore also by the new `run-verdict-pair.sh` orchestrator landed
under follow-up to #436). Its current process-matching is wide enough
that any running `federation-dashboard` instance is also terminated
when the federation shuts down.

Operators report this as friction: their dashboard tab dies at the end
of every verdict pair and they don't know why. Documented operationally
in the project memory (`feedback_kill_federation_kills_dashboard.md`)
and now in `tribes-bench/README.md` under the new **Known gotchas**
section.

## Why this is tracked separately from #436

#436 polished the install + DX path for first-time tribes-bench
operators. Touching `tools/kill-federation.sh`'s process-matching has
wider blast radius — it runs against every federation in this repo and
every sidecar process, not just tribes-bench — so any selectivity
change needs its own design review and its own QA pass. Out of scope
for #436; tracked separately here.

## Proposed approaches

Either of:

A. **Selective process matching by `--fed-dir` argument.** Only kill
   processes whose cwd or daemon-record path is under the resolved
   `--fed-dir`. Dashboard is run-dir-agnostic so it would naturally be
   excluded.

B. **`--preserve-dashboard` flag.** Opt-in operator flag passed by
   `run-stage2.sh` / `run-baseline.sh` / `run-verdict-pair.sh` that
   excludes any process whose argv matches `federation-dashboard` or
   `agentis daemon-inner` running outside the federation's daemon dir.

A is cleaner (no flag proliferation) but requires verifying every
in-repo caller resolves `--fed-dir` correctly. B is mechanical but
adds surface to maintain.

## Acceptance

- A pilot operator running `bash tribes-bench/tools/run-verdict-pair.sh`
  with the standalone `federation-dashboard` already serving on
  `:8420` finds the dashboard still serving after the verdict pair
  completes.
- Existing tribes-bench tests still pass.
- `tools/colony-lint.sh` clean.

Follow-up to #436.

ylohnitram:
## Plan for #440 — kill-federation.sh dashboard-cascade selectivity

### Approach decision

[...]

PR #1241, opened 2026-06-21
fix(kill-federation): resolve symlinked daemon registry before find/tar

### LINKED ISSUE #1240: kill-federation.sh under-cleans registry + mis-backs-up when .agentis/daemon is a symlink (store-split)
## Symptom

On a store-split federation where `<fed>/.agentis/daemon` is a **symlink** to a canonical store (e.g. a federation whose colony dir symlinks `.agentis/daemon -> ../../.agentis/daemon`), `kill-federation.sh` reports `registry files: 0` and `ok federation fully stopped` while the canonical store still holds all the stale sidecar files. The next `start-federation.sh` then reads those stale `"state":"running"` records (via `agentis daemon list` cwd walk-up to the canonical store) and **refuses to start** (`A federation is already running under this directory`). Operator workaround today is `agentis daemon prune`.

Hit live during a v1.19.1 federation redeploy: kill reported 0 registry files; 126 stale sidecar files remained; start refused.

## Root cause

`find <symlink-to-dir>` without `-L` (or path resolution) does **not** descend into the symlink target — it sees the start point as a symlink, and with `-type f` matches nothing. Every registry-dir `find` in the script uses the literal `"$AGENTIS_DIR/daemon"`:

- registry-membership PID scoping (#440) — line ~321 → sees 0 `*.pid` → registry treated as "empty" → falls back to cwd-only scoping (so processes still die, but…)
- stale-registry clean (`find … -delete`) — line ~671 → deletes 0 files → registry never cleaned
- dry-run listing, remaining-count, JSON `registry_remaining` — lines ~535/642/676/678/730/732 → all read 0

Coupled second bug: the backup `tar czf "$BACKUP_PATH" -C "$AGENTIS_DIR" daemon` archives the **symlink itself**, not the target's files (tar doesn't dereference a symlink arg without `-h`), so under store-split the backup is empty too.

## Fix

Resolve the daemon registry dir to its real path **once**, right after `AGENTIS_DIR` is established, using a portable `cd … && pwd -P` (no `readlink -f` / `realpath` / python — macOS bash 3.2 safe):

```sh
DAEMON_DIR="$AGENTIS_DIR/daemon"
if [ -n "$AGENTIS_DIR" ] && [ -d "$DAEMON_DIR" ]; then
    DAEMON_DIR="$(cd "$DAEMON_DIR" 2>/dev/null && pwd -P || echo "$AGENTIS_DIR/daemon")"
fi
```

Then use `"$DAEMON_DIR"` at all 8 `find` sites, and back up via `tar czf … -C "$(dirname "$DAEMON_DIR")" "$(basename "$DAEMON_DIR")"` so the real files are archived. Non-split federations are byte-identical (`cd && pwd -P` on a normal dir returns its own real path; basename stays `daemon`); empty-`AGENTIS_DIR` standalone mode is unchanged.

## Test

Extend `tools/test-kill-federation.sh` with a regression case: fixture where `.agentis/daemon` is a symlink to a real dir holding `*.pid`/`*.heartbeat`/`*.colony` files; run `--fed-dir <fixture> --no-backup`; assert the real files are actually deleted (and the JSON `registry_remaining` is 0). The case fails on pre-fix code (0 found → 0 deleted → files remain).

SemVer: PATCH (platform tooling bug fix; non-split federations unaffected). Test-only + one tooling script — no `.ag`, no versioned-component change.

ylohnitram:
Closes #1240

## Problem

On a store-split federation, `<fed>/.agentis/daemon` is a **symlink** to a canonical daemon store. `find <symlink-to-dir>` without `-L` does not descend into the target — with `-type f` it matches nothing. Every registry-dir `find` in `kill-federation.sh` used the literal `"$AGENTIS_DIR/daemon"`, so:

- the stale-registry clean (`find … -delete`) deleted **0** files → registry never cleaned;
- the registry-membership PID scoping (#440) saw **0** `*.pid` → silently collapsed to cwd-only scoping;
- the dry-run listing, remaining-count, and JSON `registry_remaining` all read **0**.

Result: kill reports `registry files: 0` / `federation fully stopped` while the canonical store still holds every stale `"running"` record, so the next `start-federation` **refuses to start**. Coupled bug: `tar -C "$AGENTIS_DIR" daemon` archived the *symlink*, not its contents.

Hit live during a v1.19.1 redeploy: kill reported 0 registry files; 126 stale sidecar files remained; start refused. Operator workaround was `agentis daemon prune`.

## Fix

Resolve `$AGENTIS_DIR/daemon` to its **physical path** once, up front (`DAEMON_DIR` via `cd … && pwd -P`), and route all **8 find sites** plus the **backup tar** (`-C "$(dirname "$DAEMON_DIR")" "$(basename "$DAEMON_DIR")"`) through it. Non-symlink layouts are unaffected (`pwd -P` of a real dir is itself). The `[ -d "$AGENTIS_DIR/daemon" ]` guards are left as-is — `test -d` already follows symlinks.

## Tests

- New **Test 13** in `tools/test-kill-federation.sh`: builds a fixture whose `.agentis/daemon` is a symlink to a separate canonical store with real `*.pid`/`*.heartbeat`/`*.colony` files; asserts the canonical store is **cleaned through the symlink** (`registry_remaining = 0`) and the **backup tarball contains the real sidecar files** (proving tar followed the symlink).
- Full suite: **13/13 pass** (verified in a fresh net namespace so the host's live dashboard on :8420 doesn't skew the port-clean verification — the real-kill tests 5/8/13 assert a clean port).
- `colony-lint.sh`: **429 passed, 0 failed**.

### DIFF
--- a/tools/kill-federation.sh
+++ b/tools/kill-federation.sh
@@ -232,6 +232,20 @@ if [ ! -d "$AGENTIS_DIR" ]; then
 fi
 ok "federation dir = $FED_DIR"
 [ -n "$AGENTIS_DIR" ] && ok "agentis state  = $AGENTIS_DIR"
+
+# Resolve the daemon-registry dir to its PHYSICAL path. In a store-split
+# layout, $FED_DIR/.agentis/daemon is a SYMLINK to the canonical store. `find
+# <symlink-to-dir>` without -L does NOT traverse the symlink, so every registry
+# scan below would report 0 files and the registry would never be cleaned —
+# leaving stale "running" records that block the next start-federation. `tar -C
+# .agentis daemon` would likewise archive the symlink, not its contents.
+# Resolving through the symlink once, up front, fixes all of them (#1240).
+DAEMON_DIR=""
+if [ -n "$AGENTIS_DIR" ] && [ -d "$AGENTIS_DIR/daemon" ]; then
+    DAEMON_DIR="$(cd "$AGENTIS_DIR/daemon" 2>/dev/null && pwd -P || echo "$AGENTIS_DIR/daemon")"
+fi
+[ -n "$DAEMON_DIR" ] && ok "daemon registry = $DAEMON_DIR"
+
 [ "$DRY_RUN" -eq 1 ] && warn "DRY-RUN: no signals will be sent, no files deleted"
 
 # --- Collect PIDs (once, up front, with precise patterns) ---
@@ -318,7 +332,7 @@ if [ -d /proc ] && command -v readlink >/dev/null 2>&1; then
             # Each daemon writes a single-line file containing its PID.
             # `*.pid` covers both the primary (<id>.pid) and watchdog
             # (<id>.watchdog.pid) sidecars.
-            REGISTERED_PIDS_NL="$(find "$AGENTIS_DIR/daemon" -maxdepth 1 -name '*.pid' -type f 2>/dev/null \
+            REGISTERED_PIDS_NL="$(find "$DAEMON_DIR" -maxdepth 1 -name '*.pid' -type f 2>/dev/null \
                 | while IFS= read -r f; do
                     [ -z "$f" ] && continue
                     _pid="$(head -n 1 "$f" 2>/dev/null | tr -d ' \t\r\n' || true)"
@@ -532,7 +546,7 @@ if [ "$DRY_RUN" -eq 1 ]; then
                 [ -z "$f" ] && continue
                 echo "  would remove: $f" >&2
                 REG_FILES_LIST="$REG_FILES_LIST$f"$'\n'
-            done < <(find "$AGENTIS_DIR/daemon" -maxdepth 1 -name "*.$ext" -type f 2>/dev/null)
+            done < <(find "$DAEMON_DIR" -maxdepth 1 -name "*.$ext" -type f 2>/dev/null)
         done
     fi
     REG_COUNT=$(printf '%s' "$REG_FILES_LIST" | grep -c . || true)
@@ -639,7 +653,7 @@ if [ -n "$AGENTIS_DIR" ] && [ -d "$AGENTIS_DIR/daemon" ]; then
         # (only inbox/ subdirs or it's empty). Avoids leaving a ~100-byte
         # empty tarball behind on every nothing-to-clean run.
         _has_files=0
-        if find "$AGENTIS_DIR/daemon" -maxdepth 1 -type f -print -quit 2>/dev/null | grep -q .; then
+        if find "$DAEMON_DIR" -maxdepth 1 -type f -print -quit 2>/dev/null | grep -q .; then
             _has_files=1
         fi
         if [ "$_has_files" -eq 0 ]; then
@@ -653,7 +667,7 @@ if [ -n "$AGENTIS_DIR" ] && [ -d "$AGENTIS_DIR/daemon" ]; then
             # backup tarball. PID is preferred over `mktemp` to avoid
             # adding a dependency we don't already require.
             BACKUP_PATH="$BACKUP_DIR/agentis-daemon-registry-backup-$TS-$$.tar.gz"
-            if tar czf "$BACKUP_PATH" -C "$AGENTIS_DIR" daemon 2>/dev/null; then
+            if tar czf "$BACKUP_PATH" -C "$(dirname "$DAEMON_DIR")" "$(basename "$DAEMON_DIR")" 2>/dev/null; then
                 ok "backup: $BACKUP_PATH"
             else
                 warn "backup failed (continuing): $BACKUP_PATH"
@@ -668,14 +682,14 @@ if [ -n "$AGENTIS_DIR" ] && [ -d "$AGENTIS_DIR/daemon" ]; then
     # Use find -delete (not shell globs) — avoids 'no match' failures on
     # bash 3.2 / zsh when a registry happens to be empty.
     for ext in pid watchdog.pid colony heartbeat status stop; do
-        find "$AGENTIS_DIR/daemon" -maxdepth 1 -name "*.$ext" -type f -delete 2>/dev/null || true
+        find "$DAEMON_DIR" -maxdepth 1 -name "*.$ext" -type f -delete 2>/dev/null || true
     done
     # Use `find -printf '.'` rather than `wc -l` so a filename containing
     # a literal newline doesn't inflate the count. Fall back to `wc -l` if
     # `find -printf` is unavailable (e.g. BSD/macOS find).
-    REMAINING_FILES=$(find "$AGENTIS_DIR/daemon" -maxdepth 1 -type f -printf '.' 2>/dev/null | wc -c | tr -d ' ')
+    REMAINING_FILES=$(find "$DAEMON_DIR" -maxdepth 1 -type f -printf '.' 2>/dev/null | wc -c | tr -d ' ')
     if [ -z "$REMAINING_FILES" ]; then
-        REMAINING_FILES=$(find "$AGENTIS_DIR/daemon" -maxdepth 1 -type f 2>/dev/null | wc -l | tr -d ' ')
+        REMAINING_FILES=$(find "$DAEMON_DIR" -maxdepth 1 -type f 2>/dev/null | wc -l | tr -d ' ')
     fi
     ok "registry sidecar files removed (${REMAINING_FILES} regular files remain — inbox/ subdirs kept)"
 else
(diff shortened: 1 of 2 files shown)


ylohnitram:
[draft-review] **Security Review** (automated, pending approval)

Reviewed a two-file shell-script change (kill-federation.sh registry cleanup plus its test). The code is quoting-clean: no command injection, eval, hardcoded secrets, SQL/XSS/CSRF/SSRF/deserialization, or dependency changes, and CLI-derived paths are consistently double-quoted. The one security-relevant behavior introduced is intentional symlink dereferencing (pwd -P) before `find -delete` and `tar` on the daemon registry, which enables a low-severity local symlink-following risk (CWE-59) if the .agentis/daemon symlink is attacker-writable; recommend a containment/realpath check on the resolved path. No high or critical issues found.

ylohnitram:

Test 13 is a solid, targeted regression test for the #1240 happy path: it proves the registry is cleaned and the backup tarball is produced through the store-split symlink. However coverage is incomplete relative to the diff. The DRY-RUN code path was also changed to use $DAEMON_DIR but is not exercised. The explicitly-guarded edge cases the new code introduces are untested: an empty symlinked registry (backup-skip branch) and a broken/dangling symlink (pwd -P fallback). The backup-content assertion is also weak, checking only for one .pid entry rather than verifying all created sidecars were archived. No flakiness or test-duplication concerns were found. Recommend adding dry-run, empty-registry, and dangling-symlink cases and strengthening the tarball-content assertion.

ylohnitram:

The change correctly fixes #1240. Resolving $AGENTIS_DIR/daemon to its physical path via `cd ... && pwd -P` before every find/tar makes registry scans, deletes, and the backup traverse the store-split symlink instead of silently treating it as a 0-file directory (which previously left stale 'running' records blocking the next start-federation). DAEMON_DIR is initialized to empty and re-resolved under the same `[ -d "$AGENTIS_DIR/daemon" ]` condition that guards every cleanup/delete block, so it is guaranteed non-empty wherever deletion occurs, and all uses are quoted, so there is no accidental cwd scan or mass-delete. The `cd ... 2>/dev/null && pwd -P || echo "$AGENTIS_DIR/daemon"` fallback is sound, and the broken-symlink case (-d false) cleanly skips the block. No off-by-one, null-deref, race condition, boundary, resource-leak, or wrong-operator defects found. Two low-severity items only: the backup tarball's internal path now reflects the canonical-store basename rather than 'daemon' (restore-semantics change), and the /proc PID scan uses DAEMON_DIR without a non-empty guard (currently safe purely due to quoting).

ylohnitram:

Two minor, low-severity style nits. The diff is otherwise clean: consistent 4-space indentation, clear and well-formatted block comments, and the new DAEMON_DIR variable follows the existing UPPERCASE script-global naming convention. The only style issues are a broken `=` column alignment in a new status message and mixed casing of path variables in the new test.

ylohnitram:
## QA — PR #1241 (fix: resolve symlinked daemon registry before find/tar, #1240)

End-to-end QA on the PR branch. Verified diff correctness, non-symlink regression safety, portability, and ran the full test harness + lint.

### Results

| Diff: DAEMON_DIR resolved only under `[ -d "$AGENTIS_DIR/daemon" ]`, empty otherwise | ✅ | — | `DAEMON_DIR=""` default; resolved via `cd … && pwd -P` with `|| echo "$AGENTIS_DIR/daemon"` fallback inside the existence guard. |
| Diff: all `find "$AGENTIS_DIR/daemon"` sites converted to `find "$DAEMON_DIR"` | ✅ | — | `grep` confirms 0 remaining `find "$AGENTIS_DIR/daemon"` sites; 8 `find "$DAEMON_DIR"` sites present. Surviving `AGENTIS_DIR/daemon` refs are only the resolution block, comments, the `[ -d ]` guards, and one `warn` message. |
| Diff: backup tar uses `-C "$(dirname "$DAEMON_DIR")" "$(basename "$DAEMON_DIR")"` | ✅ | — | Replaces `-C "$AGENTIS_DIR" daemon`; archives the physical contents, not the symlink. |
| Diff: every `find`/`tar` site executes inside a `[ -d "$AGENTIS_DIR/daemon" ]` guard (so DAEMON_DIR is always set) | ✅ | — | All 4 blocks (PID scan, dry-run list, backup+cleanup, verification) are guarded. |
| Non-symlink (normal) layout — no regression | ✅ | — | `pwd -P` of a real dir returns the same path (verified empirically); behaviour identical to pre-fix for the common case. |
| Portability — no GNU-only flags / new bashisms introduced | ✅ | — | `dirname`/`basename`/`pwd -P` are POSIX; the `-printf` lines already carried their `wc -l` fallback pre-fix; the one `<( )` process-sub is a pre-existing line (only the path variable changed). `shellcheck -S warning` clean on both files (rc=0). |
| `bash -n` syntax check (both files) | ✅ | — | Both parse clean. |
| Test harness — direct run (shared host) | ⚠️ | Environment artifact, dismissed — see note + clean-namespace row | A live dashboard occupies port 8420, so the real-kill tests' "fully clean" verification reports `rc=1` → tests 5, 8, 13 FAIL. For test 13 the fix-specific assertions still pass on their own (`canon_left=0`, `backup_has_files=1`); only the shared-port `rc` flips it. Not a code failure. |
| Test harness — clean network namespace (`unshare -rn`) | ✅ | — | **13 passed, 0 failed**, including the new Test 13 (store-split symlink cleaned + backed up through the symlink). The "Killed … line 358/406" lines are the test's own survivor-fixture teardown (tests 10/11), both PASS. |
| `colony-lint.sh` | ✅ | — | **429 passed, 0 failed, 5 skipped.** Kill tests correctly skipped by default (set `AGENTIS_RUN_KILL_TESTS=1` to include). |
| `test-kill-endpoint.sh` in namespace (untouched by PR) | ⚠️ | Dismissed — pre-existing environment artifact, not a regression | Fails because `unshare -rn` brings up `lo` as `state DOWN`, so the dashboard server can't bind/serve `localhost`. The script is not in the PR diff; failure is a loopback artifact, not caused by the change. |
| Backup tarball internal top-level dir name changes in store-split case | ⚠️ | Dismissed — benign, no consumer depends on it | Pre-fix tarballs always rooted at `daemon/`; post-fix they root at `basename($DAEMON_DIR)` (the canonical store's basename) for the store-split layout. No automated restore path exists — the only repo consumers count tarballs by filename glob. It's a manual operator-recovery artifact, so the changed inner dir name is cosmetic. |

### Verification commands

```
# diff audit
git diff <base>...<head> -- tools/kill-federation.sh
grep -n 'find "$AGENTIS_DIR/daemon"' tools/kill-federation.sh   # -> none
bash -n tools/kill-federation.sh && bash -n tools/test-kill-federation.sh
shellcheck -S warning tools/kill-federation.sh tools/test-kill-federation.sh   # rc=0

# harness (host shows the port-8420 artifact)
./tools/test-kill-federation.sh                          # 9 passed, 4 failed (port 8420 in use)
unshare -rn bash -c './tools/test-kill-federation.sh'    # 13 passed, 0 failed

# lint
./tools/colony-lint.sh                                   # 429 passed, 0 failed, 5 skipped
```

### Verdict

✅ ship-it — fix is correct and complete (all 8 find sites + the backup tar routed through the resolved physical path, all guarded), no regression for the normal layout, portable, clean lint, and 13/13 harness pass in a clean network namespace. The only ⚠️ rows are environment artifacts (shared port 8420 / namespace loopback down) and one benign cosmetic note, all dismissed with reason.

ylohnitram:

Test 13 is a well-targeted, discriminating regression test for the store-split symlink fix (#1240) and would correctly fail on the pre-fix code. The main gap is assertion strength and coverage: the backup check only looks for a *.pid entry though three sidecar types are created, and the symlink-routed empty-registry, dangling-symlink/fallback, normal-layout, and symlink-preservation paths are untested. No serious flakiness was found; severities range low to medium.

ylohnitram:
[draft-review] **Security Review** (automated, pending approval)

No MR IID was supplied in the payload (the only identifier present is referenced issue #1240); mr_iid is reported as 0 to signal 'not provided'. The change set is limited to two POSIX shell scripts (kill-federation.sh and its test) in a local federation kill/cleanup tool. No SQL/command injection, XSS, CSRF, hardcoded secrets, insecure auth, SSRF, or insecure deserialization were introduced: all variables and command substitutions are properly double-quoted and passed as arguments (no eval), tar is only created/listed (never extracted), and no new dependencies are added. The only security-relevant behavior change is that destructive `find -delete` and the `tar` backup now follow the .agentis/daemon symlink to its physical target (pwd -P). This is the intended bug fix (#1240) and is low risk in the normal single-user case, but on shared or world-writable paths it is a link-following / TOCTOU concern (CWE-59, CWE-367) because the guard tests the symlink while destructive actions run against the resolved physical directory. Recommend bounding the resolved path to a trusted root and hardening directory/symlink ownership and permissions.

PR #1243, opened 2026-06-21
feat(code-edit-in-checkout): forge-agnostic GitLab parity

ylohnitram:
Closes #1213

## Problem

`tools/code-edit-in-checkout.sh` hardcoded `github-api.sh` + GitHub clone-URL math and ignored `FORGE_TYPE`, so the checkout-edit path (Approach A) only worked for GitHub. GitLab repos couldn't use it — a gap for the GitLab target.

## Fix

Make the orchestrator forge-agnostic via `FORGE_TYPE` (default `github`). The colony's `gitlab-api.sh` already ships a forge-symmetric `create-mr` verb (same `--source/--title/--description` flags, reads `GITLAB_DEFAULT_BRANCH`), so this is orchestrator-side plumbing only:

- **FORGE_TYPE** resolved + validated (`github`|`gitlab`, else exit 2) + `export`ed so the `GIT_ASKPASS` helper can branch on it.
- **Token presence check** branches: `GITLAB_TOKEN` (gitlab) vs `GITHUB_TOKEN` (github).
- **Clone-URL host math** picks the API base per forge (`GITHUB_URL` default `api.github.com`; `GITLAB_URL` default `gitlab.com`); the existing `scheme://host` else-branch already maps `gitlab.com/api/v4 → gitlab.com` and forwards `file://` verbatim for the test harness.
- **GIT_ASKPASS helper** branches: username `oauth2` + `$GITLAB_TOKEN` (gitlab) vs `x-access-token` + `$GITHUB_TOKEN` (github). The token is read **only inside the helper** (a subprocess) from the env — never named as a word in the parent script (no `set -x` leak).
- **create-mr backend**: `gitlab-api.sh` vs `github-api.sh`. GitLab env: `GITLAB_URL` (instance), `GITLAB_PROJECT=owner%2Frepo` (URL-encoded), `GITLAB_DEFAULT_BRANCH`; `GITLAB_TOKEN` inherited (never named).
- **PR-URL parse** also reads `web_url` (the GitLab MR field).

### Scope notes
- Per-repo token resolution (#1212) stays **github-only** (already guarded by `FORGE_TYPE = github`); GitLab multi-repo is a future follow-up.
- The GitLab clone-auth username (`oauth2`) is exercised only against real HTTPS GitLab — a `file://` remote needs no auth — same status as GitHub's `x-access-token`.
- `GITLAB_PROJECT` is encoded as `owner%2Frepo` (2-segment); nested GitLab group paths would need fuller encoding (out of scope).

[...]

PR #1245, opened 2026-06-21
test(sse-stream): drop redundant nested test-timeline-rendering re-run

### LINKED ISSUE #1244: test-sse-stream.sh flakes under concurrent/loaded test runs (SSE readiness/timing race)
## Symptom

During QA of PR #1243, a `colony-lint.sh` run reported `test-sse-stream.sh` failing (its embedded timeline suite: `62 passed, 1 failed`, test 8). The run was concurrent with other test activity + the live `:8420` dashboard on the host.

## Not reproducible in isolation

- `tools/test-timeline-rendering.sh` standalone: **63 passed, 0 failed** on 3/3 consecutive runs (clean checkout).
- PR #1243 CI `colony-lint`: **green**.
- My own `colony-lint.sh` runs on the same branch earlier: **429 passed, 0 failed**.

So it is a transient flake, not a logic failure, and not caused by any PR (the only red observation was a loaded, concurrent local run).

## Suspected root cause

`tools/test-sse-stream.sh` spins up a real dashboard SSE server on an ephemeral free port and exercises it with tight timing windows: a ~250 ms readiness poll loop and `curl -N --max-time 6` against `/events`. Under heavy concurrent load (multiple `colony-lint` invocations + the live dashboard competing for CPU/IO), the server-readiness probe or the 6 s SSE capture window can be missed, so the embedded timeline assertion (`test-sse-stream.sh` "test 8" re-runs the timeline suite per the note at `tools/test-timeline-rendering.sh:2463`) sees incomplete output and goes red.

## Proposed fix

Make the SSE harness robust under load: lengthen/loop the server-readiness probe (poll `/` until 2xx with a generous overall deadline before any SSE capture), raise the `curl --max-time` for the `/events` capture, and/or retry the single timing-sensitive frame assertion once. No production code change — test-infra only.

## Scope

Pre-existing test-infra flake; out of scope for PR #1243 (which only touches `tools/code-edit-in-checkout.sh`). Filed per the no-pre-existing-flake-without-a-follow-up policy.

ylohnitram:
Closes #1244

## Problem

`tools/test-sse-stream.sh` test 8 shelled out and re-ran the **entire 63-test `test-timeline-rendering.sh` suite** as a nested child (a leftover guard from the one-time IIFE→renderXxx refactor). But `colony-lint.sh`'s tools-test loop **already** runs `test-timeline-rendering.sh` directly, so under colony-lint that suite executed **twice**. Under concurrent/loaded execution, one of its assertions flaked in the nested invocation — surfacing as `62 passed, 1 failed` once during QA of #1243, making a full `colony-lint.sh` run spuriously red. In isolation the suite is **63/0** (3/3 runs) and PR CI is green.

## Fix

Remove the redundant nested re-run (t8). Re-running a tile-rendering suite inside an *SSE-streaming* test is off-scope, and the double-execution only amplified load on a suite that has a marginal under-contention flake. Tile-rendering correctness remains owned by `test-timeline-rendering.sh`, which colony-lint still runs directly — **zero coverage loss**. SSE-streaming behaviour stays covered by t1–t7.

## Tests

- `tools/test-sse-stream.sh`: **8 passed, 0 failed** (t1–t7; t8 removed).
- `colony-lint.sh`: **429 passed, 0 failed, 5 skipped** — and `test-timeline-rendering.sh` now runs once per lint instead of twice.

### DIFF
--- a/tools/test-sse-stream.sh
+++ b/tools/test-sse-stream.sh
@@ -20,8 +20,10 @@
 #       host divs on the new Logs & Events tab.
 #   t7: PR 2 — bootstrap shape: single rerender(window.__data) call,
 #       no leftover anon IIFE renderers, window.__data hung off window.
-#   t8: PR 2 — test-timeline-rendering.sh stays green (29/0 baseline)
-#       under the refactor.
+#   (t8 removed in #1244 — it re-ran the full test-timeline-rendering.sh suite,
+#    which colony-lint already runs directly; the nested re-run only amplified
+#    that suite's load and flaked under concurrency. Tile-render coverage is
+#    unchanged — it is owned by test-timeline-rendering.sh on its own.)
 #
 # Self-contained: only bash, python3, curl. Cleans up via trap on every
 # exit path. Auto-skips when python3 or a free port is unavailable.
@@ -314,28 +316,12 @@ else
     fail "7: bootstrap shape regression: $T7_MSG"
 fi
 
-# t8: validate that test-timeline-rendering.sh still passes (the IIFE
-#     refactor must not break any of the 29 existing tile-rendering
-#     tests). We invoke it as a child and grep for "0 failed" — if it
-#     fails, this top-level test fails too. Skip cleanly when the
-#     timeline test isn't reachable (CI shape change).
-T8_HARNESS="$REPO_ROOT/tools/test-timeline-rendering.sh"
-if [ -x "$T8_HARNESS" ] || [ -r "$T8_HARNESS" ]; then
-    T8_OUT="$TMPDIR_TEST/timeline-rendering.log"
-    if bash "$T8_HARNESS" >"$T8_OUT" 2>&1; then
-        if grep -q '0 failed' "$T8_OUT"; then
-            pass "8: test-timeline-rendering.sh stays green after IIFE -> renderXxx refactor"
-        else
-            fail "8: test-timeline-rendering.sh exited 0 but reported failures" \
-                 "tail: $(tail -3 "$T8_OUT" | tr '\n' ' ')"
-        fi
-    else
-        fail "8: test-timeline-rendering.sh failed under the refactor" \
-             "tail: $(tail -3 "$T8_OUT" 2>/dev/null | tr '\n' ' ')"
-    fi
-else
-    skip "8: test-timeline-rendering.sh not reachable"
-fi
+# t8 (re-running test-timeline-rendering.sh) was REMOVED in #1244. colony-lint's
+# tools-test loop already runs test-timeline-rendering.sh directly, so re-running
+# its full 63-test suite here was redundant — and it doubled that suite's load
+# under colony-lint, flaking one assertion under concurrent runs (62/1 once,
+# 63/0 in isolation). SSE-streaming behaviour is covered by t1-t7 above;
+# tile-rendering correctness is owned by test-timeline-rendering.sh on its own.
 
 echo ""
 echo "Results: $PASS passed, $FAIL failed"


ylohnitram:

Test-only change that removes t8 from test-sse-stream.sh, a nested re-run of the full test-timeline-rendering.sh suite. The removal is sound: the nested run duplicated coverage colony-lint already provides and flaked under concurrency, and SSE behaviours (t1-t7) remain covered. Remaining concerns are minor: regression coverage for the IIFE refactor now relies on an external, unenforced assumption (colony-lint invoking the timeline suite) with no in-repo guard, and the file's comments contradict each other on suite size (29 vs 63). No high-severity test gaps introduced.

ylohnitram:

The diff removes test t8 (a redundant nested re-run of test-timeline-rendering.sh) from tools/test-sse-stream.sh and replaces it with explanatory comments. The removal is syntactically sound and introduces no executable logic bug. The only correctness-relevant note is that it suppresses an acknowledged concurrency flake (62/1 vs 63/0) rather than fixing the underlying race in test-timeline-rendering.sh; that suite stays covered by colony-lint, but the non-determinism remains unaddressed.

ylohnitram:

Comment style (# prefix, em-dashes, continuation indentation), formatting, whitespace and naming are all consistent with the surrounding file; no violations in the code itself. The only notes are organizational: the deleted test is replaced by a verbose tombstone comment, and its removal rationale is duplicated between the header block and the inline location.

ylohnitram:
## QA — #1245 (drop redundant nested timeline re-run from `test-sse-stream.sh`)

Reviewed the PR branch against `origin/main`. The change removes the former **t8** block from `tools/test-sse-stream.sh`, which re-ran the entire `tools/test-timeline-rendering.sh` suite (63 tests) as a nested child, and updates the header comment to explain why.

### Findings

| Diff scope | ✅ | — | Only `tools/test-sse-stream.sh` touched (10 insertions / 24 deletions). No production code, no other test file. t1–t7 (incl. t1b) are byte-for-byte unchanged. |
| No dangling vars | ✅ | — | `T8_HARNESS` / `T8_OUT` fully removed. `REPO_ROOT` is still defined and still used (`DASHBOARD_SH="$REPO_ROOT/federation-dashboard/bin/federation-dashboard"`), so it must stay — correctly retained. Only surviving `t8` mentions are in explanatory comments. |
| Zero-coverage-loss claim (KEY) | ✅ | — | colony-lint's tools-test discovery loop `find`s every `tools/test-*.sh` and runs each via the default branch. The only diversions are the two kill-tests and `test-boot-smoke.sh` — `test-timeline-rendering.sh` is **not** on any skip-list, so it is run directly by the lint. Tile-render coverage is unchanged; the nested re-run was pure duplication. |
| `test-timeline-rendering.sh` standalone | ✅ | — | `63 passed, 0 failed` on its own. |
| `test-sse-stream.sh` direct | ✅ | — | `8 passed, 0 failed` (t1, t1b, t2–t7). SSE-streaming + IIFE-refactor-shape coverage intact; spins a dashboard on a free port and runs directly. |
| colony-lint full | ✅ | — | `429 passed, 0 failed, 5 skipped`. Both `test-sse-stream.sh` **and** `test-timeline-rendering.sh` show up as discovered unit-test passes in the same run. |
| Skips are benign | ✅ | — | 5 skips: two "no colonies found" top-level dirs, `test-boot-smoke.sh` (opt-in), and the two destructive kill-tests (opt-in). None caused by this PR. |
| `bash -n` | ✅ | — | Syntax clean on the changed file. |
| Concurrency-flake claim | ✅ | — | The suite that was being doubled (`test-timeline-rendering.sh`) is green both standalone and when invoked by colony-lint in the same pass that also runs `test-sse-stream.sh` — i.e. removing the second concurrent invocation is exactly the load reduction described, with no coverage cost. |

### Verdict

✅ ship-it — change is a clean, scoped test-only dedup; no coverage lost (timeline suite is run directly by colony-lint), all targeted suites green (sse-stream 8/0, timeline 63/0, colony-lint 429/0), `bash -n` clean.

ylohnitram:

The MR removes a genuinely flaky, duplicative test (t8 re-ran the entire 63-test test-timeline-rendering.sh suite nested inside test-sse-stream.sh, flaking under concurrency), which is a sound test-quality improvement. Two residual concerns: (1) the underlying concurrency/order-dependent assertion in test-timeline-rendering.sh is only amplified-away, not root-caused, so it can still flake where the suite runs directly; and (2) tile-render coverage now implicitly depends on colony-lint's tools-test loop with no enforcement that the suite stays wired in. Recommend fixing the flaky assertion at its source and adding a guard that the timeline suite remains in CI.

Issue #1254, opened 2026-06-22
operationalize task_decomposer: split epics into single-shot subtasks on one branch [complex-task M3]


ylohnitram:
Final milestone of the complex-task arc — **M3**, builds on M1 (#1251 iterate) + M2 (#1253 verify-and-fix).

## Problem

Even with M1 (iterate) + M2 (verify-and-fix), a genuinely large epic is one monolithic editing target — claude does best on small, well-scoped changes (proven: #1117, #1246). `task_decomposer.ag` already *proposes* subtask breakdowns but is **observe/suggest only** — nothing turns a big issue into a sequence of small, individually-tractable edits.

## Fix — decompose, then implement subtasks sequentially on one branch

When `code_writer` picks up an **epic-labeled** (or over-threshold) issue:

1. **Decompose**: ask the model for an ordered list of small, self-contained sub-edits (each the size of a proven single-shot task), with a one-line goal each.
2. **Implement each subtask** through the existing M1+M2 code-edit path, **sequentially on the SAME per-issue branch** — each subtask's edit accumulates as commit(s); M1 lets each finish, M2 verifies each before moving on.
3. **One PR per epic** carrying the ordered sub-commits (simpler than multi-PR; no cross-PR dependency graph).
4. **Stop conditions**: a subtask that can't be completed/verified after its budget halts the sequence with a clear status (partial PR + note), rather than charging ahead.

Keeps the blast radius small per step and plays to the model's strength (small edits), while the federation still ships one coherent PR for the epic.

## Open design questions (resolve during impl)
- Decomposition trigger: explicit `epic` label vs a size heuristic (issue body length / file-count estimate).
- Where the subtask list lives (memo vs job-dir file) and how a resumed/retried epic picks up mid-sequence.
- Per-subtask commit granularity vs one squashed commit at the end.

## Test
Stub-driven: an epic task decomposes into 3 sub-edits; assert each ran through the code-edit path in order on one branch, the PR carries all three, and a mid-sequence verify-failure halts cleanly with a partial-progress status.

Issue #1185, opened 2026-06-20
Federation pickup operability on a mature repo: trigger-label env passthrough, learn-backlog, half-completed-flow retry


ylohnitram:
## Problem (found dogfooding the federation on this repo)
Getting the federation to autonomously pick up + work an assigned issue on a *mature* repo (real history, non-default labels) hit a stack of operational papercuts, separate from the capability fixes (#1171/#1172/#1181). The pickup capability is correct, but it is hard to operate:

1. **Trigger-label override does not reach the forge calls.** `exec.env_passthrough = COLONY_DIR,FORGE_TYPE,GITLAB_*,GITHUB_*` does **not** include `IMPLEMENTATION_TRIGGER_LABEL` / `PLANNING_TRIGGER_LABEL`, so a `colony.toml` `trigger_label` override is stripped when `code_writer`/planning agents `exec sh forge-api.sh assigned-issues` — the forge verb falls back to the default `implementation` / `needs-planning`. In practice you MUST label issues with the default trigger label; the override is silently inert at the exec boundary. Add the two trigger-label env vars to `exec.env_passthrough` (install.sh + the doc), or have the forge wrappers read the label from a passed-through var.

2. **`learn-from-MR` backlog grinds on a cleared `last_check`.** With `code_writer:last_check` empty, the step-1 `merge-requests --since` learning query walks the whole merged-MR history (one `Learned from MR <N>` per tick) before step-2 drafting gets useful air. On a mature repo, seed `last_check` to ~now on first run, or cap the learning window.

3. **Half-completed autonomous flow is not retried.** `code_writer` created the branch (`create-branch` OK) but the `commit-files` edit did not land, and the #200 staleness gate (`last_drafted_iid`) then marked the issue drafted, so it never retried — leaving an empty branch. The branch-then-commit steps should be atomic-ish (roll back the branch on a failed commit, or don't set the drafted marker until the commit succeeds), which also relates to the open-PR-skip / queue-advance follow-up from #1181.

## Done
- A plain issue **assigned** to the bot, with the repo's default trigger label, is picked up + worked end-to-end on a mature repo without per-run memo/env surgery.
- `trigger_label` overrides are honoured at the exec boundary (or the requirement to use the default label is documented).
- A half-completed draft (branch but no commit) is retried, not stranded.

PR #1258, opened 2026-06-22
feat(code_writer): auto-pass --decompose for epic-labeled issues

### LINKED ISSUE #1257: code_writer: auto-pass --decompose for epic-labeled issues (operationalize M3)
Follow-up to #1254 (M3). The decompose **capability** shipped: `code-edit-in-checkout.sh --decompose` splits an epic into per-subtask edits on one branch, and `code-edit-job.sh` forwards the flag. What remains is the **policy/trigger**: `code_writer.ag` should pass `--decompose` automatically when the issue it picks up is epic-class.

## Task
In `dev-apprenticeship/implementation/agents/code_writer.ag`, when building `job_cmd` (the `code-edit-job.sh` call), append ` --decompose` iff the assigned issue carries the epic label.

## Design notes
- Epic label value: the planning colony seeds `planning:labels:epic` (from `[planning.labels].epic`, default `epic`) via `agentis memo set`. code_writer is in the implementation colony — confirm it can read that memo (cross-colony `recall_latest`) or fall back to the literal `epic`. A small `is_epic(labels_json)` helper that substring-checks the issue's `labels` (from `assigned-issues [0].labels`) against the configured epic label.
- Keep it conservative: only trigger on a clear epic label; non-epic issues stay single-shot (the proven path).
- Test: extend the code_writer tier-branch / `.ag` coverage so an epic-labeled fixture appends `--decompose` and a normal issue does not. colony-lint runs `.ag` syntax + tier checks.

## Why separate from #1254
The capability (federation-agnostic orchestrator + launcher, fully tested) is isolated from this policy change on the **live** code_writer agent, which carries more runtime risk and deserves its own focused review/QA.

ylohnitram:
Closes #1257 · operationalizes M3 (#1254) — the policy trigger on top of the decompose capability.

## What

When `code_writer` picks up an **epic-class** assigned issue, it now appends `--decompose` to the `code-edit-job.sh` launch, so the orchestrator splits the epic into per-subtask edits on one branch → one PR (the M3 mechanism). Non-epic issues stay single-shot.

## How

- Epic label = the planning vocabulary `recall_latest("planning:labels:epic")` (read **cross-colony** — the same memo `task_decomposer` already reads), or `"epic"` by default.
- Matched as an **exact quoted JSON array element** (`"epic"`) on the issue's labels (`issues_raw[0].labels`), so substrings like `not-epic` / `epic-foo` don't false-match.
- The flag is a constant appended **after** the `shell_escape`'d args (no injection surface); the existing `// colony-lint: safe-exec-concat` annotation still applies.
- Low-stakes by design: decomposing a non-epic is harmless (the decompose drive yields one subtask = the whole task ≈ normal), and not decomposing an epic just keeps the prior single-shot behaviour.

## Validation

`colony-lint.sh`: **`code_writer.ag` syntax OK + tier-branch convention OK**; **430 passed, 0 failed**. The `--decompose` passthrough through `code-edit-job.sh` is covered by that script's scenario E (#1254). (`.ag` agents have no unit-test harness — colony-lint's syntax + tier checks are the gate, plus the launcher/orchestrator tests already merged.)

### DIFF
--- a/dev-apprenticeship/implementation/agents/code_writer.ag
+++ b/dev-apprenticeship/implementation/agents/code_writer.ag
@@ -417,8 +417,21 @@ fn tick_for_repo(owner: string, repo: string) -> void {
                 // slow job, so this `exec sh` returns fast and never hits the
                 // 120s cap. A non-zero exit (a usage / launch error) collapses
                 // the `try` to "" => job_word "" => the ERROR/else branch retries.
+                // Epic auto-trigger (#1257): an epic-class issue gets --decompose so
+                // the orchestrator splits it into per-subtask edits (M3, #1254). The
+                // epic label is the planning vocabulary (planning:labels:epic, read
+                // cross-colony) or "epic" by default, matched as an EXACT quoted JSON
+                // array element on the issue's labels so "not-epic"/"epic-foo" don't
+                // false-match. Decomposing a non-epic is harmless (one subtask = the
+                // whole task), so a loose match here is low-stakes.
+                let epic_vocab = recall_latest("planning:labels:epic");
+                let epic_label = if len(epic_vocab) > 0 { epic_vocab; } else { "epic"; };
+                let labels_str = to_string(json_get(issues_raw, "[0].labels"));
+                let is_epic = index_of(labels_str, "\"" + epic_label + "\"") > 0;
+                if is_epic { print("[code_writer] issue", issue_iid, "is epic-class — auto-decomposing (#1257)"); };
+                let decompose_flag = if is_epic { " --decompose"; } else { ""; };
                 // colony-lint: safe-exec-concat
-                let job_cmd = "$COLONY_DIR/../../tools/code-edit-job.sh --owner " + shell_escape(owner) + " --repo " + shell_escape(repo) + " --issue " + shell_escape(issue_iid) + " --branch " + shell_escape(branch_name) + " --title " + shell_escape(draft.summary) + " --task " + shell_escape(task_text);
+                let job_cmd = "$COLONY_DIR/../../tools/code-edit-job.sh --owner " + shell_escape(owner) + " --repo " + shell_escape(repo) + " --issue " + shell_escape(issue_iid) + " --branch " + shell_escape(branch_name) + " --title " + shell_escape(draft.summary) + " --task " + shell_escape(task_text) + decompose_flag;
                 let job_out = try { exec sh job_cmd; } catch e {
                     print("[code_writer] code-edit-job launcher failed — not marking drafted, will retry:", e);
                     "";


ylohnitram:

This MR introduces new epic auto-trigger branching (label detection, vocab fallback, and --decompose flag composition) with no tests at all. The most important gaps are the complete absence of coverage for epic vs non-epic paths, the untested `> 0` boundary (potential off-by-one), the unverified exact-match guarantee, the epic_label default fallback, empty/missing labels edge cases, and the decompose_flag command composition. Recommend adding focused unit tests for detection, fallback, negative matches, boundary, and final job_cmd assertions before merge.

ylohnitram:

The epic auto-trigger logic is functionally low-risk (the comment correctly notes a wrong match only causes a harmless single-subtask decompose), but it has three correctness concerns: a boundary check using `> 0` where `>= 0` is the correct test for index_of, a hardcoded `[0]` label lookup that can read a different issue than the one being launched, and unguarded nil/empty results from recall_latest and the json_get label path. None are critical; the `[0]` index mismatch is the most worth fixing so the decompose decision reflects the actual target issue.

ylohnitram:
## QA — PR #1258 (closes #1257): auto-pass `--decompose` for epic-labeled issues

Single-file `.ag` change to `dev-apprenticeship/implementation/agents/code_writer.ag` (+14 / -1). No unit-test harness for agent logic exists; the gate is careful diff review + reasoning + `colony-lint` (syntax + tier-branch). Verified against `origin/main...HEAD`.

| Logic correctness — epic detection | ✅ | — | `epic_vocab = recall_latest("planning:labels:epic")`; `epic_label = epic_vocab if len>0 else "epic"`; `labels_str = to_string(json_get(issues_raw, "[0].labels"))`; `is_epic = index_of(labels_str, "\"" + epic_label + "\"") > 0`. Quoted-element match means `not-epic`/`epic-foo` do NOT false-match; a real `epic` array element sits at index ≥1 (labels_str starts with `[`), so `> 0` is correct and the not-found sentinel is excluded. Empty memo → `"epic"` default. |
| `issues_raw` in scope at use site | ✅ | — | Defined from the `assigned-issues` poll, then used for `first_iid_str`/`first_upd` a few lines above; the new block is the same `[0]` element. Guarded by `if len(issues_raw) > 10`, so `[0].labels` indexing is safe (never empty). New block is nested inside `should_draft_code(...)` → `my_tier == "autonomous"`. |
| Cross-colony recall is an established pattern | ✅ | — | `planning:labels:epic` is seeded by `planning/scripts/start-colony.sh` (`agentis memo set planning:labels:epic "$EPIC_LABELS"`) and already read cross-colony by `planning/agents/task_decomposer.ag`. code_writer reading it is not a new assumption. |
| No regression / additive only | ✅ | — | Only ADDS lines before `job_cmd` and appends `+ decompose_flag`. draft / should_draft / #1185 markers / learn flow untouched. `decompose_flag` is defined on BOTH `if` branches, so the concat can never reference an unset var. |
| Injection safety | ✅ | — | `decompose_flag` is a literal (`" --decompose"` or `""`) appended AFTER all `shell_escape()`'d args — no user data, no injection vector. The `// colony-lint: safe-exec-concat` annotation remains directly above the `job_cmd` line. |
| `.ag` builtins are real | ✅ | — | `recall_latest`, `json_get`, `to_string`, `len`, `index_of`, `print`, `if`-expression, `\"` concat all used elsewhere. `index_of` precedent: `triage/agents/labeler.ag`. `let x = if ... { } else { }` precedent: `code-review/agents/approval_decider.ag` (and many others). |
| colony-lint (real exit) | ✅ | — | `exit=0`, `Results: 430 passed, 0 failed, 5 skipped`. Explicit `code_writer.ag syntax OK` + `code_writer.ag: tier-branch convention OK`. The lone `[WARN]` (plaintext tokens in `colony.example.toml`) is a pre-existing baseline warning on files this PR does not touch — not a failure, not introduced here. |
| Multi-token vocab edge | ⚠️ | Out of scope for #1257; follow-up only. Shipped default `epic = "epic"` is a single token and works; severity LOW. | If an operator sets `[planning.labels] epic = "epic, big-feature"` (CSV), code_writer's mechanical `index_of` match looks for the literal `"epic, big-feature"` as one quoted array element and never splits the CSV — so `is_epic` stays false and the issue is NOT auto-decomposed. `task_decomposer.ag` consumes the same memo via a `prompt()` context where the LLM can split it, so the two consumers diverge for multi-token configs. Failure mode is safe (missed auto-decompose, never a false-positive decompose; manual `--decompose` still works). Worth a follow-up to split-and-OR the CSV for parity, but does NOT block this PR. |

### Verdict
✅ ship-it — logic is correct for the default and single-label configs, additive-only with no regression, injection-safe, all builtins real, and colony-lint is green (430 passed / 0 failed, explicit `code_writer` syntax + tier-branch OK). The one ⚠️ is a pre-existing multi-token-vocab parity gap, LOW severity and out of scope for #1257 — recommend a follow-up issue to split CSV epic vocab for code_writer/task_decomposer parity.

ylohnitram:

The MR adds branching logic (epic detection, vocabulary fallback, and conditional '--decompose' flag) to tick_for_repo but ships no tests. Coverage gaps are concentrated in the happy path (epic vs non-epic command construction), the false-match prevention the comment explicitly promises, the index_of '> 0' boundary, the empty/custom vocabulary fallback, and the exec-sh error path for decomposed jobs. Recommend table-driven tests asserting on the exact job_cmd string plus an error-path test before merge.

ylohnitram:

The epic auto-trigger logic is largely sound: shell args remain escaped, decompose_flag only appends a static flag, and the worst case of a misclassification is harmless (a non-epic decomposed into a single subtask). Two low-severity correctness notes: the label match uses '> 0' where '>= 0'/'!= -1' is the correct not-found test for index_of (latent only because the JSON array string always starts with '['), and len(epic_vocab) assumes recall_latest returns "" rather than nil on a miss. No high-severity bugs, race conditions, or resource leaks introduced.

Issue #1267, opened 2026-06-22
tools/detect-doc-drift.sh: version-badge vs VERSION drift check (M1 of #1266, step 1)


ylohnitram:
Small, self-contained first slice of #1266 M1. No dependencies on other #1266 work — just detect + print.

**Add `tools/detect-doc-drift.sh`** (bash, ~40–70 lines). It compares ONE documented fact against reality and prints a TSV line per drift on stdout, then exits 0 (it is a detector, not a gate — print nothing when there is no drift):

```
DRIFT<TAB>version-badge<TAB><documented><TAB><actual>
```

Check: the `version-X.Y.Z` shields badge in `dev-apprenticeship/README.md` vs the version in `dev-apprenticeship/VERSION`. If they differ, print the DRIFT line with documented=badge value, actual=VERSION value.

Keep it dependency-free (bash + grep/sed only). Resolve the repo root relative to the script location so it runs from anywhere.

**Add `tools/test-detect-doc-drift.sh`** following the existing `tools/test-*.sh` style (plain bash, pass/fail counts, non-zero exit on failure): one fixture where badge and VERSION disagree (assert the DRIFT line is printed) and one where they match (assert empty output).

Scope is deliberately tiny — one check + its test. Further checks (federation count, dashboard tab count) are separate follow-ups under #1266.

Issue #1266, opened 2026-06-22
epic: self-improving issue-creation loop — federation detects its own work, files issues, learns from outcomes


ylohnitram:
## Goal

Close the **front-end** of the autonomous SDLC. Instead of consuming only a human-curated backlog, `dev-apprenticeship` detects work from its **own signals** (repo health + its own operation), files GitHub issues for it, runs them through the existing triage → plan → implement → review → release pipeline, and **learns which self-filed issues were valuable** so it improves at what is worth filing.

## What already exists

- `triage/agents/issue_creator.ag` drafts issues and, on the `autonomous` tier, files them via `forge-api.sh create-issue` (GitHub `POST /issues` works) and emits `triage:new_issue`.
- The rest of the pipeline (router/labeler/prioritizer → code_writer → review → release) is already wired to `triage:new_issue`.
- So the **draft → file → triage → plan → implement → review → release** path is already end-to-end. The missing pieces are the *source*, *dedup*, and *outcome feedback*.

## Gaps to close

1. **Source** — `issue_creator` reads `poll_inbox()` (external events only). No detector observes the repo, the codebase, or the federation's own logs.
2. **Dedup** — no fingerprint against open issues, so autonomous filing would refile the same signal every tick. Hard blocker for safe autonomous operation.
3. **Feedback** — it learns title/label conventions, but not whether a *filed* issue led to a merged fix (good) vs was closed wontfix/duplicate (noise).

## Milestones

- **M1 — Detectors (low-noise first).** A `tools/detect-*.sh` family emitting structured candidates `{title, body, labels, fingerprint, source}` into a memo queue `issue_creator` reads. Ship the most verifiable signals first:
  - doc/code drift (documented counts/versions vs reality — e.g. the federation count in `CLAUDE.md`, README version badges, dashboard tab count);
  - `TODO`/`FIXME`/`XXX` markers with no tracking issue;
  - then CI flake/fail signals.
- **M2 — Dedup + rate-limit gate.** Fingerprint each candidate (stable hash of source + location + normalized title); skip if an open issue already carries that fingerprint (search via `github-api.sh`); cap new issues per run (`ISSUE_CREATOR_MAX_NEW`). This is the noise gate that makes autonomous filing safe.
[...]

PR #1268, opened 2026-06-22
Add a small, self-contained `tools/detect-doc-drift.sh` that compares the version shields badge in `dev-apprenticeship/README.md` against `dev-apprenticeship/VERSION` and prints a TSV `DRIFT` line on mismatch, exiting 0 either way. Ship it with a matching `tools/test-detect-doc-drift.sh` covering the drift and no-drift cases. Keep it bash + grep/sed only with the repo root resolved relative to the script, per the issue scope.

ylohnitram:
Implements #1267.

Issue:
{"iid": 1267, "title": "tools/detect-doc-drift.sh: version-badge vs VERSION drift check (M1 of #1266, step 1)", "description": "Small, self-contained first slice of #1266 M1. No dependencies on other #1266 work \u2014 just detect + print.\n\n**Add `tools/detect-doc-drift.sh`** (bash, ~40\u201370 lines). It compares ONE documented fact against reality and prints a TSV line per drift on stdout, then exits 0 (it is a detector, not a gate \u2014 print nothing when there is no drift):\n\n```\nDRIFT<TAB>version-badge<TAB><documented><TAB><actual>\n```\n\nCheck: the `version-X.Y.Z` shields badge in `dev-apprenticeship/README.md` vs the version in `dev-apprenticeship/VERSION`. If they differ, print the DRIFT line with documented=badge value, actual=VERSION value.\n\nKeep it dependency-free (bash + grep/sed only). Resolve the repo root relative to the script location so it runs from anywhere.\n\n**Add `tools/test-detect-doc-drift.sh`** following the existing `tools/test-*.sh` style (plain bash, pass/fail counts, non-zero exit on failure): one fixture where badge and VERSION disagree (assert the DRIFT line is printed) and one where they match (assert empty output).\n\nScope is deliberately tiny \u2014 one check + its test. Further checks (federation count, dashboard tab count) are separate follow-ups under #1266.", "state": "opened", "labels": ["dev-apprenticeship", "implementation"], "assignees": [{"username": "ylohnitram"}], "author": {"username": "ylohnitram"}, "created_at": "2026-06-22T17:31:57Z", "updated_at": "2026-06-22T17:31:57Z", "user_notes_count": 0, "priority": null}

Plan:
- `tools/detect-doc-drift.sh` — new dependency-free bash detector (~40-70 lines). Resolves repo root relative to its own script location so it runs from anywhere. Greps the `version-X.Y.Z` shields badge out of `dev-apprenticeship/README.md` (via grep/sed), reads `dev-apprenticeship/VERSION`, and when they differ prints `DRIFT\tversion-badge\t<badge>\t<version>` to stdout (nothing when they match). Always exits 0 — detector, not gate.
- `tools/test-detect-doc-drift.sh` — new test following the existing `tools/test-*.sh` style (plain bash, pass/fail counters, non-zero exit on failure). Uses temp fixtures with README.md + VERSION: case 1 badge and VERSION disagree (asserts the single DRIFT line is printed), case 2 they match (asserts empty output).

Add a small, self-contained `tools/detect-doc-drift.sh` that compares the version shields badge in `dev-apprenticeship/README.md` against `dev-apprenticeship/VERSION` and prints a TSV `DRIFT` line on mismatch, exiting 0 either way. Ship it with a matching `tools/test-detect-doc-drift.sh` covering the drift and no-drift cases. Keep it bash + grep/sed only with the repo root resolved relative to the script, per the issue scope.

### DIFF
--- a/tools/detect-doc-drift.sh
+++ b/tools/detect-doc-drift.sh
@@ -1,0 +1,53 @@
+#!/bin/bash
+# tools/detect-doc-drift.sh: documentation drift detector (M1 of #1266, step 1).
+#
+# Compares ONE documented fact against reality and prints a TSV line per drift
+# on stdout, then exits 0. It is a DETECTOR, not a gate: it prints nothing when
+# there is no drift and never fails the build.
+#
+# Check: the `version-X.Y.Z` shields badge in dev-apprenticeship/README.md vs
+# the version in dev-apprenticeship/VERSION. On mismatch it prints:
+#
+#   DRIFT<TAB>version-badge<TAB><documented><TAB><actual>
+#
+# where documented = the badge value and actual = the VERSION value.
+#
+# Dependency-free (bash + grep/sed only). The repo root is resolved relative to
+# this script's location so it runs from anywhere.
+#
+# Usage: ./tools/detect-doc-drift.sh
+# Exit code: always 0.
+
+set -eu
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+REPO_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"
+
+README="$REPO_ROOT/dev-apprenticeship/README.md"
+VERSION_FILE="$REPO_ROOT/dev-apprenticeship/VERSION"
+
+# Allow overriding the inputs (used by the test harness with fixtures).
+README="${DETECT_DOC_DRIFT_README:-$README}"
+VERSION_FILE="${DETECT_DOC_DRIFT_VERSION:-$VERSION_FILE}"
+
+# Extract the first `version-X.Y.Z` shields badge value from the README.
+badge_version=""
+if [ -f "$README" ]; then
+    badge_version="$(grep -oE 'badge/version-[0-9]+\.[0-9]+\.[0-9]+' "$README" \
+        | head -n1 \
+        | sed -E 's#badge/version-##')"
+fi
+
+# Read the declared version (first non-empty line, trimmed).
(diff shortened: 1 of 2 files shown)


ylohnitram:

Both new bash scripts are well-formatted and stylistically consistent: 4-space indentation throughout, UPPER_CASE for script-level/global variables and lowercase for locals, consistent test-bracket style, and thorough header comments. The only style nit is a code-organization mismatch in the test file, whose header comment lists three tests while the body labels only two. No naming, whitespace, or import/ordering issues otherwise.

ylohnitram:

The two happy-path assertions (DRIFT TSV line on mismatch, empty output on match) are correct and use isolated env-overridden fixtures, which is good practice. Coverage is incomplete, though: the both-facts-present guard, the missing-file branches, the whitespace/first-non-empty VERSION parsing, and the multi-badge head -n1 selection are all untested. The exit-code assertions are effectively tautological under `set -e` and cannot fail-report. Recommend adding one-sided, missing-file, whitespace, and multi-badge cases and fixing the exit-code capture before merge.

ylohnitram:

Both scripts are largely correct for the happy path. The most material issue is in the test harness: under `set -eu`, the `RC=$?` exit-code assertions are unreachable for any non-zero exit, so the 'detector exits 0' checks can never fail and provide false confidence; the header-documented Test 3 is also not actually implemented. The detector itself is sound but silently treats missing README/VERSION/badge as 'no drift', which hides real failure modes. No off-by-one, null-deref, or race-condition bugs found in the production logic.

Issue #1269, opened 2026-06-22
code-edit-in-checkout: make --decompose reliable so epics split into small (proven-zone) sub-edits


ylohnitram:
## Problem

Live on #1266, the federation auto-`--decompose`d an epic but it ran **monolithically** and the editing agent timed out (exit 124) with zero staged edits — three identical failures. Small single-purpose tasks (e.g. #1267) succeed reliably in one attempt; the gap is purely task size.

## Root cause (from `tools/code-edit-in-checkout.sh`)

`--decompose` drives claude to write an ordered sub-edit list to a file (the decompose-gen step), then runs the edit loop once per subtask. But the decompose-gen step lets claude **explore the repository** before listing, so on a large task it burns the whole budget exploring and writes **nothing** to the list file → `awk` yields empty → the code falls back to "whole task = one subtask" (monolithic) **silently**. The monolithic editing attempt then over-explores the same way and times out with no edits.

## Fix (prompt + visibility only; no control-flow change)

1. **Decompose-gen prompt** — instruct claude to base the sub-edit list **solely on the task description**, do **not** read/explore repository files, and write the list immediately. Decomposing a task description into ordered steps does not require repo exploration.
2. **Initial editing prompt** — add a directive to begin editing immediately and keep exploration minimal (the task already specifies the change).
3. **Visibility** — when `--decompose` is set but decomposition yields no subtasks, log the monolithic fallback explicitly (today it is silent, which hid this root cause).

## Validation

The stub-based `tools/test-code-edit-in-checkout.sh` covers orchestration logic (must stay green — prompt-text changes are behavioural, not structural). Behavioural validation = a live re-run of #1266 after deploy: decomposition should produce subtasks (split markers in the job log) and each small subtask should land like #1267.

Part of making the federation handle epic-sized work (front-end of #1266).

PR #1275, opened 2026-06-22
fix(code_writer): epic auto-decompose reads the DRAFTED issue's labels, not issues_raw[0]

### LINKED ISSUE #1271: code_writer epic auto-decompose (#1257) does not fire on real epic-labelled GitHub issues
## Symptom

Live on #1266 (which carries the `epic` label), `code_writer` launched the checkout-edit job **without** `--decompose`: the daemon log shows `Drafted code for issue 1266` + `Launched detached checkout-edit for issue 1266` but **never** the `[code_writer] issue 1266 is epic-class — auto-decomposing (#1257)` line, and the job log shows no decompose step (straight to `attempt 1`, monolithic). So `is_epic` evaluated false despite the `epic` label.

## Where

`implementation/agents/code_writer.ag`, the #1257 block:

```
let epic_vocab = recall_latest("planning:labels:epic");      // resolves to "epic" ✓
let epic_label = if len(epic_vocab) > 0 { epic_vocab; } else { "epic"; };
let labels_str = to_string(json_get(issues_raw, "[0].labels"));
let is_epic = index_of(labels_str, "\"" + epic_label + "\"") > 0;
```

`planning:labels:epic` correctly resolves to `epic`. So the fault is in `labels_str` — most likely the `forge-api.sh --view code_writer` projection does not surface label **names** in the shape `index_of` expects (e.g. labels projected as an empty/である different field, or objects without the quoted name where this match looks). Needs: dump the actual `issues_raw[0].labels` string the GitHub backend returns for an epic-labelled issue and align the projection or the match.

## Impact

The epic auto-decompose path (#1257) is effectively dead on GitHub — epics run monolithically, which (combined with the editing agent over-exploring large tasks) makes them fail with zero edits. Small single-purpose issues are unaffected and continue to ship reliably (e.g. #1267).

Related: #1269 (decompose-gen prompt reliability) sits downstream of this and is already merged; it cannot help until `--decompose` is actually passed.

ylohnitram:
Closes #1271.

## Problem (observed live on #1266)

`code_writer`'s #1257 epic auto-trigger never fired for #1266 (which carries the `epic` label): the daemon log showed `Launched detached checkout-edit for issue 1266` but never `issue 1266 is epic-class — auto-decomposing`, and the job ran without `--decompose` (monolithic).

## Root cause

```
let labels_str = to_string(json_get(issues_raw, "[0].labels"));   // <- [0]
let is_epic = index_of(labels_str, "\"" + epic_label + "\"") > 0;
```

The check read `issues_raw[0].labels`, but `code_writer` prompts the LLM with the **whole** assigned snapshot and the LLM drafts **any** of them (`draft.issue_id`) — not necessarily the first. Whenever more than one issue is assigned, the epic flag is read off the wrong issue, so `--decompose` is silently dropped. (Labels themselves project correctly as name-strings; the bug is purely the `[0]` index.)

## Fix

Read the labels of the **drafted** issue from `issue_detail` (the `forge-api.sh issue <draft.issue_id>` result already fetched a few lines above for the task text), which is a single object with a name-string `labels` array. One-line change + an explanatory comment; the exact-quoted match is unchanged.

Together with #1269 (decompose-gen prompt reliability) this makes the epic auto-decompose path actually fire **and** split.

## Validation

- colony-lint (incl. per-agent `.ag` syntax + tier-branch checks): **432 passed, 0 failed, 5 skipped**.
- `.ag` has no unit-test harness; behavioural validation = a live re-run with #1266 as a sole assigned epic (should now log `epic-class — auto-decomposing` and pass `--decompose`).

### DIFF
--- a/dev-apprenticeship/CHANGELOG.md
+++ b/dev-apprenticeship/CHANGELOG.md
@@ -17,6 +17,15 @@ is asserted until multi-version CI is in place.
 
 ### Fixed
 
+- **`code_writer` epic auto-decompose (#1257) now fires for the issue it is
+  actually building, not the first assigned issue**
+  ([#1271](https://github.com/Replikanti/agentis-colonies/issues/1271)). The
+  `is_epic` check read `issues_raw[0].labels`, but the LLM may draft any issue in
+  the assigned snapshot, so whenever more than one issue was assigned the epic
+  flag was read off the wrong issue — `--decompose` was silently dropped and the
+  epic ran monolithically (observed live on #1266). It now reads the labels of
+  the **drafted** issue (`issue_detail`, already fetched for the task text).
+  Together with #1269 this makes the epic path actually decompose.
 - **`code-edit-in-checkout.sh --decompose` now reliably splits epic-sized tasks
   instead of silently falling back to a monolithic run**
   ([#1269](https://github.com/Replikanti/agentis-colonies/issues/1269)). The
--- a/dev-apprenticeship/CHANGELOG.md
+++ b/dev-apprenticeship/CHANGELOG.md
@@ -17,15 +17,22 @@ is asserted until multi-version CI is in place.
 
 ### Fixed
 
-- **`code_writer` epic auto-decompose (#1257) now fires for the issue it is
-  actually building, not the first assigned issue**
-  ([#1271](https://github.com/Replikanti/agentis-colonies/issues/1271)). The
+- **`code_writer` epic auto-decompose (#1257) now actually fires**
+  ([#1271](https://github.com/Replikanti/agentis-colonies/issues/1271)). Two
+  stacked defects kept `--decompose` from ever being passed for an epic: (1) the
   `is_epic` check read `issues_raw[0].labels`, but the LLM may draft any issue in
-  the assigned snapshot, so whenever more than one issue was assigned the epic
-  flag was read off the wrong issue — `--decompose` was silently dropped and the
-  epic ran monolithically (observed live on #1266). It now reads the labels of
-  the **drafted** issue (`issue_detail`, already fetched for the task text).
-  Together with #1269 this makes the epic path actually decompose.
+  the assigned snapshot, so with more than one assigned issue the flag was read
+  off the wrong issue; (2) more fundamentally, `to_string(json_get(obj,
+  "labels"))` of an **array-valued** field yields `"void"` on the runtime, so the
+  quoted-label `index_of` never matched at all — epic detection had been a no-op
+  since #1257 shipped. The check now matches the quoted label in the **raw JSON
+  of the drafted issue** (`index_of(issue_detail, "\"<epic>\"")`), which is
+  reliable on the runtime and preserves the anti-false-match (an `epic`-prefixed
+  label lacks the closing quote; title/body quotes are JSON-escaped). Together
+  with #1269 this makes the epic path actually decompose. Verified by live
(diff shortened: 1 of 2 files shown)


ylohnitram:

The core fix is sound: reading labels from the drafted issue (issue_detail) instead of issues_raw[0] correctly addresses #1271, and the label-object JSON shape is consistent between the list and detail endpoints so the substring match still holds. However, the surrounding logic carries a latent off-by-one: the epic check uses index_of(...) > 0 instead of >= 0, which would miss a match at position 0, and json_get(issue_detail, "labels") is not guarded against a null/missing issue_detail, so a failed fetch silently disables decomposition.

ylohnitram:
## QA Report — PR #1275 (`fix/epic-detection`)

Independent QA (incl. live `agentis` v1.19.1 runtime probes) of the epic auto-decompose detection fix.

| Area | Result | Notes |
|------|--------|-------|
| Diff scope (only the epic-detection check + comment + CHANGELOG) | ✅ | No stray edits. |
| `issue_detail` is the DRAFTED issue, in scope, fetched above | ✅ | Assigned from `forge-api.sh issue <draft.issue_id>` before the check. |
| Reads the drafted issue, not `issues_raw[0]` | ✅ | Fixes the wrong-issue selection when >1 issue is assigned. |
| `colony-lint.sh` (incl. per-agent `.ag` syntax + tier checks) | ✅ | 432 passed, 0 failed, 5 skipped. |
| Crash-safety on empty `issue_detail` | ✅ | Fetch-fail → no match → `is_epic` false → no `--decompose`. Harmless. |
| **`is_epic` actually fires** | ✅ (after fix) | **Critical finding (initially ❌):** on agentis v1.19.1, `to_string(json_get(obj, "labels"))` of an **array** field returns the literal `"void"`, so the quoted-label `index_of` never matched — epic detection had been a no-op since #1257. **Resolved** by matching the quoted label in the **raw drafted-issue JSON** (`index_of(issue_detail, "\"<epic>\"")`), which the runtime probe confirmed returns `true` while preserving the anti-false-match. |

### Resolution
The critical ❌ (epic detection never fired due to the array-stringify `"void"` quirk) was fixed in this PR by adopting the raw-text match the QA recommended. Net verdict: **✅ ship-it.** End-to-end behaviour (epic actually passes `--decompose` and decomposes) is confirmed by a live `#1266` re-run after deploy — the authoritative behavioural check beyond static/lint QA.

<sub>🤖 QA by agentis-qa-agent + operator follow-up</sub>

ylohnitram:

Mostly clean. Naming is consistent snake_case, indentation and whitespace are correct, and there are no import-ordering issues; the CHANGELOG.md additions are well-formed markdown. The only style nits are in code_writer.ag: the added inline comment block uses ALL-CAPS emphasis and is overly long (9 comment lines for one line of code).

ylohnitram:

The fix is plausible but ships with no automated test coverage at all, which is the central problem: the bug it repairs was a silent no-op that survived multiple releases specifically because the epic-detection path was never tested. Highest priority is a regression test asserting is_epic + '--decompose' on a drafted epic issue. Also missing: negative tests for the comment's anti-false-match guarantees (epic-foo prefix labels, JSON-escaped 'epic' in title/body), a test for the multi-assigned-issue case that is the actual root defect, an off-by-one boundary check on index_of(...) > 0, and replacement of live post-deploy probes with a deterministic CI fixture for the array-stringify-yields-void runtime quirk.

Issue #1283, opened 2026-06-22
tools/detect-todo-markers.sh: exclude vendored/generated dirs (node_modules, .solc-cache, dist, .git)


ylohnitram:
Surfaced by a live `tools/self-observe.sh` dry-run: `detect-todo-markers.sh` reports TODO/FIXME/XXX markers inside `node_modules/` and `.solc-cache/` (third-party / generated code), which are noise and must not become self-improvement issues.

**Fix:** in `tools/detect-todo-markers.sh`, exclude vendored/generated directories from the scan. Extend the `grep -r` exclusions beyond `.git` to also skip: `node_modules`, `.solc-cache`, `dist`, `build`, `target`, `__pycache__`, `vendor`. Use repeated `--exclude-dir=<name>` flags (already uses `--exclude-dir=.git`).

**Test:** extend `tools/test-detect-todo-markers.sh` — add a fixture with a `TODO` inside a `node_modules/` subdir and assert it is NOT reported, while a `TODO` in a normal file still is.

Tiny scope: one detector tweak + a test case. Found by the federation observing itself (#1266 M3).

PR #1284, opened 2026-06-22
Harden the in-tree TODO detector so it stops surfacing markers from third-party and generated code: in tools/detect-todo-markers.sh, broaden the grep scan's --exclude-dir set beyond .git to also skip node_modules, .solc-cache, dist, build, target, __pycache__, and vendor. Add a unit test proving a TODO inside node_modules/ is not reported while a TODO in a normal file still is, keeping scope to one detector tweak plus one test as the issue specifies.

ylohnitram:
Implements #1283.

Issue:
{"iid": 1283, "title": "tools/detect-todo-markers.sh: exclude vendored/generated dirs (node_modules, .solc-cache, dist, .git)", "description": "Surfaced by a live `tools/self-observe.sh` dry-run: `detect-todo-markers.sh` reports TODO/FIXME/XXX markers inside `node_modules/` and `.solc-cache/` (third-party / generated code), which are noise and must not become self-improvement issues.\n\n**Fix:** in `tools/detect-todo-markers.sh`, exclude vendored/generated directories from the scan. Extend the `grep -r` exclusions beyond `.git` to also skip: `node_modules`, `.solc-cache`, `dist`, `build`, `target`, `__pycache__`, `vendor`. Use repeated `--exclude-dir=<name>` flags (already uses `--exclude-dir=.git`).\n\n**Test:** extend `tools/test-detect-todo-markers.sh` \u2014 add a fixture with a `TODO` inside a `node_modules/` subdir and assert it is NOT reported, while a `TODO` in a normal file still is.\n\nTiny scope: one detector tweak + a test case. Found by the federation observing itself (#1266 M3).", "state": "opened", "labels": ["dev-apprenticeship", "implementation"], "assignees": [{"username": "ylohnitram"}], "author": {"username": "ylohnitram"}, "created_at": "2026-06-22T20:50:32Z", "updated_at": "2026-06-22T20:50:32Z", "user_notes_count": 0, "priority": null}

Plan:
- `tools/detect-todo-markers.sh` — extend the `grep -rInE` scan to skip vendored/generated trees by adding repeated `--exclude-dir=` flags for `node_modules`, `.solc-cache`, `dist`, `build`, `target`, `__pycache__`, and `vendor` alongside the existing `--exclude-dir=.git`; update the header comment to document the excluded directories.
- `tools/test-detect-todo-markers.sh` — new self-contained unit test (the file does not yet exist despite the issue saying "extend"); builds a fixture tree under a temp `DETECT_TODO_ROOT`, asserts a `TODO` planted inside a `node_modules/` subdir is NOT emitted, asserts a `TODO` in a normal source file IS still emitted, and follows the existing `test-check-exec-sh.sh` PASS/FAIL + mktemp/trap-cleanup convention.

Harden the in-tree TODO detector so it stops surfacing markers from third-party and generated code: in tools/detect-todo-markers.sh, broaden the grep scan's --exclude-dir set beyond .git to also skip node_modules, .solc-cache, dist, build, target, __pycache__, and vendor. Add a unit test proving a TODO inside node_modules/ is not reported while a TODO in a normal file still is, keeping scope to one detector tweak plus one test as the issue specifies.

### DIFF
--- a/tools/detect-todo-markers.sh
+++ b/tools/detect-todo-markers.sh
@@ -1,9 +1,13 @@
 #!/bin/bash
 # tools/detect-todo-markers.sh: in-tree TODO/FIXME/XXX marker detector (M1 of #1266).
 #
-# Walks the repo tree (skipping .git/) and prints one TSV line per marker on
-# stdout, then exits 0. It is a DETECTOR, not a gate: it prints nothing when
-# there are no markers and never fails the build.
+# Walks the repo tree and prints one TSV line per marker on stdout, then exits
+# 0. It is a DETECTOR, not a gate: it prints nothing when there are no markers
+# and never fails the build.
+#
+# Vendored and generated directories are skipped so third-party / build output
+# does not surface as noise: .git, node_modules, .solc-cache, dist, build,
+# target, __pycache__, vendor (#1283).
 #
 # For every TODO, FIXME, or XXX marker it prints:
 #
@@ -30,9 +34,19 @@ ROOT="${DETECT_TODO_ROOT:-$REPO_ROOT}"
 [ -d "$ROOT" ] || exit 0
 
 # Scan from inside the root so grep emits paths relative to it. -I skips binary
-# files, -n adds line numbers, -E enables the alternation. .git/ is excluded.
+# files, -n adds line numbers, -E enables the alternation. Vendored/generated
+# directories are excluded so third-party / build output is not reported.
 cd "$ROOT" || exit 0
-grep -rInE 'TODO|FIXME|XXX' --exclude-dir=.git . 2>/dev/null | while IFS= read -r hit; do
+grep -rInE 'TODO|FIXME|XXX' \
+    --exclude-dir=.git \
+    --exclude-dir=node_modules \
+    --exclude-dir=.solc-cache \
+    --exclude-dir=dist \
+    --exclude-dir=build \
+    --exclude-dir=target \
+    --exclude-dir=__pycache__ \
+    --exclude-dir=vendor \
+    . 2>/dev/null | while IFS= read -r hit; do
     # grep output is "<file>:<line>:<content>"; peel the first two colon fields.
     file="${hit%%:*}"
     rest="${hit#*:}"
--- a/tools/test-detect-todo-markers.sh
+++ b/tools/test-detect-todo-markers.sh
@@ -5,6 +5,7 @@
 #   Test 1: a file carrying a TODO marker -> prints exactly the DRIFT TSV line
 #   Test 2: a clean file                  -> no line is emitted for it
 #   Test 3: detector always exits 0 (it is a detector, not a gate)
+#   Test 4: a TODO inside node_modules/   -> excluded, not reported (#1283)
 #
 # The scanned tree is driven through DETECT_TODO_ROOT so the assertions use a
 # throwaway fixture dir and never depend on the live repo contents.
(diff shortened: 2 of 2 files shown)


ylohnitram:

The detector change broadens exclusions from one directory (.git) to eight, but the test suite only adds coverage for node_modules. Six of the seven newly excluded directories are untested, leaving the core behavior of this MR largely unverified; a flag typo would not be caught. Tests are otherwise sound (isolated tmp fixture, trap cleanup, exit-code check, no flaky timing/ordering), but coverage of the new exclusions should be completed and a positive over-exclusion control added before merge.

ylohnitram:

The diff is well aligned with the script's existing conventions: comment banners, the 4-space continuation/block indentation, and the pass/fail test helpers are all followed consistently, and the new Test 4 mirrors the existing test structure. Only one minor organization nit: the --exclude-dir list and its mirror comment use an ad-hoc ordering. No other style issues.

ylohnitram:

The new exclusion logic is only partially covered: of the 8 newly excluded directories, only node_modules has a test. The added Test 4 also uses a weak negative-only assertion with no positive control, and edge cases for nested/similarly-named directories are untested. Add per-directory exclusion fixtures, a positive control confirming valid markers are still reported, and nested/look-alike directory cases.

ylohnitram:

The diff only adds vendored/generated directory exclusions to the grep scan and a corresponding node_modules test; the parsing logic, exit-0 detector contract, and grep syntax are unchanged and correct. No off-by-one, null-deref, race, or resource-leak issues were introduced. The single correctness caveat is that --exclude-dir matches directory basenames at any depth, so legitimately-named source directories (build/dist/target/vendor) anywhere in the tree are skipped, which can mask real markers (low severity, false-negative risk).

ylohnitram:

The exclusion behavior change is only partially tested. Six of the seven newly excluded directories have no coverage, the single node_modules test uses a coarse substring assertion, and it lacks a paired positive assertion so an over-broad regression that suppresses all output would pass silently. Recommend a table-driven test over the full exclusion list, an exact-output assertion, and edge cases for nesting and similarly-named non-excluded directories.

ylohnitram:

No logic bugs found. The grep --exclude-dir additions use valid syntax and the inline comment's directory list stays in sync with the actual exclusion flags. Test 4 creates the node_modules fixture before the detector runs and asserts the path is absent from the captured output, so it is correct. The only correctness concern is that broad basename-based exclusions (build/target/dist/vendor) can mask markers in legitimately-named directories - a low-severity false-negative risk that is likely intended.

ylohnitram:

Tests are partially adequate but under-cover the change. The script adds seven new vendored/generated directory exclusions, yet only node_modules is tested; the other six (.solc-cache, dist, build, target, __pycache__, vendor) have no assertion, so a dropped or mistyped --exclude-dir flag would pass silently. Test 4's assertion is also weak (substring-only, with no positive co-check that real markers still surface in the same run) and over-exclusion boundary cases are unverified. Recommend table-driving the exclusion test across all excluded names and tightening it to an exact expected-line set.

ylohnitram:

The diff correctly adds a regression test for the node_modules exclusion, but test coverage lags the code: 6 of the 7 newly excluded directories have no test, so most of the new behavior is unverified. Test 4 also relies on a single weak negative substring check with no paired positive assertion, meaning a detector that silently produces no output would pass. Recommend a table-driven test over the full exclusion list plus a positive assertion that non-vendored markers are still reported, and optionally cases for nested and near-name directories.

ylohnitram:

No hard logic bugs found (no off-by-one, null/nil dereference, race condition, or resource leak; the mktemp+trap cleanup is unchanged and correct). The detector change correctly excludes vendored/generated trees, but the bare-basename --exclude-dir entries (build, dist, target, vendor) match at any tree depth and can silently drop markers in legitimately-named source directories -- a false-negative edge case for a tool whose purpose is completeness. One minor robustness concern in the new Test 4 assertion. Both findings are low severity.

ylohnitram:

The diff is stylistically clean and consistent. Comments are clear, the refactored multi-line grep uses uniform 4-space continuation indentation, and the new Test 4 follows the existing fixture and pass/fail helper conventions. Whitespace, naming, and organization are all in line with the surrounding code. The only minor style nit is the unsorted --exclude-dir list, which is duplicated in the header comment and could drift over time.

PR #1289, opened 2026-06-22
fix(detect-todo-markers): exclude .agentis run-dir + targets vendored crates (#1287)

### LINKED ISSUE #1287: tools/detect-todo-markers.sh: also exclude vendored crate targets + run-dir workspaces (.agentis/workspaces, tribes-bench/targets)
Surfaced by the live `self-observe.sh` sidecar running against the federation run dir: `detect-todo-markers.sh` still reports TODO/FIXME in vendored Rust crate targets (`tribes-bench/targets/stage*/.../*.rs`) and in per-issue workspace clones under `.agentis/workspaces/`. Same noise class as #1283 (node_modules).

**Fix:** extend the `--exclude-dir` set in `tools/detect-todo-markers.sh` to also skip `.agentis` (the run-dir state + workspace clones) and `targets` (tribes-bench vendored crates), alongside the node_modules/.solc-cache/dist/build/target/__pycache__/vendor already excluded in #1283. Add a fixture-backed test case for a TODO under `.agentis/` and under `targets/` asserting both are skipped.

Tiny scope. Found by the federation self-observing in production (#1266 M3 loop).

ylohnitram:
Closes #1287. Surfaced live by the `self-observe.sh` sidecar.

`detect-todo-markers.sh` still reported TODO/FIXME inside `.agentis/` (run-dir state + per-issue workspace clones) and `targets/` (tribes-bench vendored crates). Added both to the `--exclude-dir` set alongside the node_modules/.solc-cache/dist/build/target/__pycache__/vendor excludes from #1283. This is the noise gate that makes autonomous `--file` self-observation clean.

Test extended with `.agentis/` and `targets/` fixtures asserting both are skipped (6/6). `bash -n` + `shellcheck` clean; colony-lint 437/0.

### DIFF
--- a/dev-apprenticeship/CHANGELOG.md
+++ b/dev-apprenticeship/CHANGELOG.md
@@ -40,6 +40,15 @@ is asserted until multi-version CI is in place.
 
 ### Fixed
 
+- **`detect-todo-markers.sh` also excludes run-dir state and vendored crate
+  targets** ([#1287](https://github.com/Replikanti/agentis-colonies/issues/1287)).
+  Surfaced by the live `self-observe.sh` sidecar: the TODO scan still reported
+  markers inside `.agentis/` (run-dir state + per-issue workspace clones) and
+  `targets/` (tribes-bench vendored crates). Added both to the `--exclude-dir`
+  set (alongside the node_modules/.solc-cache/… excludes from #1283), so
+  self-observation no longer proposes vendored/run-dir noise — the gate that
+  makes autonomous `--file` filing clean. Test extended with `.agentis/` and
+  `targets/` fixtures (6/6).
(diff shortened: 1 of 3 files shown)


ylohnitram:

The MR adds two exclude-dir entries (.agentis, targets) to detect-todo-markers.sh and extends the test with matching fixtures plus two new cases (6/6 passing), so the new code is not untested. The weakness is test strength, not coverage: the added cases are negative-only and use unanchored substring matches against a single shared $OUT, so they would still pass if the detector silently emitted nothing, and no test confirms a non-excluded directory is still scanned. Recommend adding a positive control, anchoring assertions to the full fixture paths, and covering the over-exclusion edge case.

PR #1331, opened 2026-06-24
fix(self-observe): match TODO marker-form + stop filing raw todo-markers

ylohnitram:
The self-observe loop kept filing low-value `[self-observe] todo-marker` issues — including, absurdly, flagging the detector's **own header comment** (`TODO/FIXME/XXX marker detector`). Two root causes:

1. **detect-todo-markers matched bare words** `TODO|FIXME|XXX` anywhere → every *mention* (the detector's description, its grep pattern, doc prose) was a false positive. Now requires **marker form** `(TODO|FIXME|XXX)[:(]` → 187 bare hits → ~52 real markers; self-reference no longer matches.
2. **Even real TODO markers are notes, not work items.** self-observe now treats `todo-marker` as **log-only by default** (`SELF_OBSERVE_NOFILE_KINDS`, set to `""` to file all) — observed + logged, never filed. doc-drift + agent-failure still file (this honours 'keep filing the valuable signals').

Tests: detect-todo-markers **11/0**, self-observe **6/0**.

### DIFF
--- a/tools/detect-todo-markers.sh
+++ b/tools/detect-todo-markers.sh
@@ -37,7 +37,7 @@ ROOT="${DETECT_TODO_ROOT:-$REPO_ROOT}"
 # files, -n adds line numbers, -E enables the alternation. Vendored/generated
 # directories are excluded so third-party / build output is not reported.
 cd "$ROOT" || exit 0
-grep -rInE 'TODO|FIXME|XXX' \
+grep -rInE '(TODO|FIXME|XXX)[:(]' \
     --exclude-dir=.git \
     --exclude-dir=node_modules \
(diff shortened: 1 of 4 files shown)


ylohnitram:

The two new tests add useful negative coverage (bare-word mention not flagged; todo-marker logged-not-filed) but both stop at the happy/default path. Missing: a positive test for the new parenthesized marker form, regression coverage that FIXME/XXX still match and that space-form markers are intentionally dropped, and coverage of the documented NOFILE_KINDS overrides plus proof that non-log-only kinds still file. Tests are deterministic with no flaky timing/ordering patterns, but assertions are narrow and leave the feature's configurable branches untested.

